PR #22, opened 2025-04-22
Simplify pipelines by removing unnecessary `asyncio` usage

BoggerByte:
These pipelines are executed in a synchronous context, so creating a new event loop was unnecessary. This refactor simplifies the code and avoids the overhead of managing asynchronous execution when it's not needed.

#### **Description of changes:**
- Refactored the `BuildPipeline` to remove redundant use of `asyncio_run` and the async interface `arun_transformations_wrapper`. It now uses the synchronous `run_transformations` method from `llama-index`.
- Similarly, replaced `IngestionPipeline.arun` with its synchronous counterpart `IngestionPipeline.run`.
- To retain the parallelism previously handled by `asyncio.gather`, replaced async execution with `concurrent.futures.ProcessPoolExecutor` for concurrent node processing within a synchronous context.

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution under the terms of your choice.


iansrobinson:
Hi @BoggerByte 

Thank you very much for the PR. I'll take a detailed look tomorrow morning, and I'll pick up your questions around async then as well.

Kind regards

ian

BoggerByte:
Sounds clear, I'll try to fix these errors

BoggerByte:
Hey @iansrobinson! I just fixed the `AttributeError` and tested the progress bars and they're showing up nicely now. Also cleaned up the typing for `node_batches` in the `_run_pipeline` method (from `List[BaseNode]` to `List[List[BaseNode]]`). 
Let me know if you spot anything else!

BoggerByte:
I also noticed that the OpenSearch client was introducing some asyncio overhead, so I went ahead and fixed that as well. Let me know what you think!

iansrobinson:
Hi @BoggerByte 

Thank you very much for this.

I've pushed a couple of changes:

  - Removed the last async call from OpenSearch
  - The extraction pipeline still wasn't showing progress bars (or any log output). But the build pipeline was. I realised you'd solved the problem in the build pipeline, and there didn't really need to be any difference in how the two pipelines ran their components. So I refactored out the `run_pipeline` method from the build pipeline, and now I use that in the extraction pipeline as well.
  
Have a look at the changes, and let me know what you think. I'm really pleased with this PR - it removes all the unnecessary asyncio complexity with which I burdened the project.

Kind regards

ian

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/utils/pipeline_utils.py
```diff
@@ -2,11 +2,43 @@
...
     def _sink_from(generator):
         for item in generator:
             pass
     return Pipe(_sink_from)
 
-sink = _sink()
\ No newline at end of file
+sink = _sink()
+
+def run_pipeline(
```
BoggerByte, line 22: Rename to `run_ingestion_pipeline`

iansrobinson: Agreed on the name change.

I've also found an issue in OpenSearchIndex.paginated_search(), which had switched to using query instead of search. Can we replace that method with:

```
def paginated_search(self, query, page_size=10000, max_pages=None):
        
        client = self.client._os_client

        if not client:
            pass
        
        
        search_after = None
        page = 0
        
        while True:
            body = {
                "size": page_size,
                "query": query,
                "sort": [{"_id": "asc"}]
            }
            
            if search_after:
                body["search_after"] = search_after
                
            response = client.search(
                index=self.underlying_index_name(),
                body=body
            )

            hits = response['hits']['hits']
            if not hits:
                break

            yield hits

            search_after = hits[-1]['sort']
            page += 1

            if max_pages and page >= max_pages:
                break
```

BoggerByte: Oh, right, we definitely should replace it to `search` method! Must've made a mistake when I was refactoring the the code :)

PR #62, opened 2025-07-21
Adding CypherKGLinker in ByoKGQueryEngine

cmavro:
Adding CypherKGLinker and CypherByoKGQueryEngine to better handle openncypher linking and execution

*Issue #, if available:*

*Description of changes:*

Extending KGLinker with CypherKGLinker, which performs opencypher-based entity linking.
Also adding CypherByoKGQueryEngine that relies on opencypher-based entity linking and retrieval with feedback. 
Notebook in `examples/byokg-rag/byokg_rag_neptune_analytics_with_cypher_linking.ipynb` shows an end2end example of the new class CypherByoKGQueryEngine. 


byokg-rag/src/graphrag_toolkit/byokg_rag/graph_connectors/prompts/task_prompts.yaml
```diff
@@ -111,4 +111,33 @@ opencypher: '''
...
+    <opencypher-linking>
+    MATCH ...
+    WHERE ...
+    RETURN ...
+    LIMIT 5
+    </opencypher-linking>
+
+    One example includes:
+    MATCH (t:TYPE)
+    WHERE toLower(t.keyword) CONTAINS toLower(keyword) OR toLower(t.name) CONTAINS toLower(entity)
```
halcoope, line 140: Should your prompt for the LLM above specify to always use e.g. toLower() and CONTAINS for all WHERE clauses on keywords, given that we know this will make things more robust?

cmavro:
refactored to address offline comments: Added CypherKGLinker directly into ByoKGQueryEngine, where I added `cypher_query` functionality for better usage. (I can combine this functionality with the main `query` functionality later)

byokg-rag/src/graphrag_toolkit/byokg_rag/graph_connectors/prompts/task_prompts.yaml
cmavro: Updated the instructions ^

cmavro:
added `cypher_query` functionality within `query` for better user experience

cmavro:
> > added `cypher_query` functionality within `query` for better user experience
> 
> do we want to keep cypher_query as separate in that case?

imo it is fine for now

sojiadeshina:
> > > added `cypher_query` functionality within `query` for better user experience
> > 
> > 
> > do we want to keep cypher_query as separate in that case?
> 
> imo, it is fine for now, since we may combine `cypher_query` with `query` in alternative ways in the future

yeah but that also means we implicitly commit to people using cypher_query as a top level function in QueryEngine and us supporting it as such which i don't think we should encourage.

sojiadeshina (approved):
I'm okay to merge for now. But don't see why we shouldn't just have cypher linking happen implicitly for all queries where it's necessary and handle the logic within query.

We probably shouldn't even need standalone example for cypher linking. Just add example query where cypher linking is useful to the existing NA notebook

PR #77, opened 2025-10-16
added package versions

aghassel:
*Issue #, if available:*

*Description of changes:*
Added versions to packages

### DIFF
--- a/byokg-rag/src/graphrag_toolkit/byokg_rag/requirements.txt
+++ b/byokg-rag/src/graphrag_toolkit/byokg_rag/requirements.txt
@@ -1,18 +1,17 @@
 --only-binary :all:
 
-pydantic>=2.8.2
-boto3
-xmltodict
-colorama
-pyyaml
-pytest
+boto3>=1.39.11
+botocore>=1.39.11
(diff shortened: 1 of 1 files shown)


aghassel:
Validated versions were compatible with lexical graph

byokg-rag/src/graphrag_toolkit/byokg_rag/requirements.txt
```diff
@@ -1,18 +1,17 @@
...
-pytest
+boto3>=1.39.11
+botocore>=1.39.11
+colorama==0.4.6
 faiss-cpu==1.9.0
-thefuzz
-langchain_huggingface
-langchain_aws
-llama-index-embeddings-bedrock==0.6.1
-ipykernel
```
cmavro, line 14: Do we need `ipykernel` this for the notebooks? Do they run successfully?

sojiadeshina: we can probably assume that users running notebook examples have jupyter installed and if it's only for examples then it's not really a src or package requirement.

PR #108, opened 2026-02-19
fix(indexing): prevent chunk ID boundary collision with delimiter

### LINKED ISSUE #107: [BUG]`create_chunk_id` — concatenation boundary collision
## Location

`src/graphrag_toolkit/lexical_graph/indexing/id_generator.py:84`

## Description

`create_chunk_id` hashes `text` and `metadata_str` by concatenating them directly
before computing the hash:

```python
def create_chunk_id(self, source_id: str, text: str, metadata_str: str):
    """
    Generates a unique chunk identifier by combining a source ID, a hash of the given text,
    and associated metadata.

    Args:
        source_id (str): The identifier of the source content.
        text (str): The primary content or text that needs to be identified.
        metadata_str (str): The metadata string used for constructing the identifier.

    Returns:
        str: A uniquely generated chunk identifier based on the given inputs.
    """
    return f'{source_id}:{self._get_hash(text + metadata_str)[:8]}'
```

Because there is no delimiter between the two inputs, any pair of `(text, metadata_str)`
values whose concatenation produces the same string will hash to the same value and
therefore generate the same chunk ID — even if the individual inputs are different.

## Example

```python
id1 = gen.create_chunk_id(source_id, "hello", "world")   # hashes "helloworld"
id2 = gen.create_chunk_id(source_id, "hell", "oworld")   # hashes "helloworld"

id1 == id2  # True — collision
```

Both calls produce `"helloworld"` before hashing, so the resulting chunk IDs are identical
despite representing different `(text, metadata_str)` inputs.

## Impact

Two distinct chunks with boundary-shifted text and metadata values would be assigned
the same ID. In a deduplication system this means one chunk silently overwrites the
other in the graph, causing data loss without any error.

In practice the risk depends on how callers construct `metadata_str`. If metadata is
always a structured, fixed-length string (e.g. a serialized dict), accidental boundary
shifts are unlikely. But the code offers no protection against it.

### Option A

Introduce a delimiter that cannot appear in either input, or user provided as a function param (defaulting to ex: \x00)

```python
return f'{source_id}:{self._get_hash(text + delim + metadata_str)[:8]}'
```

This ensures `("hello", "world")` hashes `"hello\x00world"` and `("hell", "oworld")` hashes
`"hell\x00oworld"` — distinct strings, distinct IDs.

### Option B

Call the _get_hash function on both inputs then call it again on the result

```python
return f'{source_id}:{self._get_hash(self._get_hash(text) + self._get_hash(metadata_str)))[:8]}'
```

## Notes

- The collision is deterministic and reproducible, so it can be caught by a unit test
  (see `test_create_chunk_id_concatenation_boundary` in `test_id_generator.py`).
  We'll have to modify it to not pass in the case of collision after commiting the changes described in this document.

abhu85:
## Summary

Fix a boundary collision vulnerability in `create_chunk_id()` where different `(text, metadata)` pairs could produce identical hash inputs.

**Root Cause:**
The original implementation concatenated text and metadata directly before hashing:
```python
self._get_hash(text + metadata_str)
```

This allowed different input pairs to produce the same concatenation:
- `("hello", "world")` -> hashes `"helloworld"`
- `("hell", "oworld")` -> also hashes `"helloworld"` -> **COLLISION!**

**Fix:**
Add a null byte delimiter (`\x00`) between text and metadata before hashing:
```python
self._get_hash(text + self._CHUNK_ID_DELIMITER + metadata_str)
```

This ensures:
- `("hello", "world")` -> hashes `"hello\x00world"`
- `("hell", "oworld")` -> hashes `"hell\x00oworld"`

The null byte is chosen because it cannot appear in valid UTF-8 text strings.

## Breaking Change Notice

This change modifies the hash output for all chunk IDs. Existing indexed data will have different chunk IDs than newly indexed data. Users upgrading should be aware of this one-time migration concern.

## Test Plan

- [x] Added comprehensive unit tests for boundary collision prevention
- [x] Tests verify multiple collision scenarios are now prevented
- [x] Tests verify deterministic behavior is preserved
- [x] Tests verify empty string handling works correctly
- [x] All 9 new tests pass

```
$ PYTHONPATH=src pytest tests/unit/indexing/test_id_generator.py -v
tests/unit/indexing/test_id_generator.py::TestCreateChunkId::test_create_chunk_id_basic PASSED
tests/unit/indexing/test_id_generator.py::TestCreateChunkId::test_create_chunk_id_deterministic PASSED
tests/unit/indexing/test_id_generator.py::TestCreateChunkId::test_create_chunk_id_no_boundary_collision PASSED
tests/unit/indexing/test_id_generator.py::TestCreateChunkId::test_create_chunk_id_boundary_collision_more_cases PASSED
tests/unit/indexing/test_id_generator.py::TestCreateChunkId::test_create_chunk_id_empty_strings PASSED
tests/unit/indexing/test_id_generator.py::TestCreateChunkId::test_create_chunk_id_different_source_ids PASSED
tests/unit/indexing/test_id_generator.py::TestCreateSourceId::test_create_source_id_format PASSED
tests/unit/indexing/test_id_generator.py::TestCreateSourceId::test_create_source_id_deterministic PASSED
tests/unit/indexing/test_id_generator.py::TestTenantIsolation::test_chunk_id_tenant_isolation PASSED
========================= 9 passed, 1 warning in 0.03s =========================
```

Fixes #107

---
Generated with [Claude Code](https://claude.ai/code)

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/id_generator.py
+++ b/lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/id_generator.py
@@ -68,10 +68,18 @@ class IdGenerator(BaseModel):
         """
         return f"aws::{self._get_hash(text)[:8]}:{self._get_hash(metadata_str)[:4]}"
         
+    # Delimiter used to separate text and metadata in chunk ID hashing.
+    # Using null byte as it cannot appear in valid UTF-8 text strings.
+    _CHUNK_ID_DELIMITER = '\x00'
+
     def create_chunk_id(self, source_id:str, text:str, metadata_str:str):
         """
         Generates a unique chunk identifier by combining a source ID, a hash of the given text,
         and associated metadata.
+
+        The text and metadata are separated by a null byte delimiter before hashing to prevent
+        boundary collision issues where different (text, metadata_str) pairs could produce
+        identical concatenated strings.
 
         Args:
             source_id (str): The identifier of the source content.
@@ -81,7 +89,7 @@ class IdGenerator(BaseModel):
         Returns:
             str: A uniquely generated chunk identifier based on the given inputs.
         """
-        return f'{source_id}:{self._get_hash(text + metadata_str)[:8]}'
+        return f'{source_id}:{self._get_hash(text + self._CHUNK_ID_DELIMITER + metadata_str)[:8]}'
     
     def rewrite_id_for_tenant(self, id_value:str):
         """
--- a/lexical-graph/tests/unit/indexing/__init__.py
+++ b/lexical-graph/tests/unit/indexing/__init__.py
@@ -1,0 +1,2 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
--- a/lexical-graph/tests/unit/indexing/test_id_generator.py
+++ b/lexical-graph/tests/unit/indexing/test_id_generator.py
@@ -1,0 +1,148 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+import pytest
+
+
+class TestCreateChunkId:
+    """Tests for IdGenerator.create_chunk_id method."""
+
+    def test_create_chunk_id_basic(self, default_id_gen):
+        """Test basic chunk ID creation."""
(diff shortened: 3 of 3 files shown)


iansrobinson:
Hi @abhu85 

Thank you. Please would you create a new PR with these changes and I'll look to merge this week.

Kind regards

ian

PR #185, opened 2026-04-04
feat(lexical-graph): invoke graph_store init hook

zollie:
Invoke existing graph_store.init() lifecycle hook during LexicalGraphIndex initialization, so backend bootstrap logic (if implemented) runs automatically. No API change; no-op for stores that don’t override init().

### DIFF
--- a/lexical-graph/src/graphrag_toolkit/lexical_graph/lexical_graph_index.py
+++ b/lexical-graph/src/graphrag_toolkit/lexical_graph/lexical_graph_index.py
@@ -300,6 +300,9 @@ class LexicalGraphIndex():
         self.extraction_pre_processors = pre_processors
         self.extraction_components = components
 
+        # Backend bootstrap (for example index creation) is part of object lifecycle; fail fast if unavailable.
+        self.graph_store.init()
+
     def _configure_extraction_pipeline(self, config: IndexingConfig):
         """
         Configures and initializes the extraction pipeline based on the given configuration settings. This method
@@ -783,4 +786,3 @@ class LexicalGraphIndex():
         delete_sources = DeleteSources(graph_store=self.graph_store, vector_store=self.vector_store)
 
         return delete_sources.delete_source_documents(source_ids)
-
--- a/lexical-graph/tests/unit/test_lexical_graph_index.py
+++ b/lexical-graph/tests/unit/test_lexical_graph_index.py
@@ -2,13 +2,15 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import os
-from unittest.mock import patch
+from unittest.mock import Mock, patch
+import pytest
 
 from llama_index.core.llms import LLM
 from llama_index.core.llms.mock import MockLLM
 from llama_index.llms.bedrock_converse import BedrockConverse
 
 from graphrag_toolkit.lexical_graph import ExtractionConfig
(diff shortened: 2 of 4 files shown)


lexical-graph/src/graphrag_toolkit/lexical_graph/lexical_graph_index.py
```diff
@@ -290,6 +290,7 @@ def __init__(
         tenant_id = to_tenant_id(tenant_id)
 
         self.graph_store = MultiTenantGraphStore.wrap(GraphStoreFactory.for_graph_store(graph_store), tenant_id)
+        self.graph_store.init()
```
mykola-pereyma: self.graph_store.init() is called inside __init__() with no exception handling. Both Neo4jGraphStore.init() and the new FalkorDBDatabaseClient.init() (PR #186) perform network I/O. If the database is unreachable at construction time, the exception prevents self.tenant_id, self.extraction_dir, and the extraction pipeline from being set — leaving the object in a partially constructed state. This is a behavioral change for users who construct the index before the database is available.

lexical-graph/tests/unit/test_lexical_graph_index.py
```diff
@@ -42,4 +43,33 @@ def test_extraction_lmm_configured_with_llm_cache_returns_llm_cache(self):
             extraction_llm = llm_cache
         )
 
-        assert extraction_config.extraction_llm == llm_cache
\ No newline at end of file
+        assert extraction_config.extraction_llm == llm_cache
+
+class TestLexicalGraphIndex:
```
mykola-pereyma, line 49: Would be worth adding a test for the failure case (graph_store.init.side_effect = Exception(...)) to document the expected behavior.

mykola-pereyma:
LexicalGraphQueryEngine creates a MultiTenantGraphStore via the same factory pattern but doesn't call init(). The indexes created by init() are needed for queries too — should the query engine also invoke the hook?

lexical-graph/src/graphrag_toolkit/lexical_graph/lexical_graph_index.py
zollie: > self.graph_store.init() is called inside **init**() with no exception handling. Both Neo4jGraphStore.init() and the new FalkorDBDatabaseClient.init() (PR #186) perform network I/O. If the database is unreachable at construction time, the exception prevents self.tenant_id, self.extraction_dir, and the extraction pipeline from being set — leaving the object in a partially constructed state. This is a behavioral change for users who construct the index before the database is available.

This is intentional fail-fast behavior: if a backend’s init() cannot complete (for example, DB unreachable), construction fails immediately instead of deferring failure to first use. Backends that do not need setup are unaffected because base GraphStore.init() is a no-op. I also added a unit test documenting this expected behavior by asserting the constructor propagates graph_store.init() exceptions.

lexical-graph/tests/unit/test_lexical_graph_index.py
zollie: Good point. I added a failure-path unit test that sets graph_store.init.side_effect and asserts LexicalGraphIndex(...) propagates the exception. This documents the expected fail-fast constructor behavior.

acarbonetto (approved):

iansrobinson:
Thank you @zollie for the contribution. Sorry for the delay in merging

PR #211, opened 2026-04-21
perf(retrieval): merge statements + facts queries into one Cypher round-trip

### LINKED ISSUE #210: [PERF] get_statements_by_topic_and_source fires two serial Cypher queries where one OPTIONAL MATCH suffices
## Summary

`TraversalBasedBaseRetriever.get_statements_by_topic_and_source` runs two serial Cypher queries and reconciles them in Python. The second query (facts via `__SUPPORTS__`) is an outer-join that the Cypher engine can do natively via `OPTIONAL MATCH`. Under `CompositeTraversalBasedRetriever` this method runs on ~20–40 workers per `query_graph` invocation, so removing one round-trip per worker collapses a whole wave of graph-store traffic.

## Where it is

`lexical-graph/src/graphrag_toolkit/lexical_graph/retrieval/retrievers/traversal_based_base_retriever.py` — `get_statements_by_topic_and_source` at line 128.

## Today

```python
# Query 1: statements grouped by topic/source (facts omitted — filled to [] with score: 0)
statements_results = self.graph_store.execute_query(statements_cypher, statements_params)

# Query 2: facts for the same statementIds — strictly serial after Query 1
statement_facts_results = self.graph_store.execute_query(statement_facts_cypher, {'statementIds': statement_ids})

# Python-side outer-join: weld facts into each statement dict
statement_facts = {r['statementId']: r['facts'] for r in statement_facts_results}
for statements_result in statements_results:
    for topic in statements_result['result']['topics']:
        for statement in topic['statements']:
            facts = statement_facts.get(statement['statementId'], [])
            if facts:
                statement['facts'] = facts
                statement['score'] = len(facts)
```

## Proposed

Collapse to a single query with:

```cypher
OPTIONAL MATCH (f)-[:`__SUPPORTS__`]->(l)
WITH t, l, c, s, collect(DISTINCT f.value) AS facts
```

…then thread `facts` through the existing `WITH` stages and use `size(facts)` where the Python loop was writing `len(facts)`. No other changes. The Python merge loop disappears.

## Correctness

- Statements with N facts → `facts=[v1,...,vN]`, `score=size(facts)=N`. Today's `len(facts)` gives the same N.
- Statements with no facts → `OPTIONAL MATCH` miss → `collect(DISTINCT f.value)` returns `[]` → `size(facts)=0`. Today's default was `score: 0, facts: []`. Identical.
- Multi-chunk statements: the intermediate `WITH t, l, c, s, collect(DISTINCT f.value) AS facts` groups at the (topic, statement, chunk, source) level, preserving today's row shape before the topic-level aggregation.

Return shape is byte-for-byte identical. Public method signature and callers are unchanged.

## Expected impact

Each worker: 2 Cypher round-trips → 1. Today the facts query is small (single one-hop `__SUPPORTS__` pattern with a short `IN` list), but it's strictly serial after the main query — so even 30–60 ms per worker compounds across the ~20–40 concurrent workers the retriever launches.

In a Yahoo-internal benchmark (194 production queries, Neptune Serverless, `max_pool_connections=32`), the graph-traversal phase runs ~1,500 ms on average today; we estimate ~1,100–1,200 ms after the merge — net ~300–400 ms per query (~8% of total query latency).

## Out of scope

- Batching the method across workers — saves more but changes per-chunk scoring semantics and is a bigger behaviour-change surface.
- Reducing `intermediate_limit` — trims the IN-list but trades off recall.

PR: #TBD

voidwisp:
## Summary

- Collapses the two serial Cypher queries in `get_statements_by_topic_and_source` into one via `OPTIONAL MATCH`.
- Closes #210.
- Return shape is byte-for-byte identical. No caller changes.

## Motivation

Under the default `CompositeTraversalBasedRetriever`, this method runs on ~20–40 workers per `query_graph` invocation. Each worker currently pays for two serial Cypher round-trips (main traversal + facts), with the second strictly blocked on the first. Removing one round-trip per worker collapses a wave of graph-store traffic from the critical path.

In a Yahoo-internal benchmark (194 production queries, Neptune Serverless, `max_pool_connections=32`), the graph-traversal phase runs ~1,500 ms on average today; we estimate ~1,100–1,200 ms after the merge — net ~300–400 ms per query (~8% of total latency).

## What changed

One file. The method now fires a single Cypher:

```cypher
MATCH (t)<-[:`__BELONGS_TO__`]-(l:`__Statement__`)-[:`__MENTIONED_IN__`]->(c)-[:`__EXTRACTED_FROM__`]->(s)
WHERE l.statementId in $statementIds
OPTIONAL MATCH (f)-[:`__SUPPORTS__`]->(l)
WITH t, l, c, s, collect(DISTINCT f.value) AS facts
...
```

- `facts` is threaded through the existing `WITH` stages.
- `ll.facts` is populated from `facts` (was `[]` + Python weld).
- `ll.score` is `size(facts)` (was `0` + Python `len(facts)` if non-empty).
- The second Cypher query and the entire Python merge loop are gone.

## Correctness

- Multi-fact statement: `facts=[v1,...,vN]`, `score=N` — matches `len(facts)` today.
- Zero-fact statement: `OPTIONAL MATCH` miss → `collect(DISTINCT f.value) = []` → `size(facts)=0`. Today's Python path leaves `score: 0, facts: []` on the same statement. Identical.
- Multi-chunk statement: `WITH t, l, c, s, collect(DISTINCT f.value) AS facts` collapses the fact rows at the (topic, statement, chunk, source) grain before downstream aggregation, preserving today's row shape.

## Testing

The change is entirely Cypher-side. The `tests/unit/retrieval/` tree today covers the retrieval *model* and a few processors but has no fixtures for retriever methods, so a unit test here would either re-assert the query string (low value) or require standing up a new retriever test harness (out of scope for a perf PR).

Validated in Yahoo's production Neptune Serverless workload — 62 traces, consistent latency improvement as projected above. Happy to add tests in whatever pattern you prefer for this layer if you have one in mind.

## Out of scope

- Batching this method across workers (bigger behaviour-change surface).
- Reducing `intermediate_limit` (recall trade-off, app-level).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lexical-graph/src/graphrag_toolkit/lexical_graph/retrieval/retrievers/traversal_based_base_retriever.py
+++ b/lexical-graph/src/graphrag_toolkit/lexical_graph/retrieval/retrievers/traversal_based_base_retriever.py
@@ -135,71 +135,50 @@ class TraversalBasedBaseRetriever(BaseRetriever):
 
         chunk_metadata = 'properties(c)' if self.args.include_chunk_details else '{}'
 
+        # Single-query form: OPTIONAL MATCH attaches each statement's supporting
+        # facts inline. Before this, two serial Cypher round-trips were fired per
+        # worker and reconciled in Python — one per-worker round-trip is removed.
         statements_cypher = f'''
-        // get statements grouped by topic and source
-        MATCH (t)<-[:`__BELONGS_TO__`]-(l:`__Statement__`)   
+        // get statements grouped by topic and source, with facts inline
+        MATCH (t)<-[:`__BELONGS_TO__`]-(l:`__Statement__`)
               -[:`__MENTIONED_IN__`]->(c)
               -[:`__EXTRACTED_FROM__`]->(s)
         WHERE {self.graph_store.node_id("l.statementId")} in $statementIds
-        WITH {{ 
-                sourceId: {self.graph_store.node_id("s.sourceId")}, 
-                metadata: properties(s), 
+        OPTIONAL MATCH (f)-[:`__SUPPORTS__`]->(l)
+        WITH t, l, c, s, collect(DISTINCT f.value) AS facts
+        WITH {{
+                sourceId: {self.graph_store.node_id("s.sourceId")},
+                metadata: properties(s),
                 versioning: {{
-                    valid_from: coalesce(s.{VALID_FROM}, {TIMESTAMP_LOWER_BOUND}), 
+                    valid_from: coalesce(s.{VALID_FROM}, {TIMESTAMP_LOWER_BOUND}),
                     valid_to: coalesce(s.{VALID_TO}, {TIMESTAMP_UPPER_BOUND}),
                     extract_timestamp: coalesce(s.{EXTRACT_TIMESTAMP}, {TIMESTAMP_LOWER_BOUND}),
                     build_timestamp: coalesce(s.{BUILD_TIMESTAMP}, {TIMESTAMP_LOWER_BOUND}),
                     id_fields: split(coalesce(s.{VERSION_INDEPENDENT_ID_FIELDS}, ""), ";")
-                }}  
+                }}
             }} AS source,
-            t, l, c,
-            {{ chunkId: {self.graph_store.node_id("c.chunkId")}, value: NULL, metadata: {chunk_metadata} }} AS cc, 
-            {{ statementId: {self.graph_store.node_id("l.statementId")}, statement: l.value, facts: [], details: l.details, chunkId: {self.graph_store.node_id("c.chunkId")}, score: 0 }} as ll
-        WITH source, 
-            t, 
-            collect(distinct cc) as chunks, 
+            t, l, c, facts,
+            {{ chunkId: {self.graph_store.node_id("c.chunkId")}, value: NULL, metadata: {chunk_metadata} }} AS cc,
+            {{ statementId: {self.graph_store.node_id("l.statementId")}, statement: l.value, facts: facts, details: l.details, chunkId: {self.graph_store.node_id("c.chunkId")}, score: size(facts) }} as ll
+        WITH source,
+            t,
+            collect(distinct cc) as chunks,
             collect(ll) as statements
         WITH source,
-            {{ 
-                topic: t.value, 
+            {{
+                topic: t.value,
                 topicId: {self.graph_store.node_id("t.topicId")},
                 chunks: chunks,
                 statements: statements
             }} as topic
         WITH sum(size(topic.statements)/size(topic.chunks)) AS score, source, collect(topic) AS topics
         RETURN {{
-            score: score, 
+            score: score,
             source: source,
             topics: topics
         }} as result ORDER BY result.score DESC LIMIT $limit'''
-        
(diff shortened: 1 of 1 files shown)


voidwisp:
After testing this seems to be actually a performance regression

voidwisp:
Full postmortem — benchmarked against production Neptune Serverless before and after; sharing data so anyone reading this thread has context.

## What I measured

Ran OLD (installed v3.17.1) vs NEW (verbatim PR body) against prod Neptune, 120 statement IDs sampled (80 with supporting facts, 40 without — exercises both OPTIONAL MATCH hit and miss), `include_chunk_details=False` and `True`. Correctness: **both variants return byte-identical normalized results across 14 sources / 66 topics** — the semantic argument holds.

Timing sweep by IN-list size, median of 5 post-warmup runs:

| IN-list | OLD (ms) | NEW (ms) | Δ |
|---:|---:|---:|---:|
| 1 | 49 | 36 | **-13 (-26%)** |
| 3 | 51 | 41 | **-11 (-21%)** |
| 5 | 56 | 46 | **-11 (-19%)** |
| 10 | 59 | 58 | neutral |
| 25 | 69 | 91 | **+22 (+32%)** |
| 50 | 84 | 147 | **+63 (+75%)** |
| 120 | 133 | 320 | **+187 (+140%)** |

Crossover between n=10 and n=25. Default `intermediate_limit = 50` caps each caller's IN-list, and all 6 call sites (`entity_network_search`, `topic_based_search`, `entity_based_search` ×2, `chunk_based_search`, `chunk_based_semantic_search`) apply that cap via `LIMIT $statementLimit`. Production runs routinely hit sizes where NEW is significantly slower.

## Likely cause

`OPTIONAL MATCH (f)-[:__SUPPORTS__]->(l)` inside the existing `(t)<-[:BELONGS_TO]-(l)-[:MENTIONED_IN]->(c)-[:EXTRACTED_FROM]->(s)` match expands the intermediate row count to *chunks × facts* per statement, which the `WITH t, l, c, s, collect(DISTINCT f.value) AS facts` step then re-collapses. Neptune's planner appears to materialize the cross-product rather than streaming through it, so cost scales super-linearly with IN-list size.

The OLD path runs two small, cleanly-planned queries plus a cheap in-memory Python merge — it wins because Neptune plans each simple pattern efficiently, and the Python merge cost is negligible versus the second network round-trip saved only when the cross-product stays small.

## Where the original analysis went wrong

My PR claimed the second query cost ~30–60 ms per call and could be collapsed "for free". The measurement refutes "for free" — `OPTIONAL MATCH` in this topology changes the plan shape in a way that overwhelms the saved round-trip at production list sizes.

## Possible paths forward (not pursuing here)

Pattern comprehension might avoid the cross-product by evaluating facts per-row inline:

```cypher
WITH t, l, c, s, [(f)-[:__SUPPORTS__]->(l) | f.value] AS facts_raw
WITH t, l, c, s, apoc.coll.toSet(facts_raw) AS facts  // if APOC available
```

Needs its own benchmark and introduces a dedup dependency on the runtime. Not reopening without better data.

Closing #210 with the same analysis.

PR #219, opened 2026-04-23
Fix dependabot error from bumping astro and @astrojs/starlight in /docs-site

oussamahansal:
*Description of changes:*
- Fix dependabot error from bumping astro and @astrojs/starlight in /docs-site

### DIFF
--- a/docs-site/package.json
+++ b/docs-site/package.json
@@ -17,8 +17,5 @@
     "react": "^18.3.1",
     "react-dom": "^18.3.1",
     "sharp": "^0.33.5"
-  },
-  "overrides": {
-    "@astrojs/sitemap": "3.2.1"
   }
 }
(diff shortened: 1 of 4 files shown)


docs-site/package.json
```diff
@@ -17,8 +17,5 @@
     "react": "^18.3.1",
     "react-dom": "^18.3.1",
     "sharp": "^0.33.5"
-  },
-  "overrides": {
-    "@astrojs/sitemap": "3.2.1"
```
acarbonetto, line 22: consider adding `@astrojs/sitemap` as a dependency above ("9.0.1"?) if you think this is needed

acarbonetto (approved):

docs-site/package.json
oussamahansal: not needed anymore

PR #232, opened 2026-04-28
Validate that a Bedrock model or inference profile is available and enabled before attempting to use it

oussamahansal:
*Issue #176*

*Description of changes:*

- Pre-flight check to validate the model id or cross-region inference model is available
- fail in CFN templates on the condition that model is a geo-inferred model

*resolves #176*


examples/lexical-graph/cloudformation-templates/graphrag-toolkit-neptune-analytics-aurora-postgres.json
```diff
@@ -795,97 +795,19 @@
                   "Fn::Sub": "arn:${AWS::Partition}:bedrock:*::foundation-model/anthropic.claude-3-5-haiku-20241022-v1:0"
```
mykola-pereyma: is it intentional to keep haiku 3.5 and sonnet 4.6?

oussamahansal: let's discuss which model to keep across the project and open a separate issue for making the change to keep it consistent

oussamahansal: I think here it's only listed in IAM policies so the SageMaker notebook has permission to invoke it if a user chooses to

iansrobinson:
@oussamahansal The majority of integration tests are failing here with `AccessDeniedException` on the extraction side (the tests appear to be using`claude-sonnet-4-20250514-v1`, but without the requisite IAM permissions), and `Wait timed out after 120 seconds` on the query side.

oussamahansal:
@iansrobinson Thanks for checking, I was able to reproduce it. The root cause is that the CFN templates hardcoded specific models in the IAM policy ARNs, the IAM role only grants the exact models listed in the template (anthropic.claude-sonnet-4-6.). If the test code or defaults reference a model that doesn't exactly match those ARNs, the call gets denied.

I fixed it by replacing the hardcoded per-model IAM ARNs with wildcards.
Before we had 6 Anthropic entries, each hardcoded:
```
foundation-model/anthropic.claude-3-5-haiku-20241022-v1:0
inference-profile/${regionPrefix}.anthropic.claude-3-5-haiku-20241022-v1:0
foundation-model/anthropic.claude-3-7-sonnet-20250219-v1:0
inference-profile/${regionPrefix}.anthropic.claude-3-7-sonnet-20250219-v1:0
foundation-model/anthropic.claude-sonnet-4-20250514-v1:0
inference-profile/${regionPrefix}.anthropic.claude-sonnet-4-20250514-v1:0
```
Now we have 2 wildcard entries:
```
foundation-model/anthropic.*
inference-profile/*.anthropic.*
```
Basically now any Anthropic model works without needing a template update, any inference profile prefix is covered.
The regionPrefix derivation which produces invalid ap in APAC regions was also removed.
I also upgraded the default model to anthropic.claude-sonnet-4-6.

Note: `anthropic.claude-sonnet-4-20250514-v1:0 `is now a legacy model in Bedrock. Some models are moving through their lifecycle rapidly, and new users who haven't actively used a legacy model won't have access to it. This is why I updated `anthropic.claude-sonnet-4-6 ` as default and added wildcard IAM policies.

PR #235, opened 2026-04-30
Supply password to integration test CFN template from test runner script

iansrobinson:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/integration-tests/build-tests.sh
+++ b/integration-tests/build-tests.sh
@@ -75,6 +75,7 @@ if [[ "$#" -gt 0 ]]; then
 		echo "             Allowed values:"
 		echo "               nmslib"
 		echo "               faiss"
+    echo "  --db-password <database password>"
 		echo "  --topic <SNS topic name>"
 		echo "  --extraction-llm <Model id or profile name>"
     echo "  --response-llm <Model id or profile name>"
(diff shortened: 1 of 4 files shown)


oussamahansal:
Thanks @iansrobinson. I think probably one small thing we should also add it to the template so users know the option exists
something like:
`export DB_PASSWORD=<OPTIONAL: Database password – used by environments that require a database password (e.g. neo4j-aoss), default: auto-generated>
`

oussamahansal (approved):

integration-tests/build-tests.sh
```diff
@@ -75,6 +75,7 @@ if [[ "$#" -gt 0 ]]; then
 		echo "             Allowed values:"
 		echo "               nmslib"
 		echo "               faiss"
+    echo "  --db-password <database password>"
```
acarbonetto:
```suggestion
		echo "  --db-password <database password>"
```

integration-tests/build-tests.sh
```diff
@@ -411,6 +416,7 @@ if [[ -z "$DRY_RUN" ]]; then
     ParameterKey=NotebookInstanceType,ParameterValue="$NOTEBOOK_INSTANCE_TYPE" \
 		ParameterKey=IamPolicyArn,ParameterValue="$ADDITIONAL_IAM_POLICY_ARN" \
 		ParameterKey=SSHCIDR,ParameterValue="$SSHCIDR" \
+    ParameterKey=DbPassword,ParameterValue="$DB_PASSWORD" \
```
acarbonetto:
```suggestion
		ParameterKey=DbPassword,ParameterValue="$DB_PASSWORD" \
```

acarbonetto (commented):
Please also add DB_PASSWORD to the env.template - which is a better include than the command line password.

acarbonetto (approved):

PR #244, opened 2026-05-07
[FIX] Skip docs site deployment in forks

mykola-pereyma:
## Description

The docs.yml workflow fails in forks because GitHub Pages is not configured. The `deploy-pages` action requires Pages to be enabled in repository settings, which forks don't have by default.

## Changes

Adds `if: github.repository == 'awslabs/graphrag-toolkit'` condition to both `build` and `deploy` jobs in `.github/workflows/docs.yml`.

## Behavior

| Repository | Before | After |
|-----------|--------|-------|
| awslabs/graphrag-toolkit | Deploys docs ✅ | Deploys docs ✅ (unchanged) |
| Any fork | Fails with permissions error ❌ | Jobs skipped (green) ✅ |

## Testing

- No functional code changes — workflow condition only
- Verified the condition syntax matches GitHub Actions documentation
- Standard pattern used across AWS open-source repositories

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -2,6 +2,12 @@ name: Deploy docs site
 
 # Builds the Astro Starlight site under docs-site/ and publishes it to GitHub Pages.
 # Requires "Settings → Pages → Source: GitHub Actions" to be enabled on the repo.
+#
+# This workflow only runs in the upstream repository (awslabs/graphrag-toolkit).
+# To deploy docs from your fork:
+#   1. Enable GitHub Pages: Settings → Pages → Source → GitHub Actions
+#   2. Remove the 'if: github.repository == ...' conditions from both jobs below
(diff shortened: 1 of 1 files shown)


.github/workflows/docs.yml
```diff
@@ -22,6 +22,7 @@ concurrency:
 
 jobs:
   build:
+    if: github.repository == 'awslabs/graphrag-toolkit'
```
acarbonetto, line 31: doesn't this make it so that we _cannot_ deploy to forks?

mykola-pereyma: that is the intention, we need docs to be be deployed from main repo. I think forks should not start this flow.

oussamahansal (approved):

.github/workflows/docs.yml
mykola-pereyma: add comment how to handle docs site deployment from a fork

PR #285, opened 2026-05-22
[FIX] Correct off-by-one validation in ToolParameters ArgTransform limit

mykola-pereyma:
## Description

Fixes an off-by-one error in the `ToolParameters` validation that incorrectly rejects exactly 3 `ArgTransform` parameters, despite the underlying function supporting 3 placeholder argument slots (`_arg_1`, `_arg_2`, `_arg_3`).

## Changes

- Changed `if len(parameters) == 3` to `if len(parameters) > 3` in `lexical-graph/src/graphrag_toolkit/lexical_graph/protocols/mcp_server.py`
- Added unit tests covering all boundary cases in `lexical-graph/tests/unit/protocols/test_tool_parameters.py`

## Problem

The validation check used `== 3` which rejected exactly 3 parameters (the actual maximum supported) and silently allowed 4+ (which would fail at runtime). Users attempting to use the full 3-parameter capacity received a misleading `ValueError`.

## Testing

9 unit tests added and passing:
- 0, 1, 2, 3 parameters accepted (3 was previously rejected)
- 4, 5 parameters correctly rejected
- Missing parameter name rejected
- Parameters correctly marked as required and not hidden

### DIFF
--- a/lexical-graph/src/graphrag_toolkit/lexical_graph/protocols/mcp_server.py
+++ b/lexical-graph/src/graphrag_toolkit/lexical_graph/protocols/mcp_server.py
@@ -35,7 +35,7 @@ class ToolParameters():
             update_params_function:UpdateParametersFunction = lambda _, query_engine_params: query_engine_params
         ):
 
-        if len(parameters) == 3:
+        if len(parameters) > 3:
             raise ValueError('Maximum number of tool parameters exceeded. You can only supply up to 3 tool parameters.')
         
         self.parameters = parameters
--- a/lexical-graph/tests/unit/protocols/__init__.py
+++ b/lexical-graph/tests/unit/protocols/__init__.py
@@ -1,0 +1,2 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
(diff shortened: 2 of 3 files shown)


oussamahansal (approved):
lgtm

lexical-graph/tests/unit/protocols/__init__.py
```diff
@@ -0,0 +1,2 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
```
acarbonetto, line 2: thanks!

acarbonetto (approved):

lexical-graph/src/graphrag_toolkit/lexical_graph/protocols/mcp_server.py
```diff
@@ -35,7 +35,7 @@ def __init__(
             update_params_function:UpdateParametersFunction = lambda _, query_engine_params: query_engine_params
         ):
 
-        if len(parameters) == 3:
+        if len(parameters) > 3:
```
acarbonetto, line 38: I was going to say it should be:  
```
        if len(parameters) >= 3:
```
But it seems like there was a bug originally, that rejected 3 (when 3 tool parameters is fine...)

PR #288, opened 2026-05-26
chore(ci): bump actions/upload-pages-artifact from 3 to 5 with hidden files preservation

mykola-pereyma:
## Description

Bumps actions/upload-pages-artifact from v3 to v5 and adds `include-hidden-files: true` to preserve the `.nojekyll` file in the docs deployment artifact.

## Changes

- Bumped `actions/upload-pages-artifact@v3` → `@v5` in `.github/workflows/docs.yml`
- Added `include-hidden-files: true` to prevent v4's breaking change from excluding `dist/.nojekyll`

## Problem

The v4 release of upload-pages-artifact excludes dotfiles from artifacts by default. The docs site includes `public/.nojekyll` (copied to `dist/.nojekyll` during build) which signals GitHub Pages to skip Jekyll processing. Without `include-hidden-files: true`, this file would be silently dropped.

Supersedes #286.

## Testing

- Single workflow config change — no functional code affected
- The `include-hidden-files` input is available since v5.0.0

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -42,9 +42,10 @@ jobs:
           cache-dependency-path: docs-site/package-lock.json
       - run: npm ci
       - run: npm run build
-      - uses: actions/upload-pages-artifact@v3
+      - uses: actions/upload-pages-artifact@v5
         with:
           path: docs-site/dist
+          include-hidden-files: true
 
   deploy:
     needs: build


oussamahansal (approved):

.github/workflows/docs.yml
```diff
@@ -42,9 +42,10 @@ jobs:
           cache-dependency-path: docs-site/package-lock.json
       - run: npm ci
       - run: npm run build
-      - uses: actions/upload-pages-artifact@v3
+      - uses: actions/upload-pages-artifact@v5
         with:
           path: docs-site/dist
+          include-hidden-files: true
```
acarbonetto, line 48: We should remove any unnecessary hidden files from the directory (just in case someone adds a .env, or .build output, or .logs, or something else large or containing secure info.  

We can do this by adding: 
```
- name: Remove sensitive hidden files
  run: |
    # Find all top-level hidden items in ./, 
    # exclude your files, and delete the rest
    find ./ -maxdepth 1 -name ".*" ! -name "." ! -name ".nojekyll" ! -name ".config" -exec rm -rf {} +
```

acarbonetto (approved):
thanks!  Please consider cleaning up the directory first, to avoid uploading a bunch of hidden files unnecessarily

.github/workflows/docs.yml
mykola-pereyma: hmm, auto delete might be too defensive - we have to review the code submitted.

PR #291, opened 2026-05-28
[Security Fix]: Fix the Cypher safety bypass vulnerabilities

oussamahansal:
*Description of changes:*
- Fix the Cypher safety bypass vulnerabilities in `GraphQueryRetriever.is_query_safe()` by hardening the method with comment stripping, Unicode normalization, expanded blocklist (including `CALL`), restoring the `block_graph_modification` flag, and adding driver-level read-only enforcement via `NeptuneAnalyticsGraphStore.execute_query()`.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #305, opened 2026-06-05
[FIX] Validate s3_path and escape Cypher labels in Neptune graph stores

noel-improv:
## Description

Two related Cypher-injection fixes in the byokg-rag Neptune graph stores, on top of #291.

1. `read_from_csv` interpolates the user-controlled `s3_path` into a `CALL neptune.load(...)` Cypher string. `CALL neptune.load()` takes no bound parameters, so a crafted `s3_path` containing a single quote can break out and inject Cypher. This validates `s3_path` against a strict allowlist before use.
2. Schema discovery and the node/embedding sinks interpolate label names into backtick-quoted identifiers. A label containing a backtick can break out of the identifier. This escapes backticks before interpolation at every such sink.

Both surfaces are distinct from the `is_query_safe` / blocklist work in #291 and complement it. The schema-discovery queries call the graph store's `execute_query` directly, so they do not route through the retriever-level safety check.

## Changes

- New module-level `_validate_s3_path` (allowlist regex) called at the top of both `read_from_csv` overloads, before any side effect (`_upload_to_s3`, loader job, `execute_query`). `None` passes through to preserve existing caller semantics (for example `as_embedding_index` with no save location). The allowlist accepts valid S3 key characters that cannot break out of a single-quoted Cypher literal.
- New module-level `_escape_cypher_label` that doubles backticks and raises `TypeError` on non-string input. Applied at all five label sinks: `_get_edge_properties`, `_get_node_properties`, `_get_triples`, `BaseNeptuneGraphStore.nodes()`, and `NeptuneAnalyticsGraphStore.get_node_text_for_embedding_input()`. The raw label is still used as the Python-side dict key, so returned schema/grouping dicts keep the original name.
- Removed a dead `triple_template` variable in `_get_triples`.
- Extends the `byokg_cypher_safety.py` integration suite added in #291 with two live canary checks, registered in `byokg.all`: `BYOKGCypherInjectionSafety` (schema-discovery / dynamic-label sinks) and `BYOKGS3PathInjectionSafety` (a breakout `s3_path` is rejected by `read_from_csv` before any `CALL neptune.load()` / bulk-loader job runs, leaving a seeded canary untouched).

## Problem

`s3_path` and dynamic label names reach Cypher positions that cannot be parameterized — `CALL neptune.load()` arguments and backtick-quoted identifiers — so a crafted value can break out and inject Cypher. Escaping (labels) and allowlisting (`s3_path`) are the available mitigations.

Related issue (if any): #

## Testing

Unit: validator and escape suites with red-state proofs (each mitigation patched to a no-op lets the payload reach the sink), payload-diversity parametrization, positive paths, indirect callers, and a source-scanning regression guard that fails if a new backtick sink is added unescaped. Local `pytest byokg-rag/tests/unit/graphstore/` — 109 passed.

Integration Test Result:
image

### DIFF
--- a/integration-tests/byokg.all
+++ b/integration-tests/byokg.all
@@ -1,2 +1,3 @@
 byokg_setup.LoadBYOKGGraph
 byokg_cypher_safety.CypherSafetyCheck
+byokg_cypher_safety.BYOKGCypherInjectionSafety
--- a/integration-tests/byokg.all
+++ b/integration-tests/byokg.all
@@ -1,3 +1,4 @@
 byokg_setup.LoadBYOKGGraph
 byokg_cypher_safety.CypherSafetyCheck
 byokg_cypher_safety.BYOKGCypherInjectionSafety
+byokg_cypher_safety.BYOKGS3PathInjectionSafety
--- a/byokg-rag/src/graphrag_toolkit/byokg_rag/graphstore/neptune.py
+++ b/byokg-rag/src/graphrag_toolkit/byokg_rag/graphstore/neptune.py
@@ -3,6 +3,7 @@
 
 import boto3
 import os
+import re
 from urllib.parse import urlparse
 import json
 import logging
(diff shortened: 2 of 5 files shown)


mykola-pereyma:
@noel-improv , since integration tests has been added please attach result of the byokg integration tests execution

mykola-pereyma (approved):

oussamahansal (approved):

acarbonetto (approved):

integration-tests/byokg.all
```diff
@@ -1,2 +1,4 @@
 byokg_setup.LoadBYOKGGraph
-byokg_cypher_safety.CypherSafetyCheck
\ No newline at end of file
+byokg_cypher_safety.CypherSafetyCheck
+byokg_cypher_safety.BYOKGCypherInjectionSafety
+byokg_cypher_safety.BYOKGS3PathInjectionSafety
```
acarbonetto, line 4: please add all secuirty tests to integration-tests/byokg.short

PR #311, opened 2026-06-09
[Benchmarking]: Add VPC reuse support to integration test infrastructure

oussamahansal:
## Description

Allows running multiple integration test stacks in parallel without hitting the VPC-per-region limit by reusing an existing VPC.

## Changes

- Added `existing-vpc-id` and `existing-subnet-ids` CLI args to build-tests.sh
- Added `ExistingVpcId` and `ExistingSubnetIds` parameters to both CFN templates
- Security groups are always created per-stack
- CFN Rules enforce that both VPC ID and subnet IDs must be provided together or both left empty
- Existing behavior (create new VPC) is unchanged when parameters are omitted

## Testing

## Checklist

---

### DIFF
--- a/examples/lexical-graph/cloudformation-templates/graphrag-toolkit-neptune-db-opensearch-serverless.json
+++ b/examples/lexical-graph/cloudformation-templates/graphrag-toolkit-neptune-db-opensearch-serverless.json
@@ -93,7 +93,7 @@
       "Default": ""
     },
     "ExistingSubnetIds": {
-      "Description": "Optional comma-separated list of existing subnet IDs (minimum 2, spanning at least 2 AZs). If provided, ExistingVpcId must also be provided. Leave empty to create new subnets.",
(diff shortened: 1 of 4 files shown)


examples/lexical-graph/cloudformation-templates/graphrag-toolkit-neptune-db-opensearch-serverless.json
```diff
@@ -1280,7 +1436,20 @@
...
+            "CreateVPC",
+            {
+              "Ref": "Subnet4"
+            },
+            {
+              "Fn::Select": [
+                0,
+                {
+                  "Ref": "ExistingSubnetIds"
+                }
```
mykola-pereyma, line 1449: When creating a new VPC, the notebook uses Subnet4 (a public subnet with outbound internet via IGW). When reusing subnets, this takes the first from `ExistingSubnetIds` — which may be private. If the notebook needs outbound access (pip install, data downloads), this will silently break.

Consider updating the `ExistingSubnetIds` parameter description to note that the first subnet must have outbound internet access.

mykola-pereyma (approved):
LGTM — clean infrastructure change. One minor note inline about the subnet assumption for the SageMaker notebook.

PR #312, opened 2026-06-11
Fix/exclude topics from chunk embedding

aghassel:
### Problem
The chunk node builder included the `topics` metadata in the chunk's **embedding** text,
rendered with the default `{key}: {value}` template — so every chunk vector was prefixed
with a literal `topics: ['...', '...']` string (key label + Python-list repr). That's
formatting noise in the embedding.

### Fix
Add `topics` to the chunk's `excluded_embed_metadata_keys` so chunk embeddings reflect the
chunk **content** only. Topics remain in metadata and available to the LLM
(`excluded_llm_metadata_keys` is unchanged).

### Docs
Adds an "Embedding metadata" section to `graph-model.mdx` documenting what each embeddable
node (chunk / topic / statement) includes vs excludes — closing a previously undocumented gap.

### Tests
Unit test asserting `topics` is excluded from the chunk embedding text.

### DIFF
--- a/docs-site/src/content/docs/lexical-graph/graph-model.mdx
+++ b/docs-site/src/content/docs/lexical-graph/graph-model.mdx
@@ -13,6 +13,7 @@ title: Graph Model
     - [Facts](#facts)
     - [Statements](#statements)
     - [Topics](#topics)
+  - [Embedding metadata](#embedding-metadata)
 
 ### Overview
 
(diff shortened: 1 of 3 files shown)


oussamahansal (approved):
LGTM

acarbonetto (approved):
LGTM 

Although, would it be better to put all these metadata fields under a dict object that is excluded from the index, instead of listing each of the these fields separately?

mykola-pereyma (approved):

aghassel:
Thanks @oussamahansal @acarbonetto @mykola-pereyma for the reviews.

@acarbonetto — good suggestion. Grouping the excluded fields under a single nested dict (excluded from the index as one object) rather than listing each field would be cleaner. I'd prefer to do that as a separate follow-up though, because it changes the metadata shape that the node builder writes and that the retrieval/source-resolution paths read, so it warrants its own change + migration consideration rather than expanding the scope of this fix. I'll open a follow-up issue to track it. This PR keeps the explicit field list to stay minimal and backward-compatible.

Happy to merge as-is if there are no other concerns.

PR #315, opened 2026-06-11
[FIX] Escape Cypher labels in lexical-graph graph stores

noel-improv:
## Description

The lexical-graph graph stores interpolate node/relationship label names directly into backtick-quoted Cypher identifiers (`` :`{label}` ``). A label containing a backtick can close the identifier early and inject arbitrary Cypher. This applies the same backtick-escaping mitigation already landed for byokg-rag in #305, but in the lexical-graph package, behind a single shared helper applied at every label-position sink.

The main data-driven surface is `EntityGraphBuilder`: an entity's domain label comes from `label_from(classification)`, and `label_from` deliberately passes `__...__`-style reserved values through **unchanged**, so a crafted classification reaches the backtick-quoted label unescaped.

## Changes

- New `escape_cypher_label()` in `storage/graph/graph_utils.py` — doubles backticks (Cypher's escape rule) and raises `TypeError` on non-string input.
- `indexing/build/entity_graph_builder.py` — escape the domain-entity label, bind `entityId` as a query parameter instead of inlining it into a string literal, and strip newlines from the trailing `//` comment so a crafted label cannot terminate it and append Cypher.
- `storage/vector/neptune_vector_indexes.py` — escape `self.label` in the add/get-embedding queries (previously raw).
- `storage/vector/repair_opensearch_vector_store.py` — escape the derived `label_name`.

**Out of scope (intentional):** the `top_k` / `get_embeddings` vertex-filter values interpolate a tenant-derived label into a *string literal* (not a label identifier) and already strip backticks. That's a different sink, config-derived, so it's left unchanged here.

## Problem

Label names reach Cypher identifier positions that cannot be parameterised, so a value containing a backtick can break out and inject Cypher. `label_from` does not sanitise reserved `__...__` values, so escaping at the sink is required regardless of provenance.

## Testing

- Unit coverage for `escape_cypher_label`, including the `label_from` `__...__` passthrough and non-string `TypeError`.
- A test that drives the **real** `EntityGraphBuilder` with a malicious classification and asserts the emitted domain-entity query is escaped and parameterised (raw label-wrap absent, `$entityId` bound).
- A live FalkorDB canary injection test (`@pytest.mark.integration`, skips automatically when no graph is reachable) with a red-state proof that the pre-patch query deletes a seeded canary while the patched query does not.
- Full lexical-graph unit suite: every touched module passes. (The live FalkorDB test was authored and skip-guarded but not executed in my environment — Docker Hub is unreachable there; it runs wherever a FalkorDB image/registry is available.)

- [ ] Tested manually (live FalkorDB run pending an environment with registry access)

## Checklist

### DIFF
--- a/lexical-graph-contrib/falkordb/tests/conftest.py
+++ b/lexical-graph-contrib/falkordb/tests/conftest.py
@@ -1,0 +1,11 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+
+def pytest_configure(config):
+    """Register the ``integration`` marker so it is known in every run context,
+    whether the contrib package is tested standalone or alongside others."""
+    config.addinivalue_line(
+        'markers',
+        'integration: needs a live FalkorDB; skipped automatically when unreachable',
+    )
--- a/lexical-graph-contrib/falkordb/tests/test_label_injection_live.py
+++ b/lexical-graph-contrib/falkordb/tests/test_label_injection_live.py
@@ -1,0 +1,135 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Live Cypher label-injection test against a real FalkorDB graph.
+
+Exercises the exact query shape that ``EntityGraphBuilder`` emits for a domain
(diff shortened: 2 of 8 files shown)


noel-improv:
Closing — moving this to a draft on my fork (noel-improv/graphrag-toolkit#4) until it's ready for upstream review.

PR #316, opened 2026-06-11
fix(storage): use .keyword sub-field for terms query on metadata id

aghassel:
**Problem.** OpenSearch Serverless dynamically maps string fields as both `text` and `keyword`. The `terms` queries in `get_embeddings` / `_get_existing_doc_ids_for_ids` matched the analyzed `text` sub-field, so id lookups failed for mixed-case ids (e.g. `2023 Q2 NVDA:...`), which the analyzer lowercases and tokenizes.

**Solution.** Query the exact `.keyword` sub-field at both sites. No behavior change for lowercase ids.

**Testing.** Verified id lookups succeed for mixed-case ids (SEC-10Q dataset).

### DIFF
--- a/lexical-graph/src/graphrag_toolkit/lexical_graph/storage/vector/opensearch_vector_indexes.py
+++ b/lexical-graph/src/graphrag_toolkit/lexical_graph/storage/vector/opensearch_vector_indexes.py
@@ -933,7 +933,10 @@ class OpenSearchIndex(VectorIndex):
 
             query = {
                 "terms": {
-                    f'metadata.{INDEX_KEY}.key': [self._clean_id(i) for i in set(id_batch)]
+                    # Use the `.keyword` (exact-match) sub-field, not the analyzed
+                    # `text` field: OpenSearch lowercases/tokenizes the text field,
+                    # so mixed-case/uppercase ids (e.g. SEC-10Q chunk ids) would not match.
+                    f'metadata.{INDEX_KEY}.key.keyword': [self._clean_id(i) for i in set(id_batch)]
                 }
             }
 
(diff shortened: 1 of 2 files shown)


mykola-pereyma (changes requested):
A couple of suggestions before merge:

1. Could you add a **unit test** asserting the query body contains the `.keyword` suffix? Something like mocking the OpenSearch client in `get_embeddings` and checking the constructed query targets `metadata.*.key.keyword`. Prevents regression if the query construction is refactored later.

2. Could you add a **mixed-case ID test case** to the existing integration test suite? e.g. index a document with an ID like `2023_Q2_NVDA:MixedCase` and verify retrieval succeeds. The current integration tests all use lowercase IDs, which is why this wasn't caught earlier.

3. Are there other `terms` queries in this file that match on string metadata fields? If so they may have the same issue.

4. Minor: a brief inline comment explaining why `.keyword` is needed (dynamic mapping analyzes the base field) would help future readers.

aghassel:
Thanks @mykola-pereyma — addressed all four points:

1. **Unit test** — added `tests/unit/storage/vector/test_opensearch_vector_indexes.py`, which mocks the OpenSearch client and asserts the constructed `terms` query targets `metadata.<INDEX_KEY>.key.keyword` for both `get_embeddings` and `_get_existing_doc_ids_for_ids`.
2. **Mixed-case-id test** — `test_mixed_case_ids_preserved_for_exact_match` passes real-world mixed-case ids (`09xZARpUfN7F`, `SEC-10Q`) and asserts they reach the query with case preserved. I also confirmed the test fails if the query is reverted to the analyzed `.key` field, so it genuinely guards the regression.
3. **Audited the other `terms`-on-metadata queries** — there are three. The two id-lookups (`get_embeddings`, `_get_existing_doc_ids_for_ids`) are fixed. The third, in `index_is_available`, is a benign availability probe using a sentinel id (`'xxxxxxx'`) that only checks the index responds — it is not a real id lookup, so it is unaffected; I left it untouched to avoid changing the setup path.
4. **Inline comments** added at both query sites explaining why the `.keyword` (exact-match) sub-field is required.

Full unit suite is green (1625 passed). The one unrelated failure (`test_session_created_without_profile`) pre-exists on the base commit (verified by stashing this change). Squashed into the single commit.

mykola-pereyma (approved):
All feedback addressed — unit tests, mixed-case ID test, and inline comments added. LGTM.

noel-improv (approved):
looks good

PR #318, opened 2026-06-12
feat(retrieval): add TopicBeamSearch retriever (TopicGraphRAG)

aghassel:
**Summary.** Adds `TopicBeamSearch` ("TopicGraphRAG"), a topic-seeded beam-search retriever that seeds from the topic index and beams across the graph to assemble source→topic→statement context, formatted as nested `bedrock_xml`. In our benchmarks it outperformed traversal-native retrieval on every evaluated dataset.

**Configuration** (via `ProcessorArgs`, mirroring `chunk_beam_*`; constructor args override):
- `use_same_chunk_neighbors` / `use_adjacent_chunk_neighbors` (both default `True`) — adjacent-chunk expansion is on by default (consistent benchmark gains, no regressions).
- `use_entity_neighbors` (default `False`) with `max_entity_neighbors` (default `100`, ranked by connection strength) to bound fan-out.
- `topic_beam_width` / `topic_beam_max_depth` / `topic_top_k`.
- `topic_reranker` (`tfidf` / `bedrock`, default `none`) — adds a `RerankTopics` processor that prunes topics by query relevance.

**Design.** Neighbour embeddings are read from the stored topic index by id — the same basis as the seeds (name + statements), with no query-time recomputation. Topics lacking a stored embedding are skipped and logged (a seedable topic index always has them).

**Testing.** ProcessorArgs defaults, override precedence, neighbour query construction, stored-embedding resolution (asserts no recomputation), skip-on-missing, and `RerankTopics`.

### DIFF
--- a/lexical-graph/src/graphrag_toolkit/lexical_graph/retrieval/processors/rerank_topics.py
+++ b/lexical-graph/src/graphrag_toolkit/lexical_graph/retrieval/processors/rerank_topics.py
@@ -1,0 +1,123 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+RerankTopics processor.
+
+Follows the same convention as RerankStatements: the reranking strategy is selected by a
+string arg (``ProcessorArgs.topic_reranker`` -> 'tfidf' | 'bedrock' | 'none') and
+dispatched here, rather than via ad-hoc flags. Where RerankStatements reranks statements
+*within* a topic, this processor reranks whole topics by scoring each topic's
+``name + all its statements`` against the query, keeps the top ``max_topics``, and drops the
+rest before the token-budget truncation runs.
+
+It also propagates each surviving topic's relevance score down to any statement that does not
+already carry a score. This lets the statement reranker be turned off (``reranker='none'``)
+so that statement selection is driven purely by topic relevance -- the ablation that answers
+"if a topic ranks high by the reranker, is statement-level tfidf still needed?".
+"""
+import logging
+import time
+from typing import List, Dict
+
+import boto3
+
+from graphrag_toolkit.lexical_graph import GraphRAGConfig
+from graphrag_toolkit.lexical_graph.retrieval.processors.processor_base import ProcessorBase
+from graphrag_toolkit.lexical_graph.retrieval.processors.processor_args import ProcessorArgs
+from graphrag_toolkit.lexical_graph.retrieval.model import Topic
+from graphrag_toolkit.lexical_graph.utils.reranker_utils import score_values_with_tfidf
+from graphrag_toolkit.lexical_graph.metadata import FilterConfig
+
+from llama_index.core.schema import QueryBundle
+
+logger = logging.getLogger(__name__)
+
+
+class RerankTopics(ProcessorBase):
+    """Reranks and prunes whole topics by query relevance, using the strategy named in
+    ``ProcessorArgs.topic_reranker``. No-op when topic_reranker is unset/'none'."""
+
+    def __init__(self, args: ProcessorArgs, filter_config: FilterConfig):
+        super().__init__(args, filter_config)
+
+    def _topic_text(self, topic: Topic) -> str:
+        stmts = ' '.join([s.statement_str or '' for s in topic.statements])
+        return (f'{topic.topic}\n{stmts}')[:4000]  # cap per-doc length for the reranker
+
+    def _score_with_tfidf(self, texts: List[str], query: QueryBundle) -> List[float]:
+        scored = score_values_with_tfidf(texts, [query.query_str], len(texts))
+        return [scored.get(t, 0.0) for t in texts]
+
+    def _score_with_bedrock(self, texts: List[str], query: QueryBundle) -> List[float]:
+        region = boto3.Session().region_name or getattr(GraphRAGConfig, 'aws_region', None) or 'us-east-1'
+        client = boto3.client('bedrock-agent-runtime', region_name=region)
+        model_arn = f'arn:aws:bedrock:{region}::foundation-model/{GraphRAGConfig.bedrock_reranking_model}'
+        sources = [{'type': 'INLINE', 'inlineDocumentSource': {'type': 'TEXT', 'textDocument': {'text': t}}}
+                   for t in texts]
+        resp = client.rerank(
+            queries=[{'type': 'TEXT', 'textQuery': {'text': query.query_str}}],
+            sources=sources,
+            rerankingConfiguration={
+                'type': 'BEDROCK_RERANKING_MODEL',
+                'bedrockRerankingConfiguration': {
+                    'numberOfResults': len(texts),
+                    'modelConfiguration': {'modelArn': model_arn},
+                },
+            },
+        )
+        scores = [0.0] * len(texts)
+        for r in resp['results']:
+            scores[r['index']] = r['relevanceScore']
+        return scores
+
+    def _process_results(self, search_results, query: QueryBundle):
+        mode = (self.args.topic_reranker or 'none').lower()
+        if mode == 'none':
+            return search_results
+
+        # Flatten (result_index, topic) pairs and build one document per topic.
+        pairs = []
+        texts = []
+        for ri, sr in enumerate(search_results.results):
+            for topic in sr.topics:
+                pairs.append((ri, topic))
+                texts.append(self._topic_text(topic))
+        if not texts:
+            return search_results
+
+        start = time.time()
+        if mode == 'tfidf':
+            scores = self._score_with_tfidf(texts, query)
+        elif mode == 'bedrock':
+            scores = self._score_with_bedrock(texts, query)
+        else:
+            logger.warning(f'Unknown topic_reranker "{mode}", skipping topic rerank')
+            return search_results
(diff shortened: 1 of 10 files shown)


lexical-graph/src/graphrag_toolkit/lexical_graph/retrieval/retrievers/topic_beam_search.py
```diff
@@ -0,0 +1,252 @@
...
+        self._query_embedding = None
+
+    def _init(self, query_bundle: QueryBundle):
+        # Topic beam search does not use entity contexts; skip that extraction to
+        # avoid the extra keyword/entity provider round-trips.
+        return
+
+    # --- Phase 1: seed topics ---
+
+    def get_start_node_ids(self, query_bundle: QueryBundle) -> List[str]:
```
mykola-pereyma, line 101: `_topic_cache` may accumulate embeddings across queries without eviction. If this retriever instance is reused (typical — instantiated once in a query engine), it grows indefinitely. Consider clearing it at the start of each query:

```python
def get_start_node_ids(self, query_bundle: QueryBundle) -> List[str]:
    self._topic_cache = {}  # Reset per query to bound memory
    topic_index = self.vector_store.get_index('topic')
```

lexical-graph/src/graphrag_toolkit/lexical_graph/retrieval/retrievers/topic_beam_search.py
```diff
@@ -0,0 +1,252 @@
...
+        if self.use_entity_neighbors:
+            parts.append(f"""
+            OPTIONAL MATCH (t)<-[:`__BELONGS_TO__`]-(s:`__Statement__`)<-[:`__SUPPORTS__`]-(f:`__Fact__`)<-[:`__SUBJECT__`|`__OBJECT__`]-(e:`__Entity__`)
+            WITH t, COLLECT(DISTINCT e) AS entities
+            UNWIND CASE WHEN size(entities) = 0 THEN [null] ELSE entities END AS entity
+            OPTIONAL MATCH (entity)-[:`__SUBJECT__`|`__OBJECT__`]->(f2:`__Fact__`)-[:`__SUPPORTS__`]->(s2:`__Statement__`)-[:`__BELONGS_TO__`]->(nt:`__Topic__`)
+            WHERE entity IS NOT NULL AND nt <> t
+            WITH t, {self.graph_store.node_id('nt.topicId')} AS ntid, count(f2) AS strength
+            ORDER BY strength DESC
+            WITH t, COLLECT(DISTINCT ntid)[..{int(self.max_entity_neighbors)}] AS entity_neighbors
```
mykola-pereyma, line 197: > Entity neighbors are capped in Cypher via `[..max_entity_neighbors]`, but chunk_neighbors and adjacent_neighbors are unbounded. Is there a practical upper bound on topics per chunk in typical deployments? If not, consider adding a cap for consistency:
>
> ```python
> COLLECT(DISTINCT {self.graph_store.node_id('ct.topicId')})[..200] AS chunk_neighbors
> ```
>
> The beam_width cap on overall search limits total expansion regardless, so this is more of a defensive guard than a correctness issue.

lexical-graph/src/graphrag_toolkit/lexical_graph/retrieval/processors/rerank_topics.py
```diff
@@ -0,0 +1,123 @@
...
+        order = sorted(range(len(pairs)), key=lambda i: scores[i], reverse=True)
+        keep = set(order[: self.args.max_topics])
+
+        kept_topics_by_result: Dict[int, List[Topic]] = {}
+        for i, (ri, topic) in enumerate(pairs):
+            if i not in keep:
+                continue
+            # Propagate topic relevance to statements that have no score yet, so selection can
+            # run on topic relevance alone when the statement reranker is disabled.
+            for stmt in topic.statements:
```
mykola-pereyma, line 109: This mutates Statement.score in-place and reassigns sr.topics directly. I notice existing processors (e.g. TruncateStatements) also mutate in-place, so this is consistent with the current codebase. However, the PR description states "results are rebuilt via model_copy/with_new_results (no in-place mutation)" — if that's the intended direction going forward, this processor should follow suit. Not blocking, just flagging the inconsistency with stated intent.

aghassel:
Thanks @mykola-pereyma — addressed the three points:

1. **`_topic_cache` eviction** — `get_start_node_ids` now calls `self._topic_cache.clear()` at the start of each query, so a reused retriever instance no longer accumulates topic embeddings across queries. Added `test_get_start_node_ids_clears_cache_each_query` (seeds a stale entry, asserts it's evicted on the next query).
2. **Chunk / adjacent-chunk neighbour caps** — added `max_chunk_neighbors` (default 100, mirrors `max_entity_neighbors`) and applied `[..max_chunk_neighbors]` to both the same-chunk and adjacent-chunk `COLLECT`s, so a chunk referenced by many topics can't blow up the neighbour set. Added `test_neighbor_query_caps_chunk_and_adjacent_neighbours` (distinct caps to prove both chunk strategies are bounded). Both new tests verified non-vacuous (they fail if the fixes are reverted).
3. **`RerankTopics` in-place mutation** — you're right that this is consistent with the existing processors (e.g. `TruncateStatements`), and that's intentional. I've corrected the PR description, which over-stated "rebuilt via model_copy"; topic reranking mutates in place by design.

Full unit suite green (1637 passed; the unrelated `test_session_created_without_profile` failure is pre-existing). Squashed into the single commit.

Separately: the neighbour-strategy/scoring **defaults** in this PR were tuned on the older embedding model; I'm running a re-validation on the current embedding to confirm they still hold before relying on them, and will follow up here with results.

aghassel:
**v4 re-validation follow-up** (promised above).

Re-ran the neighbour-strategy + scoring-mode ablation on cohere-embed-v4 graphs (judge sonnet-4-6, 20k-token budget). Accuracy by variant:

| dataset (n) | chunk_only | entity_only | all | no_entity (default) | cosine·chunk_only |
|---|---:|---:|---:|---:|---:|
| musique (18) | 94.44 | 94.44 | 94.44 | 94.44 | 94.44 |
| architecture (25) | 100 | 100 | 100 | 100 | 100 |
| synfile (144) | 19.44 | 20.83 | 20.14 | 18.06 | 18.75 |
| ntsb (49) | 6.12 | 4.08 | 2.04 | 8.16 | 4.08 |

**Honest read: inconclusive on these datasets, but the defaults are not contradicted.**
- musique / architecture **saturate** (every variant identical) — no discrimination on small, easy sets.
- synfile spread is **~2.8 pts (≈4 questions)** — within noise; no clear ordering.
- ntsb **directionally reproduces v3** (chunk_only > entity_only > all; path_weighted > cosine), and the shipped default `no_entity` (same-chunk + adjacent) is **best**, but at 1–4 correct out of 49 it's very noisy.

Across all four, the shipped default is never the worst and is best on ntsb, so I'm keeping it. A clean confirmation would need a PGA-scale, mid-difficulty dataset (where v3 showed a 7-pt separation); the available v4 graphs either saturate or are too noisy. I've documented this caveat rather than overclaiming a v4 confirmation.

mykola-pereyma (approved):
All feedback addressed — topic cache cleared per query, neighbor expansion capped. LGTM.

lexical-graph/src/graphrag_toolkit/lexical_graph/retrieval/retrievers/topic_beam_search.py
```diff
@@ -0,0 +1,258 @@
...
+        missing = [tid for tid in ids if tid not in self._topic_cache]
+        if not missing:
+            return {tid: self._topic_cache[tid] for tid in ids if tid in self._topic_cache}
+
+        try:
+            topic_index = self.vector_store.get_index('topic')
+            # Fetch stored vectors by id through the VectorIndex abstraction (backend-agnostic),
+            # rather than reaching into a concrete client. get_embeddings batches internally.
+            for rec in topic_index.get_embeddings(missing):
+                tid, emb = rec.get('id'), rec.get('embedding')
```
noel-improv: **[potentially blocking]**: can we confirm this, it seems `_get_embeddings` reads `rec.get('id')` but Neptune Analytics' `get_embeddings` returns `{'embedding': …, 'topic': {'topicId': …}}` with no top-level `id`.

we could use this instead: `tid = rec.get('id') or (rec.get('topic') or {}).get('topicId')` 

Also could a Neptune-shaped test be added here to further strengthen coverage

noel-improv (changes requested):
overall logic is sound, just pointed out one potentially blocking item for clarification

aghassel:
Good catch @noel-improv — confirmed and fixed.

You're right: `_get_embeddings` read `rec.get('id')`, which works for the OpenSearch/AOSS backend (its `get_embeddings` returns a top-level `id`) but **not** for Neptune Analytics, whose `get_embeddings` returns the index's return_fields shape — `{'embedding': ..., 'topic': {'topicId': ...}}` — with no top-level `id`. So on a Neptune Analytics vector store every neighbour embedding was silently skipped and beam expansion degraded to seeds-only. (Our benchmarking ran on AOSS, which is why it wasn't caught.)

Fix: resolve the id from either shape —
```python
tid = rec.get('id') or (rec.get('topic') or {}).get('topicId')
```
Added `test_get_embeddings_handles_both_backend_id_shapes` covering both the AOSS (`{'id':...}`) and Neptune (`{'topic':{'topicId':...}}`) record shapes; verified it fails without the fix. Full unit suite green. Squashed into the single commit.

noel-improv (approved):
lgtm, thanks for applying the changes!

PR #319, opened 2026-06-12
feat(retrieval): add TruncateByTokens processor for token-budget context

aghassel:
**Summary.** Adds a `TruncateByTokens` processor that truncates retrieved statements to a token budget (tiktoken `cl100k_base`) rather than a fixed statement count, enabling budget-controlled evaluation and context-window management across retrievers.

**Behavior.**
- Modes: `global_rank` (default) keeps the highest-scoring statements across topics up to the budget; `per_topic_cap` is the token-based analog of `max_statements_per_topic`. The mode is validated (and the tokenizer built) only when a budget is set, so without `max_context_tokens` the processor is a true no-op.
- `max_context_tokens` / `token_truncation_mode` live in `ProcessorArgs`, following the `ProcessorBase(args, filter_config)` contract.
- Auto-wired into the default traversal pipeline when `max_context_tokens` is set: `TruncateByTokens` is inserted immediately **after** the count-based `TruncateStatements` (not replacing it), so `max_statements_per_topic` (coarse) and `max_context_tokens` (fine) both apply. Pipeline is unchanged when no budget is set.
- Rebuilds results via `model_copy` / `with_new_results`, preserving fields like `SearchResult.score`; always keeps ≥1 statement. `per_topic_cap` enforces `max_context_tokens` as a hard global ceiling. The budget counts statement text only (an under-count vs the rendered prompt).

**Testing.** Both modes, unset no-op, keep-at-least-one, and the auto-wire swap.


lexical-graph/src/graphrag_toolkit/lexical_graph/retrieval/processors/truncate_by_tokens.py
```diff
@@ -0,0 +1,146 @@
...
+        if self.mode == 'per_topic_cap':
+            return self._per_topic_cap(search_results)
+        return self._global_rank(search_results)
+
+    def _per_topic_cap(self, search_results: SearchResultCollection) -> SearchResultCollection:
+        """Give each topic an equal share of the token budget."""
+        all_topics = [t for r in search_results.results for t in r.topics]
+        if not all_topics:
+            return search_results
+        per_topic_budget = self.max_tokens // len(all_topics)
```
mykola-pereyma, line 86: The docstring notes per_topic_cap isn't a hard global cap, but from the user's perspective setting `max_context_tokens=3000` creates an expectation of ~3000 tokens out. With many topics + the always-keep-one rule, output can far exceed this. Consider either a second-pass global trim after per-topic selection, or making this clearer in the user-facing docs (the current docs page says "sets a token budget" without qualifying that per_topic_cap can overshoot).

lexical-graph/src/graphrag_toolkit/lexical_graph/retrieval/processors/truncate_by_tokens.py
```diff
@@ -0,0 +1,146 @@
...
+
+    def _global_rank(self, search_results: SearchResultCollection) -> SearchResultCollection:
+        """Flatten all statements, sort by score, fill budget from top, re-group."""
+        all_stmts = []
+        for result in search_results.results:
+            for topic in result.topics:
+                for stmt in topic.statements:
+                    score = stmt.score if hasattr(stmt, 'score') and stmt.score is not None else 0.0
+                    all_stmts.append((score, stmt, topic.topicId, topic.topic, result.source))
+        all_stmts.sort(key=lambda x: x[0], reverse=True)
```
mykola-pereyma, line 124: Worth noting in the docstring that global_rank can concentrate all surviving statements in a single source/topic, potentially undoing the source-diversity work done upstream by `get_diverse_vss_elements`. This is a valid design choice (relevance-first), but users choosing global_rank should understand this tradeoff.

lexical-graph/src/graphrag_toolkit/lexical_graph/retrieval/processors/truncate_by_tokens.py
```diff
@@ -0,0 +1,146 @@
...
+            if topic_id not in source_topics[sk]['topics']:
+                source_topics[sk]['topics'][topic_id] = {'name': topic_name, 'stmts': []}
+            source_topics[sk]['topics'][topic_id]['stmts'].append(stmt)
+
+        new_results = []
+        for data in source_topics.values():
+            topics = [
+                Topic(topicId=tid, topic=tdata['name'], statements=tdata['stmts'])
+                for tid, tdata in data['topics'].items()
+            ]
```
mykola-pereyma, line 162: The PR description states results are rebuilt via `model_copy`/`with_new_results`, and `_per_topic_cap` correctly uses `model_copy(update=...)` to preserve other fields. But `_regroup` constructs new `Topic(...)` and `SearchResult(...)` from scratch, dropping any additional fields (e.g. `SearchResult.score`). Should this store references to the original objects and use `model_copy` for consistency?

lexical-graph/src/graphrag_toolkit/lexical_graph/retrieval/retrievers/traversal_based_base_retriever.py
```diff
@@ -120,7 +120,18 @@ def __init__(self,
...
         self.graph_store = graph_store
         self.vector_store = vector_store
-        self.processors = processors if processors is not None else DEFAULT_PROCESSORS
+        if processors is not None:
+            self.processors = processors
+        else:
+            # When a token budget is configured, swap the count-based TruncateStatements
+            # for the token-based TruncateByTokens; otherwise keep the default pipeline.
+            base_processors = list(DEFAULT_PROCESSORS)
+            if getattr(self.args, 'max_context_tokens', None):
```
mykola-pereyma, line 131: When `max_context_tokens` is set, `TruncateStatements` is fully replaced, so `max_statements_per_topic` silently becomes a no-op. A user passing both would expect both to apply. Consider either keeping both in the pipeline (count-based as coarse pre-filter, token-based as fine trim), or logging a warning when both are configured.

aghassel:
Thanks @mykola-pereyma — all four addressed:

1. **Budget overshoot in `per_topic_cap`** — it now tracks a hard global `max_context_tokens` ceiling across all topics (in addition to the per-topic share), so the total no longer scales with the number of topics. The always-keep-one-per-topic rule can still overshoot by at most one statement, which is documented. Added `test_per_topic_cap_enforces_global_ceiling` (10 topics, asserts the total stays within budget + one statement).
2. **`_regroup` dropping fields** — `_global_rank`/`_regroup` now carry the original `Topic`/`SearchResult` objects and rebuild via `model_copy(update=...)`, so `SearchResult.score` (and any other fields) are preserved. Added `test_global_rank_preserves_search_result_score` (verified non-vacuous — fails if `_regroup` rebuilds from scratch).
3. **`max_statements_per_topic` becoming a no-op** — when `max_context_tokens` is set, `TruncateByTokens` is now inserted **immediately after** `TruncateStatements` rather than replacing it, so the count-based cap (coarse) and the token budget (fine) both apply. Updated the pipeline test to assert both are present and ordered.
4. **`global_rank` vs source diversity** — documented in the docstring that `global_rank` is relevance-first and can concentrate statements in a single source/topic, undoing upstream `get_diverse_vss_elements` diversity; `per_topic_cap` preserves breadth.

Also relaxed one pre-existing identity-check test (`processors is DEFAULT_PROCESSORS` → `==`) since the pipeline installs a defensive copy of the default list. Full unit suite green (1633 passed; the lone `test_session_created_without_profile` failure is pre-existing/unrelated).

mykola-pereyma (approved):
All feedback addressed — global budget ceiling enforced, model_copy preserves fields, both count and token limits apply together. LGTM.

lexical-graph/src/graphrag_toolkit/lexical_graph/retrieval/processors/truncate_by_tokens.py
```diff
@@ -0,0 +1,160 @@
...
+            self._enc = tiktoken.get_encoding('cl100k_base')
+
+    def _count_tokens(self, text: str) -> int:
+        if not text:
+            return 0
+        return len(self._enc.encode(text))
+
+    @staticmethod
+    def _stmt_text(stmt) -> str:
+        return (stmt.statement if hasattr(stmt, 'statement') else str(stmt)) or ''
```
noel-improv, line 72: **[comment]:** This only counts `stmt.statement` text not facts/topic/source formatting added later this will make the rendered context exceed the "budget" might be useful to document it as an undercount;

lexical-graph/src/graphrag_toolkit/lexical_graph/retrieval/processors/truncate_by_tokens.py
```diff
@@ -0,0 +1,160 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+from collections import OrderedDict
+
+import tiktoken
```
noel-improv: **[sugg]**: a lazy import might help this users who'd not set a budget

noel-improv (approved):
budget boundary logic looks good, added a few notes nothing blocking.

aghassel:
Thanks @noel-improv — both addressed (non-blocking, but worth doing):

1. **Budget is an under-count** — you're right that only `stmt.statement` text is counted, not the topic/source headers + context formatting added downstream, so the rendered prompt is somewhat larger than `max_context_tokens`. Documented this explicitly in the docstring (treat the budget as a statement-text target, not a hard ceiling on the final prompt) rather than changing behaviour.
2. **Lazy import** — moved `import tiktoken` out of module scope and into `__init__`, gated behind `max_context_tokens` being set, so callers who don't set a budget don't pay the import.

Full unit suite green. Squashed into the single commit.

mykola-pereyma:
@aghassel  - please resolve conflict

PR #320, opened 2026-06-12
fix(retrieval): guard source-diverse retrieval against indexes missing source

aghassel:
**Problem.** `get_diverse_vss_elements` groups results by `source['sourceId']` for diversity. Correctly-built indexes carry `source`, but topic indexes missing source metadata (inconsistent/legacy builds) cause source-diverse retrieval to fail when the topic index is active.

**Solution.** For elements lacking `source`, lift it from the element's own `metadata['source']` first (no I/O — the normal path). Only when source is absent everywhere, resolve it from the graph (`Topic→Chunk→Source`), memoised in a bounded LRU cache (a topic's source is immutable). Fall back to the un-diversified slice if it can't be determined. `graph_store` is threaded through the four call sites.

This is a defensive guard, not a happy-path change: correctly-built indexes never hit the graph lookup (verified with a freshly built pga topic index returns `source` and diversifies without this fix).

**Testing.** Missing-source regression; metadata-first resolution (asserts no graph call); graph fallback; existing-source preserved; caching; cache bounding/eviction.

### DIFF
--- a/lexical-graph/src/graphrag_toolkit/lexical_graph/retrieval/utils/vector_utils.py
+++ b/lexical-graph/src/graphrag_toolkit/lexical_graph/retrieval/utils/vector_utils.py
@@ -3,6 +3,8 @@
 
 import logging
 import queue
+import threading
+from collections import OrderedDict
 from typing import Optional
 
 from graphrag_toolkit.lexical_graph.metadata import FilterConfig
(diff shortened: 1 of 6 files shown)


lexical-graph/src/graphrag_toolkit/lexical_graph/retrieval/utils/vector_utils.py
```diff
@@ -13,7 +15,86 @@
...
+        WHERE {graph_store.node_id('t.topicId')} IN $topicIds
+        RETURN DISTINCT {graph_store.node_id('t.topicId')} AS topicId,
+               {graph_store.node_id('s.sourceId')} AS sourceId
+        """
+        try:
+            results = graph_store.execute_query(cypher, {'topicIds': ids_to_resolve})
+            for r in results:
+                _cache_put(r['topicId'], r['sourceId'])
+        except Exception as e:
+            logger.warning(f'Failed to resolve topic sources: {e}')
```
mykola-pereyma:
```suggestion
            logger.warning(f'Failed to resolve topic sources: {e}', exc_info=True)
```

mykola-pereyma: nit: exc_info=True gives you the traceback in debug scenarios without changing the fallback behavior.

lexical-graph/tests/unit/retrieval/utils/test_vector_utils.py
```diff
@@ -78,3 +80,76 @@ def test_caps_at_vss_top_k(self):
...
+    """Graph store whose query returns {topicId, sourceId} for requested ids."""
+    gs = MagicMock()
+    gs.node_id.side_effect = lambda x: x
+    gs.execute_query.side_effect = lambda cypher, params: [
+        {'topicId': t, 'sourceId': mapping[t]} for t in params['topicIds'] if t in mapping
+    ]
+    return gs
+
+
+class TestSourceResolution:
```
mykola-pereyma, line 95: The current tests in `TestSourceResolution` cover the extremes well (all elements have source vs. none have source), but the intermediate states aren't exercised. A few cases I'd like to see added to this class:

1. **Partial resolution** — e.g. 4 elements where 2 have source, 1 is resolvable via graph, and 1 is completely unresolvable. This exercises the fallback behavior when resolution is incomplete and would validate that diversity still works for elements that *can* be resolved rather than bailing entirely.

2. **Mixed index type** — call `get_diverse_vss_elements` with `index_name='chunk'` where elements lack source. Currently `_resolve_source_ids` runs a Topic-specific Cypher query regardless of index type, so this would document the expected behavior or reveal a gap.

3. **Mixed source presence in one batch** — elements where some carry `source` directly, some have it only in `metadata`, and some require graph resolution. This validates the priority chain (existing > metadata > graph > fallback) works correctly when all paths fire in one call.

mykola-pereyma (changes requested):

aghassel:
Thanks @mykola-pereyma — addressed all the feedback:

1. **`exc_info=True`** — applied to the `Failed to resolve topic sources` warning.
2. **Mixed index type (the real gap you flagged)** — `_resolve_source_ids` previously ran the Topic→Chunk→Source Cypher regardless of `index_name`, so a `chunk`/`statement` index would issue a query whose ids could never match a `__Topic__` node. Restructured so that:
   - metadata-`source` lift runs first for **any** index (no I/O), and
   - the **graph fallback is gated to `index_name == 'topic'`**; other indexes are left unresolved gracefully rather than firing a bogus query.
3. **New `TestSourceResolution` cases** for the intermediate states:
   - `test_partial_resolution_fills_what_it_can` — one element each resolved via existing source / metadata / graph / unresolvable, in a single batch.
   - `test_non_topic_index_skips_topic_graph_query` — a `chunk` index lifts metadata source but **does not** call the graph (asserts `execute_query` not called). Verified non-vacuous: it fails if the index guard is removed.
   - `test_resolution_priority_existing_then_metadata_then_graph` — validates the `existing > metadata > graph` precedence and that only the genuinely-unresolved id reaches the graph.

Full unit suite green (1631 passed; the lone `test_session_created_without_profile` failure is pre-existing and unrelated). Squashed into the single commit.

mykola-pereyma (approved):
All feedback addressed — exc_info added, partial resolution tests cover intermediate states. LGTM.

noel-improv (approved):
lgtm

PR #328, opened 2026-06-18
[FIX] Escape Cypher labels in lexical-graph graph stores

noel-improv:
## Description

The lexical-graph graph stores interpolate node and relationship label names directly into backtick-quoted Cypher identifiers (`` :`{label}` ``). A label that contains a backtick can close the identifier early and inject arbitrary Cypher. This applies the backtick-escaping mitigation already landed for byokg-rag in awslabs/graphrag-toolkit#305 to the lexical-graph package, behind one shared helper used at every label-position sink.

The main data-driven surface is `EntityGraphBuilder`. An entity's domain label comes from `label_from(classification)`, and `label_from` passes `__...__`-style reserved values through unchanged, so a crafted classification reaches the backtick-quoted label unescaped.

## Changes

- New `escape_cypher_label()` in `storage/graph/graph_utils.py`. Doubles backticks (Cypher's escape rule) and raises `TypeError` on non-string input.
- `indexing/build/entity_graph_builder.py`: escape the domain-entity label, bind `entityId` as a query parameter instead of inlining it into a string literal, and strip newlines from the trailing `//` comment so a crafted label cannot terminate it and append Cypher.
- `storage/vector/neptune_vector_indexes.py`: escape `self.label` in the add and get-embedding queries (previously raw).
- `storage/vector/repair_opensearch_vector_store.py`: escape the derived `label_name`.

Out of scope, intentionally: the `top_k`/`get_embeddings` vertex-filter values interpolate a tenant-derived label into a string literal, not a label identifier, and already strip backticks. That sink is config-derived and left unchanged.

## Problem

Label names reach Cypher identifier positions that cannot be parameterised, so a value containing a backtick can break out and inject Cypher. `label_from` does not sanitise reserved `__...__` values, so escaping at the sink is required regardless of where the label came from.

## Testing

- Unit coverage for `escape_cypher_label`, including the `label_from` `__...__` passthrough and the non-string `TypeError`.
- A test that drives the real `EntityGraphBuilder` with a malicious classification and asserts the emitted domain-entity query is escaped and parameterised (no raw label-wrap, `$entityId` bound).
- A live FalkorDB canary injection test (`@pytest.mark.integration`, skips when no graph is reachable) with a red-state proof: the pre-patch query deletes a seeded canary, the patched query leaves it intact. Point `FALKORDB_URL` at a running FalkorDB to run it; see the package README.
- Full lexical-graph unit suite passes on every touched module.

## Checklist

- [x] No breaking changes

### DIFF
--- a/integration-tests/lexical.short
+++ b/integration-tests/lexical.short
@@ -3,6 +3,7 @@ local_entities.BuildWithoutLocalEntities
 extract.ExtractToFileSystem
 build.BuildFromFileSystem
 build_facts.BuildFacts
+lexical_graph_cypher_safety.LexicalGraphLabelInjectionSafety
 query.TraversalBasedQuery
 query.MetadataFilteringQuery
 query.TraversalBasedQueryWithModelReranker
--- a/integration-tests/test-scripts/graphrag_toolkit_tests/lexical_graph_cypher_safety.py
+++ b/integration-tests/test-scripts/graphrag_toolkit_tests/lexical_graph_cypher_safety.py
@@ -1,0 +1,250 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Verify lexical-graph escapes Cypher labels at the EntityGraphBuilder
+domain-label sink against a live Neptune graph (Analytics or Database).
+
+`label_from` passes `__...__` classifications through unescaped, so a backtick
+in one can close the identifier and inject Cypher. The test drives the real
+EntityGraphBuilder.build() with a malicious classification and checks a seeded
+canary survives. Requires env var GRAPH_STORE (neptune-graph:// or neptune-db://).
+"""
+
+import os
+import unittest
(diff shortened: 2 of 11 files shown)


mykola-pereyma (approved):

mykola-pereyma (approved):
Reviewed on the fork PR (noel-improv#4) — correct escape logic, all sinks covered, thorough testing. LGTM.

acarbonetto (commented):
We should create an security test for this.  The test should follow the same pattern as byokg_cypher_safety.py (which already has a BYOKGCypherInjectionSafety class for byokg-rag). 
Something like: integration-tests/test-scripts/graphrag_toolkit_tests/lexical_graph_cypher_safety.py

The test should include:
- Add a newline injection variant — a classification like "__Evil\n MATCH (c:\Canary`) DETACH DELETE c__"` to confirm the comment-termination defense works end-to-end.
- Test the entityId parameterization — a crafted entity ID with ' to confirm the old string-literal inlining is gone.

acarbonetto (approved):

PR #329, opened 2026-06-18
[FEATURE] Add table extraction to AdvancedPDFReaderProvider

### LINKED ISSUE #284: [FEATURE] Add table extraction support to PDF readers
## Summary

Tables in PDF documents are currently omitted or extracted as unstructured text fragments, losing row/column relationships. This makes the toolkit ineffective for documents that contain significant tabular data (financial reports, technical specifications, regulatory filings, etc.).

## Current Behavior

The toolkit has two PDF readers, neither of which handles tables:

**PDFReaderProvider** (basic):
- Uses LlamaIndex's PyMuPDFReader
- Calls `page.get_text()` which returns text in reading order without preserving table structure
- Tables come out as jumbled text fragments

**AdvancedPDFReaderProvider**:
- Docstring claims "image and table extraction" but only implements image extraction
- Uses `page.get_text()` for text (same limitation as basic reader)
- Uses `page.get_images()` for images
- No table extraction code exists

## Expected Behavior

Tables in PDFs should be extracted with their structure preserved (rows, columns, headers) and converted to a text representation (e.g., markdown table format) that downstream extraction prompts can understand.

## Proposed Solution

PyMuPDF (already a dependency) supports table extraction via `page.find_tables()` (available since pymupdf4llm). The `AdvancedPDFReaderProvider` could be enhanced to:

1. Call `page.find_tables()` on each page
2. Convert extracted tables to markdown format
3. Append the markdown tables to the page text with clear delimiters

Example implementation sketch:

```python
tables = page.find_tables()
for i, table in enumerate(tables):
    df = table.to_pandas()
    text += f"\n\n[TABLE {page_num}_{i}]\n"
    text += df.to_markdown(index=False)
    text += f"\n[/TABLE]\n"
```

## Alternatives Considered

- **Pre-processing PDFs externally**: Users could convert PDFs to markdown before ingestion, but this adds friction and breaks the integrated pipeline
- **Using a multimodal LLM for table extraction**: More accurate but significantly more expensive and slower
- **Using Textract/Document AI**: Cloud-dependent, adds cost and latency

## Additional Context

- File: `lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/load/readers/providers/advanced_pdf_reader_provider.py`
- The AdvancedPDFReaderProvider docstring should be corrected regardless of whether table support is added
- pymupdf's `find_tables()` works well for structured tables but may struggle with complex merged cells

noel-improv:
## Description

`AdvancedPDFReaderProvider` dropped tables or flattened them into unstructured text, losing row/column structure. This extracts tables and appends each as pipe-delimited markdown under a `[TABLE_<page>_<index>]` marker, preserving the grid for downstream indexing, gated by `PDFReaderConfig.extract_tables` (default `True`). `table_count` is recorded in page metadata. Closes #284.

The first commit fixes a prerequisite bug: `read()` referenced a bare `pymupdf` name while the only import was local to `__init__`, so every real read raised `NameError` (caught and rewrapped). The reader was non-functional before this.

## Changes

- `PDFReaderConfig.extract_tables: bool = True` (the basic `PDFReaderProvider` ignores it).
- Hold the pymupdf module on the instance and use it in `read()`, so the reader runs.
- Detect tables per page, strip the table region from the inline text (`get_text()` returns cell text too), then append each table as markdown, so cell content reaches the document once, not twice.
- `table_count` in page metadata.

## Testing

- New `test_advanced_pdf_table_extraction.py`: markdown output with row/column content, opt-out, multi-table page, find_tables/render failure paths, zero-table page, a no-duplication assertion (`count == 1`), and image-block exclusion.
- Hardened the existing ImportError test to mock the absent dependency.
- Full reader suite green.

## Backward compatibility

No breaking API changes. `extract_tables` defaults to `True`; this is the explicitly "advanced" reader and its read path was previously broken, so there is no prior text-only output to preserve. Set `PDFReaderConfig(extract_tables=False)` for text-only.

- [x] Unit tests added
- [x] Existing tests pass

### DIFF
--- a/lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/load/readers/providers/advanced_pdf_reader_provider.py
+++ b/lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/load/readers/providers/advanced_pdf_reader_provider.py
@@ -28,18 +28,83 @@ class AdvancedPDFReaderProvider(LlamaIndexReaderProviderBase, S3FileMixin):
         # so read() cannot reach the bare `pymupdf` name without this.
         self._pymupdf = pymupdf
         self.config = config
+        self.extract_tables = config.extract_tables
         self.metadata_fn = config.metadata_fn
-        logger.debug("Initialized AdvancedPDFReaderProvider")
+        logger.debug(f"Initialized AdvancedPDFReaderProvider with extract_tables={config.extract_tables}")
+
+    def _find_tables(self, page, page_num):
+        """Return the tables detected on a page, or an empty list on failure."""
+        try:
+            return list(page.find_tables().tables)
+        except Exception as e:
+            logger.warning(f"Failed to detect tables on page {page_num}: {e}")
+            return []
+
+    @staticmethod
+    def _block_in_a_table(x0, y0, x1, y1, table_bboxes, threshold=0.5):
+        """True if more than `threshold` of a block's area lies inside any table
+        bbox (default: more than half), i.e. the block belongs to the table."""
+        block_area = max(0.0, x1 - x0) * max(0.0, y1 - y0)
+        if block_area == 0:
+            return False
+        for bx0, by0, bx1, by1 in table_bboxes:
+            overlap_w = max(0.0, min(x1, bx1) - max(x0, bx0))
+            overlap_h = max(0.0, min(y1, by1) - max(y0, by0))
+            if overlap_w * overlap_h > threshold * block_area:
+                return True
+        return False
+
+    def _page_text(self, page, tables):
+        """Page text with table-region text removed.
+
+        get_text() returns each table's cell text inline, and the table is also
+        appended as markdown, so without this the cell content would land in the
+        document twice and inflate downstream extraction (which dedups only on
+        exact-string output). Drop any text block sitting mostly inside a table's
+        bounding box, leaving the markdown as the table's single representation.
+        """
+        if not tables:
+            return page.get_text()
+        table_bboxes = [tuple(table.bbox) for table in tables]
+        kept = []
+        # "blocks" tuple layout: (x0, y0, x1, y1, text, block_no, block_type).
+        # block_type 1 is an image block (placeholder text); images are handled
+        # separately below, so keep only text blocks (type 0).
+        for block in page.get_text("blocks"):
+            x0, y0, x1, y1, block_text = block[:5]
+            block_type = block[6] if len(block) > 6 else 0
+            if block_type != 0:
+                continue
+            if not self._block_in_a_table(x0, y0, x1, y1, table_bboxes):
+                kept.append(block_text)
+        return "".join(kept)
+
+    def _append_tables(self, page_num, tables, text):
+        """Append each table to the text as a markdown block.
+
+        Returns the augmented text and the number of tables rendered. Row/column
+        relationships are preserved via the pipe-delimited markdown that pymupdf's
+        TableFinder produces. The count reflects tables actually written, not just
+        detected, so table_count never overstates what a consumer can find.
+        """
+        rendered = 0
+        for tbl_index, table in enumerate(tables):
+            try:
+                text += f"\n[TABLE_{page_num}_{tbl_index}]\n{table.to_markdown()}"
+                rendered += 1
+            except Exception as e:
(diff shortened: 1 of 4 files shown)


mykola-pereyma (approved):
Reviewed on the fork PR (noel-improv#5) — dedup fix via bbox overlap stripping, split commits, no duplication in output. LGTM.

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/load/readers/providers/advanced_pdf_reader_provider.py
```diff
@@ -18,16 +18,83 @@ class AdvancedPDFReaderProvider(LlamaIndexReaderProviderBase, S3FileMixin):
...
+
+    def _append_tables(self, page_num, tables, text):
+        """Append each table to the text as a markdown block.
+
+        Returns the augmented text and the number of tables rendered. Row/column
+        relationships are preserved via the pipe-delimited markdown that pymupdf's
+        TableFinder produces. The count reflects tables actually written, not just
+        detected, so table_count never overstates what a consumer can find.
+        """
+        rendered = 0
```
acarbonetto, line 92: Page numbering inconsistency in markers vs metadata
- markers use 0-indexed page numbers
- metadata uses 1-index
This could cause debugging issues.  We should make everything 1-indexed (since pdf page numbering usually starts with 1).

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/load/readers/providers/advanced_pdf_reader_provider.py
```diff
@@ -18,16 +18,83 @@ class AdvancedPDFReaderProvider(LlamaIndexReaderProviderBase, S3FileMixin):
...
+
+        get_text() returns each table's cell text inline, and the table is also
+        appended as markdown, so without this the cell content would land in the
+        document twice and inflate downstream extraction (which dedups only on
+        exact-string output). Drop any text block sitting mostly inside a table's
+        bounding box, leaving the markdown as the table's single representation.
+        """
+        if not tables:
+            return page.get_text()
+        table_bboxes = [tuple(table.bbox) for table in tables]
```
acarbonetto:
```suggestion
        table_bboxes = [tuple(t.bbox) for t in tables if t.bbox]
```
`_find_tables` catches exceptions from `find_tables()`, but if a table object has a malformed `bbox` (e.g., `None`), `_page_text` will crash when unpacking `tuple(table.bbox)`. One-line fix:

lexical-graph/tests/unit/indexing/load/readers/providers/test_advanced_pdf_table_extraction.py
```diff
@@ -0,0 +1,167 @@
...
+        page.draw_line((x0, y), (x0 + n_cols * cw, y))
+    for c in range(n_cols + 1):
+        x = x0 + c * cw
+        page.draw_line((x, y0), (x, y0 + n_rows * ch))
+    for r in range(n_rows):
+        for c in range(n_cols):
+            page.insert_text((x0 + c * cw + 5, y0 + r * ch + 20), rows[r][c])
+
+
+def _make_pdf(path, with_table=True):
```
acarbonetto, line 30: All tests use single-page PDFs. A two-page test (page 1 with table, page 2 without) would confirm `page_num` indexing works correctly across pages and that per-page `table_count` is independent.

acarbonetto (approved):
Looks good to me.  Please address comments before merging.

lexical-graph/tests/unit/indexing/load/readers/providers/test_advanced_pdf_table_extraction.py
noel-improv: Ahh yes the test builds a two-page PDF and checks the 1-indexed `page_number`

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/load/readers/providers/advanced_pdf_reader_provider.py
noel-improv: Thank you for this, fixed it so that page numbers are 1-indexed everywhere now

PR #352, opened 2026-06-25
[CHORE] Add gitleaks secret-scanning workflow for pull requests

noel-improv:
## Description
Add a `Secret Scan` workflow that runs gitleaks on every pull request and on a weekly schedule, scanning the full git history for committed secrets and uploading the results to the Security tab. This replaces the earlier git-secrets approach with gitleaks, which is the awslabs org pattern (agent-plugins, aidlc-workflows) and an approved tool.

## Changes
- Added `.github/workflows/secret-scan.yml` with a `gitleaks` job on `ubuntu-latest`
- Installs the MIT-licensed gitleaks binary pinned by version, not the gitleaks-action (which needs a paid license for org repos)
- Runs `gitleaks git` over full history; an optional `.gitleaks.toml` config and `.gitleaks-baseline.json` baseline are picked up if present
- Uploads SARIF to the GitHub Security tab and as a build artifact
- Actions are SHA-pinned; permissions are `contents: read` plus `security-events: write` for the SARIF upload

## Problem
The repo relies on a local git-secrets pre-commit hook, which is opt-in and bypassable with `--no-verify`. Org-level push protection covers AWS and partner-provider tokens, but not generic or custom secrets, private keys, or non-partner token formats. This workflow adds an independent server-side gate for that long tail, with results in the Security tab.

## Testing
- Measured on this repo with gitleaks 8.30.1: full-history scan (1,339 commits, ~208 MB) runs in ~4.2s, working-tree scan in ~1.9s, no leaks found
- Confirmed the run produces a valid SARIF report
- No functional code change; CI workflow only

### DIFF
--- a/.github/workflows/secret-scan.yml
+++ b/.github/workflows/secret-scan.yml
@@ -1,0 +1,96 @@
+# Scans for committed secrets with gitleaks: new commits on pull_request/push, full
+# history on the weekly schedule. Uploads SARIF to the Security tab. Uses the
+# MIT-licensed gitleaks binary, not the licensed action.
+
+name: Secret Scan
+
+on:
+  pull_request:
+    branches: [main]
+  push:
+    branches: [main]
+  schedule:
+    - cron: '17 9 * * 1'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: secret-scan-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  gitleaks:
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    permissions:
+      contents: read
+      security-events: write
+    env:
+      GITLEAKS_VERSION: "8.30.1"
+      # SHA-256 of gitleaks_<version>_linux_x64.tar.gz — update together with the version.
+      GITLEAKS_SHA256: "551f6fc83ea457d62a0d98237cbad105af8d557003051f41f3e7ca7b3f2470eb"
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7
+        with:
+          fetch-depth: 0
+
+      - name: Install gitleaks
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/secret-scan.yml
```diff
@@ -0,0 +1,28 @@
...
+jobs:
+  git-secrets:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+
+      - name: Install git-secrets
+        run: |
+          git clone https://github.com/awslabs/git-secrets /tmp/git-secrets
+          git -C /tmp/git-secrets checkout 7d6b970cbd3c216353cb22b383b70c150140662e
```
acarbonetto: comment to keep this in sync with the hook

.github/workflows/secret-scan.yml
```diff
@@ -0,0 +1,28 @@
...
+          sudo make -C /tmp/git-secrets install
+
+      - name: Register AWS secret patterns
+        run: git secrets --register-aws
```
acarbonetto: we don't need? 
```
          git secrets --install
```

.github/workflows/secret-scan.yml
```diff
@@ -0,0 +1,28 @@
...
+
+      - name: Scan tracked files for secrets
+        run: git ls-files -z | xargs -0 git secrets --scan
```
acarbonetto: do we need the git ls-files? 
```
      - name: Scan repository
        run: |
          git secrets --scan
```

acarbonetto (approved):

mykola-pereyma:
Thanks for thinking about this. Before adding CI time to every PR, worth checking — this repo already has GitHub secret scanning and push protection enabled at the org/repo level, which blocks credentials before they're visible in the PR. Did you check whether the existing setup leaves a gap that git-secrets would cover? If there's a specific pattern git-secrets catches that GitHub's native scanning misses, happy to reconsider.

PR #354, opened 2026-06-25
[FEATURE] Add progress construct for extract/build

### LINKED ISSUE #317: [DRAFT] feat: Add pluggable ProgressMonitor interface for extraction and indexing pipelines
> **⚠️ DRAFT — This issue is under discussion and not yet ready for implementation.**

## Description

Add an abstract `ProgressMonitor` interface that can be passed into the extraction (`.extract()`) and indexing (`.build()`) pipelines to report progress at the document and chunk level across each processing stage (LLM extraction, graph build, vector indexing).

A no-op default implementation should be provided so existing users are unaffected. Consumers needing external progress tracking (e.g. writing counters to DynamoDB for a UI) can supply their own implementation.

## Motivation

When graphrag-toolkit is used as a library within a larger system, the host application needs to track pipeline progress for user-facing status updates. The current logging-based progress (#309) serves operational observability but does not provide a programmatic callback for external state tracking.

## Proposed Interface

```python
from abc import ABC, abstractmethod

class ProgressMonitor(ABC):
    @abstractmethod
    def increment_llm_processed_documents(self, count: int = 1) -> None: ...

    @abstractmethod
    def increment_llm_processed_chunks(self, count: int = 1) -> None: ...

    @abstractmethod
    def increment_graph_processed_documents(self, count: int = 1) -> None: ...

    @abstractmethod
    def increment_graph_processed_chunks(self, count: int = 1) -> None: ...

    @abstractmethod
    def increment_vector_processed_documents(self, count: int = 1) -> None: ...

    @abstractmethod
    def increment_vector_processed_chunks(self, count: int = 1) -> None: ...
```

## Default Implementation

```python
class NoOpProgressMonitor(ProgressMonitor):
    def increment_llm_processed_documents(self, count=1): pass
    def increment_llm_processed_chunks(self, count=1): pass
    def increment_graph_processed_documents(self, count=1): pass
    def increment_graph_processed_chunks(self, count=1): pass
    def increment_vector_processed_documents(self, count=1): pass
    def increment_vector_processed_chunks(self, count=1): pass
```

## Integration Points

| Stage | Method | Calls |
|-------|--------|-------|
| LLM extraction | `PropositionExtractor.extract()`, `TopicExtractor.extract()` | `increment_llm_processed_chunks` per chunk, `increment_llm_processed_documents` per document batch |
| Graph build | `LexicalGraphIndex.build()` graph write phase | `increment_graph_processed_chunks` per chunk written |
| Vector indexing | `LexicalGraphIndex.build()` vector write phase | `increment_vector_processed_chunks` per chunk indexed |

## Acceptance Criteria

- [ ] `ProgressMonitor` ABC defined in `graphrag_toolkit.lexical_graph`
- [ ] `NoOpProgressMonitor` provided as default
- [ ] `LexicalGraphIndex.build()` accepts optional `progress_monitor` parameter
- [ ] Extractors invoke `increment_llm_processed_chunks` on each completed chunk
- [ ] Graph/vector write stages invoke their respective increment methods
- [ ] Existing behavior unchanged when no monitor is passed
- [ ] Unit tests verifying a mock monitor receives correct increment calls

acarbonetto:
## Description

Adds a pluggable ProgressMonitor interface for tracking pipeline progress at the document and chunk level across the LLM extraction, graph build, and vector indexing stages. Existing behavior is unchanged when no monitor is provided.

## Changes

New: lexical_graph/indexing/progress_monitor.py
  - ProgressMonitor — abstract base class with 6 increment methods (documents + chunks × 3 stages)
  - NoOpProgressMonitor — concrete no-op subclass; users needing only a subset of callbacks can subclass this instead of implementing all 6 abstract methods

Modified: LexicalGraphIndex.extract(), .build(), .extract_and_build()
  - Added optional progress_monitor: Optional[ProgressMonitor] = None parameter to all three methods
  - LLM stage: monitor receives increment_llm_processed_documents and increment_llm_processed_chunks after each SourceDocument completes extraction
  - Build stage: monitor receives graph and vector increment calls after each document batch completes in BuildPipeline.build(); graph and vector increments fire together since both stages run in the same worker process

Exported from graphrag_toolkit.lexical_graph for top-level discoverability.

### Design Notes

  - Granularity: Chunk counts are reported at document boundaries (all chunks for a document reported at once), not as each individual chunk finishes within a worker process. This avoids multiprocessing complexity. A TODO comment marks where true per-chunk streaming could be added in future.
  - Thread safety: Monitor methods are called only from the main process/thread (at document yield boundaries after results return from workers). No thread safety required of implementations.
  - Error isolation: All monitor calls are wrapped in try/except — a monitor that throws will log a warning and never crash the pipeline.

### Usage Example

```
  from graphrag_toolkit.lexical_graph import NoOpProgressMonitor

  class MyMonitor(NoOpProgressMonitor):
      def __init__(self):
          self.docs_done = 0

      def increment_llm_processed_documents(self, count=1):
          self.docs_done += count
          print(f"Extracted {self.docs_done} documents")

  index.extract_and_build(nodes, progress_monitor=MyMonitor())
```

## Problem

Related issue (if any): #

fixes: #317 

## Testing

Added `tests/unit/test_progress_monitor.py`

- [X] Unit tests added/updated
- [X] Integration tests added (as appropriate)
- [X] Existing tests pass (`pytest`)
- [X] Tested manually (describe below)

Tested manually using integration tests, and added progress monitory (logging) to workshop.  Debugging output shows: 

```

```

## Checklist

---

### DIFF
--- a/lexical-graph/src/graphrag_toolkit/lexical_graph/lexical_graph_index.py
+++ b/lexical-graph/src/graphrag_toolkit/lexical_graph/lexical_graph_index.py
@@ -35,6 +35,7 @@ from graphrag_toolkit.lexical_graph.indexing.build.null_builder import NullBuild
 from graphrag_toolkit.lexical_graph.indexing.build.delete_sources import DeleteSources
 from graphrag_toolkit.lexical_graph.utils.arg_utils import coalesce
 from graphrag_toolkit.lexical_graph.utils.llm_cache import LLMCache
+from graphrag_toolkit.lexical_graph.indexing.progress_monitor import ProgressMonitor
 
 from llama_index.core.node_parser import SentenceSplitter, NodeParser
 from llama_index.core.schema import BaseNode
@@ -428,6 +429,7 @@ class LexicalGraphIndex():
             handler: Optional[NodeHandler] = None,
             checkpoint: Optional[Checkpoint] = None,
             show_progress: Optional[bool] = False,
+            progress_monitor: Optional[ProgressMonitor] = None,
             **kwargs: Any) -> None:
         """
         Executes the extraction process for a given set of nodes using the specified handler,
@@ -443,7 +445,26 @@ class LexicalGraphIndex():
                 progress during extraction and build stages.
             show_progress (Optional[bool], optional): Indicates whether to display progress during
                 the pipeline execution.
+            progress_monitor (Optional[ProgressMonitor], optional): A monitor to receive progress
+                callbacks as documents complete LLM extraction. Receives
+                increment_llm_processed_documents and increment_llm_processed_chunks calls at
+                document boundaries.
             **kwargs (Any): Additional keyword arguments for pipeline configurations or overrides.
+
+        Example::
+
+            from graphrag_toolkit.lexical_graph import ProgressMonitor, NoOpProgressMonitor
+
+            class MyMonitor(NoOpProgressMonitor):
+                def __init__(self):
+                    self.docs_done = 0
+
+                def increment_llm_processed_documents(self, count=1):
+                    self.docs_done += count
+                    print(f"Extracted {self.docs_done} documents")
+
+            monitor = MyMonitor()
+            index.extract(nodes, progress_monitor=monitor)
         """
 
         if not self.tenant_id.is_default_tenant():
@@ -471,10 +492,17 @@ class LexicalGraphIndex():
             **kwargs
         )
 
-        if handler:
-            nodes | extraction_pipeline | Pipe(handler.accept) | build_pipeline | sink
+        if progress_monitor:
+            extraction_monitor = self._create_extraction_monitor_pipe(progress_monitor)
+            if handler:
+                nodes | extraction_pipeline | extraction_monitor | Pipe(handler.accept) | build_pipeline | sink
+            else:
+                nodes | extraction_pipeline | extraction_monitor | build_pipeline | sink
         else:
-            nodes | extraction_pipeline | build_pipeline | sink
+            if handler:
+                nodes | extraction_pipeline | Pipe(handler.accept) | build_pipeline | sink
+            else:
+                nodes | extraction_pipeline | build_pipeline | sink
 
     def build(
             self,
@@ -482,6 +510,7 @@ class LexicalGraphIndex():
             handler: Optional[NodeHandler] = None,
             checkpoint: Optional[Checkpoint] = None,
             show_progress: Optional[bool] = False,
+            progress_monitor: Optional[ProgressMonitor] = None,
             **kwargs: Any) -> None:
         """
         Builds an indexing pipeline for processing nodes, constructing a graph, and creating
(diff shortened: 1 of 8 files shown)


mykola-pereyma (approved):
Follows #317 faithfully. Document-boundary granularity is a justified trade-off over per-chunk IPC. Error isolation, single-threaded guarantee, and forward-compatible interface all verified. LGTM.

lexical-graph/tests/unit/test_progress_monitor.py
```diff
@@ -0,0 +1,288 @@
...
+        # Simulate the extraction monitor pipe firing (LLM stage)
+        from graphrag_toolkit.lexical_graph.lexical_graph_index import LexicalGraphIndex
+        from llama_index.core.schema import TextNode
+
+        doc = SourceDocument(nodes=[TextNode(text="a", id_="a1"), TextNode(text="b", id_="a2")])
+
+        extraction_monitor = LexicalGraphIndex._create_extraction_monitor_pipe(monitor)
+        list([doc] | extraction_monitor)
+
+        # Simulate the build monitor firing (build stage)
```
noel-improv, line 247: **[non-blocking]**  should this verify the real method routes via `extract_and_build` instead of calling the build increments directly so that it matches the integration test pattern

lexical-graph/src/graphrag_toolkit/lexical_graph/lexical_graph_index.py
```diff
@@ -591,11 +632,32 @@ def extract_and_build(
...
+            extraction_monitor = self._create_extraction_monitor_pipe(progress_monitor)
+            nodes | extraction_pipeline | extraction_monitor | build_pipeline | sink_fn
+        else:
+            nodes | extraction_pipeline | build_pipeline | sink_fn
+
+    @staticmethod
+    def _create_extraction_monitor_pipe(progress_monitor: ProgressMonitor) -> Pipe:
+        def _monitor_extraction(source_documents):
+            for doc in source_documents:
+                # TODO: Chunk-level reporting is batched at document boundaries.
```
noel-improv, line 650: **[nit]** This four-line _TODO_ might have been mistakenly duplicated in `build_pipeline.py` (the build-stage block at line 330) please confirm.

lexical-graph/tests/unit/test_progress_monitor.py
```diff
@@ -0,0 +1,288 @@
...
+        monitor.increment_graph_processed_documents(2)
+
+        assert monitor.doc_count == 5
+
+
+class TestExtractionMonitorPipe:
+
+    def test_extraction_monitor_fires_correctly(self):
+        from unittest.mock import MagicMock
+        from pipe import Pipe
```
noel-improv: **[nit]** `from pipe import Pipe` is unused here (and again at line 94). Safe to drop.

noel-improv (commented):
This is clean and well-scoped; no blockers from me except two non-blocking notes worth a look

noel-improv (approved):

lexical-graph/src/graphrag_toolkit/lexical_graph/lexical_graph_index.py
acarbonetto: It applies to both places (same comment)

lexical-graph/tests/unit/test_progress_monitor.py
acarbonetto: we could, but that would take a lot longer?  The purpose of the test is to see the progress monitor.

PR #362, opened 2026-06-29
fix(lexical-graph): parameterize SQL filters and id lists in PGVector store

noel-improv:
## Description

Bind metadata filter values, metadata keys, and id lists in the PGVector store (`pg_vector_indexes.py`) as psycopg2 parameters instead of string-interpolating them into SQL. A filter value or id containing a single quote could break out of its literal; because `get_embeddings`, `update_versioning`, and `delete_embeddings` build raw `IN (...)` lists the same way, the blast radius includes `UPDATE` and `DELETE`, not only `SELECT`. The store already binds the embedding vector and `LIMIT` with `%s`; this applies the same pattern to the remaining sinks.

## Changes

- `parse_metadata_filters_recursive` and `filter_config_to_sql_filters` now return `(clause, params)`; filter values become `%s` placeholders and the metadata key is bound into the JSON path operand (`->>%s`).
- `top_k` splices the filter params between the embedding and `LIMIT` placeholders.
- `get_embeddings`, `update_versioning`, and `delete_embeddings` build `IN (%s, ...)` placeholder lists; `update_versioning` also binds its timestamp.
- `formatter_for_type` coerces a value for binding rather than quoting it into SQL.
- The id sinks return early on an empty list, so an empty `ids` no longer produces `IN ()` (a Postgres syntax error).
- Only the type cast and the table/schema/index identifiers (sourced from config, not user input) stay interpolated, so the existing `# nosec B608` annotations are kept, with justifications still accurate.

## Problem

SQL injection in the PGVector store: filter values, metadata keys, and id lists were interpolated directly into query strings instead of bound as parameters.

Related issue (if any): #

## Testing

New `lexical-graph/tests/unit/storage/vector/test_pg_vector_injection.py` drives each sink against a mocked cursor and asserts the payload reaches `cur.execute` as a bound parameter, never as inline SQL. Covers the filter-value, filter-key, and three id-list sinks; a parametrized set of break-out payloads (tautology, stacked statement, comment terminator, null byte, unicode quote, backslash, newline); the clause-builder `(clause, params)` contract; the empty-id guard; and a positive path. Before the fix the injection tests fail and print the actual injected SQL; after the fix they pass. `tests/unit/storage` and the related filter/metadata/graph suites pass unchanged; bandit `-ll` reports no medium+ findings.

`integration-tests/test-scripts/graphrag_toolkit_tests/lexical_graph_pgvector_safety.py`, wired into `integration-tests/lexical.short`, runs the real PGIndex sinks against a live Postgres+pgvector engine. It confirms injected filter values, metadata keys, and id lists are inert, and that the pre-fix interpolation pattern still breaks out on the same engine. Skips when the backend isn't configured for the run.

## Checklist

---


noel-improv:
Fix validation, an a short live integration test against a PGVector-backed store to confirm the fix end-to-end
-   Config used the `pg_vector_store` backend so the patched `pg_vector_indexes.py` sinks were actually exercised against a real DB.
-   Confirmed the previously vulnerable paths, filter values, metadata keys, and the `IN (...)` id lists in `get_embeddings`, `update_versioning`, and `delete_embeddings`, now bind values as psycopg2 parameters; payloads containing single quotes are stored/queried as literal data instead of breaking out into SQL.
-   Verified normal read/write/delete behavior is unchanged.

mykola-pereyma (approved):
Correct parameterization of all SQL sinks — filter values, metadata keys, and id lists all bound via psycopg2 %s placeholders. Thorough red-state tests. LGTM.

acarbonetto (commented):
Fix looks good.  We need a security fix integration test for review to be complete.

aghassel:
Approach LGTM. Parameterizing values, keys, and id-lists is the right fix. Two notes, both non-blocking; the first is a real latent bug.

**[bug, non-blocking] Empty id list produces `IN ()` (Postgres syntax error).** In `get_embeddings`/`update_versioning`/`delete_embeddings`, `placeholders = ','.join(['%s'] * len(unique_ids))` is `''` when `ids` is empty, so the query becomes `... WHERE chunkId IN ();`, which Postgres rejects. Not a regression (the old `format_ids` also produced `IN ()`), but worth guarding while we're here:

```python
unique_ids = list(set(ids))
if not unique_ids:
    return []        # delete_embeddings: return ids
```

Verified: `ids=[]` yields `WHERE chunkId IN ();`.

**[optional] Identifiers still interpolated.** `schema_name`/`index_name` remain f-string-interpolated. You've documented them as internal config in the `nosec`, which is reasonable. If they can ever derive from tenant input, consider an allowlist (`^[A-Za-z_][A-Za-z0-9_]*$`) or `psycopg2.sql.Identifier`.

noel-improv:
> Approach LGTM. Parameterizing values, keys, and id-lists is the right fix. Two notes, both non-blocking; the first is a real latent bug.
> 
> **[bug, non-blocking] Empty id list produces `IN ()` (Postgres syntax error).** In `get_embeddings`/`update_versioning`/`delete_embeddings`, `placeholders = ','.join(['%s'] * len(unique_ids))` is `''` when `ids` is empty, so the query becomes `... WHERE chunkId IN ();`, which Postgres rejects. Not a regression (the old `format_ids` also produced `IN ()`), but worth guarding while we're here:
> 
> ```python
> unique_ids = list(set(ids))
> if not unique_ids:
>     return []        # delete_embeddings: return ids
> ```
> 
> Verified: `ids=[]` yields `WHERE chunkId IN ();`.
> 
> **[optional] Identifiers still interpolated.** `schema_name`/`index_name` remain f-string-interpolated. You've documented them as internal config in the `nosec`, which is reasonable. If they can ever derive from tenant input, consider an allowlist (`^[A-Za-z_][A-Za-z0-9_]*$`) or `psycopg2.sql.Identifier`.

thank you @aghassel, applied the fix, `get_embeddings`, `update_versioning` and `delete_embeddings` now return early on an empty list, before opening a connection, so there's no `IN ()` and no wasted connection.

integration-tests/security/test_pgvector_sql_injection.py
```diff
@@ -0,0 +1,171 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Security integration test: SQL injection in the PGVector store.
+
+Runs with its parent fix PR. Drives the real PGIndex sinks against a live
```
mykola-pereyma: What PR you are referring to?

mykola-pereyma (changes requested):

aghassel:
Confirmed: returning early before opening the connection is cleaner than what I suggested (it avoids the wasted connection too), and the three empty-list tests cover it. Thanks for the fast turnaround. LGTM.

oussamahansal (approved):
lgtm

.github/workflows/security-integration-pgvector.yml
```diff
@@ -0,0 +1,50 @@
+# Runs the SQL injection integration test for the PGVector store against a live
+# Postgres+pgvector engine (service container), so the fix is demonstrated
+# working on release and on PRs that touch the sink or the test.
+name: Security Integration - PGVector SQL injection
```
acarbonetto: we don't need a workflow to run integration tests - most integration tests require running resources in an environment and should be kicked off manually. 

lets remove this until we have a plan to run all tests in CI.

integration-tests/security/test_pgvector_sql_injection.py
```diff
@@ -0,0 +1,171 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Security integration test: SQL injection in the PGVector store.
```
acarbonetto: The existing integration tests under test-scripts/graphrag_toolkit_tests/ use a custom framework: they extend IntegrationTestBase, implement _run_test() and description, and are orchestrated via test_suite.py on a SageMaker notebook. This test is a standalone pytest file that runs via a GitHub Actions workflow with a service container — it's a completely separate execution path.  

Please follow existing test framework when adding integ tests.

integration-tests/security/test_pgvector_sql_injection.py
```diff
@@ -0,0 +1,171 @@
...
+
+Drives the real PGIndex sinks against a live Postgres+pgvector engine and shows
+the fix working: legitimate filters resolve, injected filter values, metadata
+keys, and id lists are inert, and the pre-fix interpolation pattern still breaks
+out on the same engine.
+
+Backend: PGVECTOR_TEST_DSN. Skips when unset; CI provides a Postgres+pgvector
+service container.
```
acarbonetto: Add instructions explaining how to run this test (which templates/variables to use to include PGVector)

acarbonetto (changes requested):
Please update the security test to include it as part of the https://github.com/awslabs/graphrag-toolkit/blob/main/integration-tests/byokg.short integration test

mykola-pereyma (approved):
All feedback addressed. Integration test ported to IntegrationTestBase framework, registered in lexical.short, standalone workflow removed, README updated. Source fix is correct parameterization across all four sinks. Comprehensive unit tests with diverse payloads and red-state methodology. Docstrings are meaningful and security-aware. LGTM.

PR #367, opened 2026-06-30
Fix: GitHub reader hangs notebook when GITHUB_TOKEN is unset

### LINKED ISSUE #363: [BUG] Notebook 01 hangs when GITHUB_TOKEN is unset
Running the lexical-graph-local-dev notebook 01 end to end with no `GITHUB_TOKEN` set causes the run to hang. It stalls on the PPTX reader cell, but the cause is the GitHub reader cell a couple of cells earlier.

When `GITHUB_TOKEN` is missing, the GitHub cell still builds the client and calls `load_data`:

    github_token = os.environ.get('GITHUB_TOKEN', '')
    github_client = GithubClient(github_token=github_token, ...)
    github_docs = reader.load_data(branch="main")

The `if github_token:` check just above only wraps a print, so the client gets built with an empty string either way. The empty string gets past llama-index's `if github_token is None` guard, and the request goes out with an `Authorization: Bearer ` header that httpx rejects:

    LocalProtocolError: Illegal header value b'Bearer '

The call runs under `nest_asyncio`, and when it fails it leaves the httpx connection pool in a bad state. The next reader cell that spawns worker processes (PPTX) inherits that state, a worker never exits, and the pool shutdown blocks. That is why PPTX looks like the problem when it isn't.

One more thing: `GithubClient` has no unauthenticated path. Passing `None` raises a `ValueError` asking for a token, so the cell's "unauthenticated access" comment doesn't hold. Without a token the cell can't do anything useful and should be skipped.

Steps to reproduce:

1. Leave `GITHUB_TOKEN` unset in `notebooks/.env`.
2. Run notebook 01 end to end: `jupyter nbconvert --to notebook --execute`.
3. The GitHub cell raises the `LocalProtocolError` above, then the PPTX cell hangs until the cell timeout.

Expected: with no token the GitHub cell skips and the rest of the notebook runs, the same way the S3 and Bedrock-prompt cells skip when they aren't configured.

Suggested fix: gate the GitHub cell body on a token being present and skip when it's missing. With that change the notebook runs start to finish.

### LINKED ISSUE #366: lexical-graph-local-dev notebook 01 hangs when GITHUB_TOKEN is unset
## Summary

Running the `lexical-graph-local-dev` notebook `01-Combined-Extract-and-Build.ipynb` end-to-end with no `GITHUB_TOKEN` set causes the notebook to hang. It stalls on the PPTX reader cell, but the root cause is the GitHub reader cell earlier.

## Root Cause

When `GITHUB_TOKEN` is missing, the GitHub cell (cell 15) still builds the client and calls `load_data`:

```python
github_token = os.environ.get('GITHUB_TOKEN', '')
# if github_token: check only wraps a print
github_client = GithubClient(github_token=github_token, verbose=True)
github_docs = reader.load_data(branch="main")
```

The empty string passes llama-index's `if github_token is None` guard, and the request goes out with an `Authorization: Bearer ` header that httpx rejects:

```
LocalProtocolError: Illegal header value b'Bearer '
```

The call runs under `nest_asyncio`, and when it fails it leaves the httpx connection pool in a bad state. The next reader cell (PPTX) inherits that state, a worker never exits, and pool shutdown blocks.

## Additional Note

The library code in `github_reader_provider.py` has the same potential: it passes `self.github_config.github_token` directly to `GithubClient()` without validating non-empty.

## Steps to Reproduce

1. Leave `GITHUB_TOKEN` unset in `notebooks/.env`
2. Run: `jupyter nbconvert --to notebook --execute 01-Combined-Extract-and-Build.ipynb`
3. GitHub cell raises `LocalProtocolError`, PPTX cell hangs

## Expected

With no token, the GitHub cell skips and the notebook runs to completion.

## Suggested Fix

**Notebook:** Gate the cell body on token being present:
```python
github_token = os.environ.get('GITHUB_TOKEN', '').strip()
if not github_token:
    print('GITHUB_TOKEN not set — skipping GitHub reader.')
    github_docs = []
else:
    # ... existing code
```

**Library (`github_reader_provider.py`):**
```python
def read(self, input_source) -> List[Document]:
    if not self.github_config.github_token:
        logger.warning('No GitHub token configured — skipping')
        return []
```

evanerwee:
## Summary

Fixes #366. Fixes #363. When `GITHUB_TOKEN` is unset, the `lexical-graph-local-dev` notebook 01 hangs on the PPTX cell because the GitHub reader cell leaves httpx in a bad state.

## Changes

**Notebook** (`examples/lexical-graph-local-dev/notebooks/01-Combined-Extract-and-Build.ipynb`):
- Gate the GitHub reader block on `github_token` being non-empty
- When missing: skip with message, set `github_docs = []`
- Gate the `extract_and_build` call on `github_docs` being non-empty
- Notebook now runs end-to-end without a token

**Library** (`lexical-graph/.../github_reader_provider.py`):
- Add early return in `read()` when `github_token` is falsy
- Returns `[]` with a warning log instead of failing on the HTTP call
- Consistent with how other readers skip when unconfigured

## Root Cause

`os.environ.get('GITHUB_TOKEN', '')` produces an empty string (not `None`).
The empty string passes llama-index's `if github_token is None` guard.
httpx rejects the `Authorization: Bearer ` header (trailing space, no token).
The async failure under `nest_asyncio` corrupts the connection pool.
The next cell (PPTX) inherits the bad state and hangs.

## Testing

- Unit tests added for `GitHubReaderProvider.read()` with empty/None token:
  - `test_read_returns_empty_list_when_token_missing` — empty string token
  - `test_read_returns_empty_list_when_token_none` — None token
  - Both assert `result == []` and verify warning is logged
- Manual notebook execution verified:
  ```bash
  cd examples/lexical-graph-local-dev
  unset GITHUB_TOKEN
  jupyter nbconvert --to notebook --execute notebooks/01-Combined-Extract-and-Build.ipynb
  ```
- Verified existing `test_raises_exception_if_dependencies_not_installed` still passes

## Checklist

- [x] Code follows project style and conventions
- [x] Unit tests added for new behavior
- [x] Existing tests unaffected (no breaking changes)
- [x] Documentation updated (notebook comments explain the guard)
- [x] No new dependencies introduced
- [x] Backwards compatible — token-present behavior unchanged


mykola-pereyma:
Thanks, we will review you PR shortly.

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/load/readers/providers/github_reader_provider.py
```diff
@@ -31,6 +31,10 @@ def __init__(self, config: GitHubReaderConfig):
     def read(self, input_source) -> List[Document]:
         """Read GitHub repository documents with metadata handling."""
         
+        if not self.github_config.github_token:
+            logger.warning("No GitHub token configured — skipping GitHub read. "
+                          "Set github_token in GitHubReaderConfig to enable.")
+            return []
```
mykola-pereyma, line 37: This needs a unit test. Same pattern as other reader providers in the test suite — mock the config with empty token, call read(), assert returns empty list and logs warning. Without it nothing prevents regression if someone refactors this guard away later.

mykola-pereyma (changes requested):
Thanks for the fix — root cause analysis is solid and the two-layer approach (notebook + library) is right thing to do. Few things before merge:

Also the PR template sections for Testing and Checklist are not filled in — could you add those? Repository requires them for all submissions.

And could you add Fixes #363 to the PR description? That issue describes same bug from different reporter and should auto-close together with #366 when this merges.

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/load/readers/providers/github_reader_provider.py
evanerwee: 1. PR Description (edit at https://github.com/awslabs/graphrag-toolkit/pull/367 (https://github.com/awslabs/graphrag-toolkit/pull/367) → click "Edit" on the description):
  
  ## Summary
  
  Fixes #366. Fixes #363. When `GITHUB_TOKEN` is unset, the `lexical-graph-local-dev` notebook 01 hangs on the PPTX cell because the GitHub reader cell leaves httpx in a bad state.
  
  ## Changes
  
  **Notebook** (`examples/lexical-graph-local-dev/notebooks/01-Combined-Extract-and-Build.ipynb`):
  - Gate the GitHub reader block on `github_token` being non-empty
  - When missing: skip with message, set `github_docs = []`
  - Gate the `extract_and_build` call on `github_docs` being non-empty
  - Notebook now runs end-to-end without a token
  
  **Library** (`lexical-graph/.../github_reader_provider.py`):
  - Add early return in `read()` when `github_token` is falsy
  - Returns `[]` with a warning log instead of failing on the HTTP call
  - Consistent with how other readers skip when unconfigured
  
  ## Root Cause
  
  `os.environ.get('GITHUB_TOKEN', '')` produces an empty string (not `None`).
  The empty string passes llama-index's `if github_token is None` guard.
  httpx rejects the `Authorization: Bearer ` header (trailing space, no token).
  The async failure under `nest_asyncio` corrupts the connection pool.
  The next cell (PPTX) inherits the bad state and hangs.
  
  ## Testing
  
  - Unit tests added for `GitHubReaderProvider.read()` with empty/None token:
    - `test_read_returns_empty_list_when_token_missing` — empty string token
    - `test_read_returns_empty_list_when_token_none` — None token
    - Both assert `result == []` and verify warning is logged
  - Manual notebook execution verified:
    ```bash
    cd examples/lexical-graph-local-dev
    unset GITHUB_TOKEN
    jupyter nbconvert --to notebook --execute notebooks/01-Combined-Extract-and-Build.ipynb
  
  - Verified existing test_raises_exception_if_dependencies_not_installed still passes
  
  Checklist
  
  - [x] Code follows project style and conventions
  - [x] Unit tests added for new behavior
  - [x] Existing tests unaffected (no breaking changes)
  - [x] Documentation updated (notebook comments explain the guard)
  - [x] No new dependencies introduced
  - [x] Backwards compatible — token-present behavior unchanged
  
  ---
  
  **2. Reply comment** (paste as a new comment on the PR):
  
  @mykola-pereyma — addressed all feedback in 4e015671:
  
  Unit test added:
  
  - test_read_returns_empty_list_when_token_missing — empty string token → returns [], logs warning
  - test_read_returns_empty_list_when_token_none — None token → returns [], logs warning
  - Same mock pattern as other provider tests (patch.dict('sys.modules', ...))
  
  PR description updated:
  
  - Added Fixes #363 (same bug, different reporter)
  - Testing section filled in with test names and manual verification steps
  - Checklist completed
  
  Let me know if you'd like any changes to the test approach.

evanerwee:
@mykola-pereyma — addressed all feedback in 4e015671:

**Unit test added:**
- `test_read_returns_empty_list_when_token_missing` — empty string token → returns `[]`, logs warning
- `test_read_returns_empty_list_when_token_none` — `None` token → returns `[]`, logs warning
- Same mock pattern as other provider tests (`patch.dict("sys.modules", ...)`)

**PR description updated** to include:
- `Fixes #363` (same bug, different reporter)
- Testing section with test names and manual verification steps
- Checklist filled in

Let me know if you would like any changes to the test approach.

mykola-pereyma (approved):
All feedback addressed — tests added, both issues referenced, PR template filled. LGTM.

oussamahansal (approved):
LGTM

noel-improv (approved):
lgtm

PR #368, opened 2026-06-30
Fix: create_config TypeError when read_timeout is passed via config

### LINKED ISSUE #361: [BUG]  Neptune create_config raises TypeError when read_timeout is passed
### Package version

3.18.5

### Package

lexical-graph

### Python version

3.12.0

### Operating System

Linux

### Description

Passing a custom read_timeout to a Neptune graph store via its config raises a TypeError, so there's no supported way to shorten (or lengthen) the botocore client's read timeout from the public API. This matters when running behind a strict request-timeout (e.g. an API Gateway / load-balancer idle ceiling), where the library's 600s default read timeout is far longer than the surrounding budget and a half-open pooled socket blocks instead of failing fast.

### Steps to reproduce

```python
from graphrag_toolkit.lexical_graph.storage import GraphStoreFactory

graph_store = GraphStoreFactory.for_graph_store(
    'neptune-db://my-cluster.cluster-xxxx.us-east-1.neptune.amazonaws.com',
    config='{"read_timeout": 5}',
)
graph_store.execute_query('MATCH (n) RETURN n LIMIT 1')  # client is built lazily here
```

Minimal equivalent (the failing call inside create_config):
```python
from botocore.config import Config
Config(read_timeout=600, **{"read_timeout": 5})
```

```shell
Traceback (most recent call last):
  File "<repro>", line 7, in <module>
    create_config(json.dumps({"read_timeout": 5}))
  File ".../storage/graph/neptune_graph_stores.py", line 149, in create_config
    return Config(
           ^^^^^^^
TypeError: botocore.config.Config() got multiple values for keyword argument 'read_timeout'
```

evanerwee:
Fixes #361.

## Problem

`create_config()` hardcodes `read_timeout=600` as a keyword argument AND spreads user config via `**config_args`. When a user passes `read_timeout` in their config JSON, Python raises:

```
TypeError: botocore.config.Config() got multiple values for keyword argument 'read_timeout'
```

Same issue affects `retries` and `user_agent_appid`.

## Fix

Use `setdefault()` for all hardcoded values — the same pattern already used for `max_pool_connections` on the line immediately above. User-provided values take precedence; defaults apply when absent.

```python
# Before (breaks):
return Config(read_timeout=600, **config_args)  # duplicate kwarg

# After (works):
config_args.setdefault('read_timeout', 600)
return Config(**config_args)                     # no conflict
```

## Impact

- 4-line change, no new logic
- All existing defaults preserved (non-breaking)
- Users can now configure `read_timeout`, `retries`, and `user_agent_appid`
- Critical for production deployments behind API Gateway / ALB / Lambda timeouts

### DIFF
--- a/lexical-graph/src/graphrag_toolkit/lexical_graph/storage/graph/neptune_graph_stores.py
+++ b/lexical-graph/src/graphrag_toolkit/lexical_graph/storage/graph/neptune_graph_stores.py
@@ -146,15 +146,11 @@ def create_config(config:Optional[str]=None):
 
     config_args.setdefault('max_pool_connections', DEFAULT_MAX_POOL_CONNECTIONS)
 
-    return Config(
-        retries={
-            'total_max_attempts': 1,
-            'mode': 'standard'
-        },
-        read_timeout=600,
-        user_agent_appid=f'graphrag-lexical-graph-{toolkit_version}',
-        **config_args
-    )
+    config_args.setdefault('read_timeout', 600)
+    config_args.setdefault('retries', {'total_max_attempts': 1, 'mode': 'standard'})
+    config_args.setdefault('user_agent_appid', f'graphrag-lexical-graph-{toolkit_version}')
+
+    return Config(**config_args)
 
 def create_property_assigment_fn_for_neptune(key:str, value:Any) -> Callable[[str], str]:
     """


mykola-pereyma:
Thanks, we will review you PR shortly.

lexical-graph/src/graphrag_toolkit/lexical_graph/storage/graph/neptune_graph_stores.py
```diff
@@ -146,15 +146,11 @@ def create_config(config:Optional[str]=None):
...
-            'mode': 'standard'
-        },
-        read_timeout=600,
-        user_agent_appid=f'graphrag-lexical-graph-{toolkit_version}',
-        **config_args
-    )
+    config_args.setdefault('read_timeout', 600)
+    config_args.setdefault('retries', {'total_max_attempts': 1, 'mode': 'standard'})
+    config_args.setdefault('user_agent_appid', f'graphrag-lexical-graph-{toolkit_version}')
+
```
mykola-pereyma, line 152: Fix looks correct and follows same pattern already used for max_pool_connections above. Could you add regression tests for the three fields that were breaking? The existing TestCreateConfig class in test_neptune_graph_stores.py already has this pattern for max_pool_connections so should be straightforward. Without regression tests there is nothing preventing someone from refactoring this back to kwargs and reintroducing the TypeError.

mykola-pereyma (changes requested):
Thanks for the fix — the setdefault approach is clean and correct. Asking for regression tests before merge since this is a bug fix and best practice is to have a test that would have caught the original issue.

marora90:
Thanks for opening this @evanerwee I’d like to help get this over the line. Since the requested changes have been waiting for a couple weeks, I’m going to open a follow-up PR that cherry-picks your commits and addresses the review feedback. Happy to close mine if you’d prefer to continue this one.

mykola-pereyma:
Superseded by #396 which includes this fix plus regression tests.

PR #369, opened 2026-06-30
fix(lexical-graph): escape OpenCypher metadata filter values and keys

noel-improv:
## Summary

Escape metadata filter values and property keys in the OpenCypher filter builder (`graph_utils.py`) instead of interpolating them raw into the filter substring, closing an OpenCypher injection where a value or key containing a single quote (or backtick) could break out and inject Cypher. The substring is spliced into queries by several callers, so it bypasses the parameter binding the graph store uses on its own `_execute_query` paths.

This follows the same escaping approach already merged for Cypher label injection (`escape_cypher_label`), reusing that helper for the property-key case.

## Changes

A new `escape_cypher_string` doubles backslashes and then backslash-escapes single quotes, so a value cannot close its literal. `formatter_for_type` routes text and timestamp values through it. `to_key` backtick-quotes the property key via `escape_cypher_label` (`source.\`{key}\``), so a key cannot break out of the identifier. Inline notes record that doubling a backslash changes match semantics for a literal-backslash value (acceptable) and that the numeric path is intentionally unquoted.

## Unit tests

New `test_graph_utils_injection.py` asserts that values are escaped inside their single-quoted literal and that keys are backtick-quoted with embedded backticks doubled, plus the benign positive paths. The existing `test_graph_utils.py` assertions are updated for the backtick-quoted keys. Bandit reports no findings.

## Release integration test

Adds `integration-tests/security/test_opencypher_filter_injection.py` and a `security-integration-opencypher-filter.yml` workflow that runs the clause emitted by the real filter builder against a live openCypher engine (Neo4j service container) on release, and on PRs touching the sink or the test. It confirms legitimate filters resolve, the injection payloads are inert, and the pre-fix interpolation still breaks out on the same engine. It skips when `NEO4J_TEST_URI` is unset.


mykola-pereyma (approved):
Correct escaping for OpenCypher string literals and identifiers. Escaping order verified (backslash first, then quote), all adversarial inputs produce valid Cypher. LGTM.

acarbonetto (commented):
Make sure to add security integration tests to release testing.

aghassel:
LGTM. Escaping plus backtick-quoting scoped to `graph_utils.py` (no changes to the 5 filter consumers) is the right call: lowest blast radius, and it fully closes the breakout. Backslash-before-quote ordering in `escape_cypher_string` is correct.

**[optional]** Doubling backslashes is correct for injection safety, but it changes stored-match semantics for a value containing a literal backslash, so a one-line confirmation that this is acceptable would help. A brief inline note that the `number/int/float` path is intentionally unquoted (safe because the type is derived from the value's Python type) would also help the next reader.

.github/workflows/security-integration-opencypher-filter.yml
```diff
@@ -0,0 +1,52 @@
+# Runs the OpenCypher filter-injection integration test for the lexical-graph
```
mykola-pereyma: Why add a specific workflow?

mykola-pereyma (changes requested):

aghassel:
Thanks for the doc notes and the integration test. LGTM.

oussamahansal (approved):
LGTM

integration-tests/test-scripts/graphrag_toolkit_tests/lexical_graph_cypher_safety.py
```diff
@@ -248,3 +258,204 @@ def test_vulnerable_query_deletes_canary_or_is_rejected(self):
...
+        return 'OpenCypher metadata filter builder escapes filter values and keys'
+
+    def _make_graph_store(self):
+        graph_store_id = os.environ['GRAPH_STORE']
+
+        if graph_store_id.startswith('neptune-graph://'):
+            graph_identifier = graph_store_id[len('neptune-graph://'):]
+            return 'analytics', NeptuneAnalyticsClient(graph_id=graph_identifier)
+
+        if graph_store_id.startswith('neptune-db://'):
```
acarbonetto, line 307: nit: duplicate code consider adding this to a helper function

acarbonetto (approved):

PR #370, opened 2026-06-30
fix(byokg-rag): escape property names in Neptune get_nodes/get_one_hop_edges

noel-improv:
## Summary

Escape property names in the byokg-rag Neptune graph store. `get_nodes` and `get_one_hop_edges` interpolate property names from `node_type_to_property_mapping` straight into the OpenCypher WHERE clause (`n.{prop}`). Node ids are bound as `$node_ids`, but the property name is raw, so a name containing a backtick (sourced from schema or caller config) can break out of the identifier and inject Cypher.

## Changes

Backtick-quote the property and escape it with the module's existing `_escape_cypher_label` helper, so the property is emitted as a backtick-quoted identifier with any embedded backtick doubled. That is the same handling node labels already get in this file, and it brings both sinks under the existing `TestNoUnescapedLabelSinks` regression guard, which only inspects backtick-quoted spans.

## Unit tests

One parametrized test asserts the property is backtick-quoted with embedded backticks doubled on both the `get_nodes` and `get_one_hop_edges` sinks, and that the raw unescaped form is absent. A benign property covers the positive path. An escape-disabled red-state test confirms the doubling in `_escape_cypher_label` is what neutralizes the injection. Bandit reports no findings.

## Integration test

Adds `BYOKGPropertyNameInjectionSafety` to `byokg_cypher_safety.py`, registered in `byokg.short` and `byokg.all` and built on the existing `IntegrationTestBase` framework. It drives the real `get_nodes` / `get_one_hop_edges` sinks against a live Neptune graph (Analytics or Database) with a backtick-bearing property name and confirms the escaped property cannot break out and delete a seeded canary node. It runs on the SageMaker notebook through the standard test suite, alongside the other BYOKG security tests, which now share `make_graph_store()` and `canary_count()` helpers in that module.


aghassel:
LGTM. Matches the module's existing `_escape_cypher_label` convention and brings both sinks under the `TestNoUnescapedLabelSinks` guard.

**[nit, optional]** The tests assert the escaped form is present but not that the raw (unescaped) form is absent, and there is no escape-disabled red-state test. Minor: an explicit "raw form is absent" assertion would make the non-vacuous intent clearer.

noel-improv:
thanks @aghassel added an explicit "raw form absent" assertion to both the `get_nodes` and `get_one_hop_edges` tests and a red-state test that disables `_escape_cypher_label` and confirms the un-doubled backtick reaches the query

integration-tests/security/test_byokg_property_injection.py
```diff
@@ -0,0 +1,112 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Security integration test: property-name injection in the byokg-rag Neptune store.
+
+Runs with its parent fix PR. Drives the real get_nodes / get_one_hop_edges sinks
```
mykola-pereyma: What PR you are referring to?

mykola-pereyma (changes requested):

aghassel:
Thanks, the raw-form-absent assertion plus the escape-disabled red-state test cover the nit. LGTM.

oussamahansal (approved):
LGTM

acarbonetto (approved):
Thanks.  Consider consolidating some of the test code if you can.

noel-improv:
Refractored `test_get_nodes_backtick_quotes_property` and `test_get_one_hop_edges_backtick_quotes_property` to collapse into one `@pytest.mark.parametrize` over the sink method. Also `make_graph_store` + `canary_count` is duplicated across 3 classes the right cleanup and would drop ~60 lines but it refactors already-merged code inside a security-fix PR expanding it's original scope
 
 @acarbonetto

integration-tests/test-scripts/graphrag_toolkit_tests/byokg_cypher_safety.py
```diff
@@ -434,3 +440,112 @@ def test_canary_survives_injection_payload(self):
...
+
+class BYOKGPropertyNameInjectionSafety(IntegrationTestBase):
+    """Drive get_nodes / get_one_hop_edges with a backtick-bearing property name
+    against a live Neptune graph and confirm the escaped property cannot break
+    out and delete a canary node."""
+
+    @property
+    def description(self):
+        return 'get_nodes and get_one_hop_edges escape interpolated property names'
+
```
mykola-pereyma, line 444: There are now three copies of _make_graph_store in this file with identical logic (including the port 8182 fix you added to each). Consider extracting to a module-level helper to avoid divergence if someone updates one but not the others.

integration-tests/test-scripts/graphrag_toolkit_tests/byokg_cypher_safety.py
```diff
@@ -434,3 +440,112 @@ def test_canary_survives_injection_payload(self):
                 self.assertEqual(self._canary_after, 1)
 
         handler.run_assertions(S3PathInjectionSafetyAssertions)
```
mykola-pereyma, line 421: PR description still mentions the standalone workflow and integration-tests/security/test_byokg_property_injection.py which no longer exist after you consolidated. Please update description to reflect the current state so future readers are not confused.

mykola-pereyma (changes requested):
Addressed feedback — integration test now lives in the existing byokg suite and runs against real Neptune. Much better than the standalone workflow. Two things to clean up before merge.

integration-tests/test-scripts/graphrag_toolkit_tests/byokg_cypher_safety.py
noel-improv: Done. Pulled `make_graph_store()` out to module level and pointed all three classes at it and did the same for `canary_count()` since it had the identical triplication

integration-tests/test-scripts/graphrag_toolkit_tests/byokg_cypher_safety.py
noel-improv: Updated the description

Issue #359, opened 2026-06-29
[FEATURE] Add support for AOSS NextGen vector search collections


acarbonetto:
### Package

lexical-graph

### Problem statement

The existing OpenSearch Serverless (AOSS) vector store integration targets Classic collections. When pointed at an **AOSS NextGen** vector search collection endpoint, the integration fails at index creation with:

```
illegal_argument_exception: "Field parameter 'engine' is not supported"
```

This is because AOSS NextGen (OpenSearch 3.x) removed the `engine` and `mode` parameters from `knn_vector` field mappings — the system now determines the optimal configuration internally. The existing code unconditionally includes `"engine": "nmslib"` or `"engine": "faiss"` in the method block, which is rejected by NextGen endpoints.

NextGen is AWS's recommended collection type for new deployments and offers significant advantages: scale-to-zero pricing, GPU-accelerated index builds, 32x compression by default, and simplified index management. Users who create new collections today will likely be using NextGen, making this a growing compatibility issue.

### Proposed solution

Add a `GraphRAGConfig` flag to opt in to NextGen-compatible behaviour. The flag approach is intentional — if NextGen supersedes Classic entirely in future, removing the flag and its conditional branches is straightforward.

### 1. Two new `GraphRAGConfig` properties

**`opensearch_serverless_nextgen: bool`**
- Default: `False`
- Env var: `OPENSEARCH_SERVERLESS_NEXTGEN` (`"true"` to enable)
- Controls whether NextGen-compatible index mappings and settings are used

**`opensearch_serverless_nextgen_compression: Optional[str]`**
- Default: `None` (uses NextGen's built-in default of `32x`)
- Env var: `OPENSEARCH_SERVERLESS_NEXTGEN_COMPRESSION`
- Valid values: `"1x"`, `"2x"`, `"8x"`, `"16x"`, `"32x"`
- When set, adds `compression_level` to the field mapping

Both follow the existing `opensearch_engine` pattern (private backing field, property with env var fallback, setter).

### 2. NextGen index mapping in `opensearch_vector_indexes.py`

When `GraphRAGConfig.opensearch_serverless_nextgen` is `True`, `index_exists()` builds a NextGen-compatible mapping:

**Classic mapping (existing, unchanged):**
```python
{
    "settings": {"index": {"knn": True, "knn.algo_param.ef_search": 100}},
    "mappings": {
        "date_detection": False,
        "properties": {
            "embedding": {
                "type": "knn_vector",
[...]

PR #391, opened 2026-07-10
[FEATURE] Add OpenSearch Serverless AOSS NextGen vector store support

noel-improv:
## Description

Adds support for AOSS NextGen vector search collections. Classic collections and NextGen collections need different `knn_vector` mappings (NextGen rejects the `engine`/`method` fields Classic requires), so `index_exists()` now auto-detects which one a collection needs: it tries the Classic mapping first, and only if the collection rejects it as NextGen-incompatible, retries once with the NextGen mapping. No new configuration is required. `GraphRAGConfig.opensearch_serverless_nextgen` still exists as an explicit `True`/`False` override for callers who want to skip detection, but it's unset by default.

## Changes

- `GraphRAGConfig.opensearch_serverless_nextgen` is now `Optional[bool]`: `None` (default) auto-detects, `True`/`False` forces one mapping.
- `index_exists()` builds the NextGen mapping (no `method` block, `space_type` at the field level, optional `compression_level`) or the unchanged Classic mapping, via `_build_index_mapping()`.
- Auto-detection retries once with NextGen on the specific rejection error. An explicit `False` override still raises an actionable `ValueError` on that same rejection instead of the previous silent fallback to a no-op client.
- New example CloudFormation template (`graphrag-toolkit-neptune-db-opensearch-nextgen-serverless.json`) provisioning a NextGen collection via an explicit `CollectionGroup` with `Generation: NEXTGEN`. `StandbyReplicas` must be `ENABLED` on both the group and the collection — found by deploying and reading the actual `CREATE_FAILED` errors. Wired into the integration-test `EnvTypeMap` as `neptune-db-aoss-nextgen`.
- `test_aoss_nextgen_cfn_template_parity.py` diffs the Classic and NextGen templates field by field, so a future fix to their shared resources can't silently miss one of them.
- `repair_opensearch_vector_store()`'s batch loop now catches the `ValueError` case per index instead of aborting the whole run on the first misconfigured one.

## Problem

Pointing `graphrag-toolkit` at an AOSS NextGen collection fails at index creation because the code unconditionally sends `engine`/`method` parameters NextGen doesn't support.

Related issue (if any): #359

## Testing

35 new unit tests covering both mapping shapes, auto-detection (retry-on-rejection, no-retry on explicit overrides), the error-raising branch's false-positive guards, CFN template parity, and the repair script's per-index handling. Full suite in the touched packages: 175 passed.

Manually verified end-to-end against real AOSS NextGen collections (created and torn down during this work): reproduced the issue's exact error, confirmed auto-detection retries and succeeds against a real NextGen collection with no configuration set, confirmed the created index's mapping is actually NextGen-shaped, and confirmed the explicit-`False` override still raises instead of retrying.

Also ran the existing `lexical.short` integration playbook (24 tests: build, extract, 10+ query variants, checkpointing, FalkorDB, batch fallback, injection safety) against a real Neptune DB + Classic AOSS stack, since auto-detection now sits on the default path for every deployment, not just NextGen ones. Result: 22 passed, 2 skipped (no pgvector store or GPU instance in this env, both expected), 0 failed — no regression on the Classic path.

## Checklist

## Still open

The integration test script (`aoss_nextgen.py`) and `lexical.nextgen` playbook from the issue are left out, pending a dedicated NextGen test endpoint for CI. The new template is already wired into the `EnvTypeMap` so that's a small follow-up.

Considered proactively detecting generation via `BatchGetCollection`/`BatchGetCollectionGroup` instead of the reactive retry. Went reactive — no new IAM permissions or CFN changes needed, since it only uses OpenSearch calls the client already has permission for.

---

### DIFF
--- a/lexical-graph/src/graphrag_toolkit/lexical_graph/config.py
+++ b/lexical-graph/src/graphrag_toolkit/lexical_graph/config.py
@@ -50,7 +50,7 @@ DEFAULT_INCLUDE_CLASSIFICATION_IN_ENTITY_ID = True
 DEFAULT_ENABLE_CACHE = False
 DEFAULT_METADATA_DATETIME_SUFFIXES = ['_date', '_datetime']
 DEFAULT_OPENSEARCH_ENGINE = 'nmslib'
-DEFAULT_OPENSEARCH_SERVERLESS_NEXTGEN = False
+DEFAULT_OPENSEARCH_SERVERLESS_NEXTGEN = None
 DEFAULT_ENABLE_VERSIONING = False
 DEFAULT_CHUNK_EXTERNAL_PROPERTIES = None
 DEFAULT_LOCAL_OUTPUT_DIR = 'output'  # Local staging directory for batch files (use /tmp for EKS)
@@ -1176,13 +1176,16 @@ class _GraphRAGConfig:
         self._opensearch_engine = opensearch_engine
 
     @property
-    def opensearch_serverless_nextgen(self) -> bool:
+    def opensearch_serverless_nextgen(self) -> Optional[bool]:
+        """None means unset: index_exists() auto-detects NextGen vs Classic by trying
+        Classic first and retrying with NextGen only if the collection rejects it. Set to
+        True/False (or the env var) to force one mapping and skip auto-detection."""
         if self._opensearch_serverless_nextgen is None:
             self._opensearch_serverless_nextgen = string_to_bool(os.environ.get('OPENSEARCH_SERVERLESS_NEXTGEN'), DEFAULT_OPENSEARCH_SERVERLESS_NEXTGEN)
         return self._opensearch_serverless_nextgen
 
     @opensearch_serverless_nextgen.setter
-    def opensearch_serverless_nextgen(self, opensearch_serverless_nextgen: bool) -> None:
+    def opensearch_serverless_nextgen(self, opensearch_serverless_nextgen: Optional[bool]) -> None:
         self._opensearch_serverless_nextgen = opensearch_serverless_nextgen
 
     @property
--- a/lexical-graph/src/graphrag_toolkit/lexical_graph/config.py
+++ b/lexical-graph/src/graphrag_toolkit/lexical_graph/config.py
@@ -14,6 +14,7 @@ from botocore.exceptions import SSOTokenLoadError
 
 import threading
 import logging
+from enum import Enum
 from dataclasses import dataclass, field
 from typing import Optional, Union, Dict, Set, List
 from boto3 import Session as Boto3Session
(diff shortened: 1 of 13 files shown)


mykola-pereyma:
Nice work on the auto-detection approach. One observation for a potential future improvement — as of boto3>=1.43.17, the `BatchGetCollectionGroup` API exposes a `generation` field with enum values `CLASSIC` | `NEXTGEN`. This enables deterministic detection without relying on error message text:

```python
aoss = boto3.client("opensearchserverless")
coll = aoss.batch_get_collection(names=[collection_name])["collectionDetails"][0]
group_name = coll.get("collectionGroupName")
if not group_name:
    is_nextgen = False  # no group = Classic
else:
    group = aoss.batch_get_collection_group(names=[group_name])["collectionGroupDetails"][0]
    is_nextgen = group.get("generation") == "NEXTGEN"
```

This would remove the error-string coupling but requires:
- Bumping minimum boto3 from >=1.40.61 to >=1.43.17
- Two new IAM permissions: `aoss:BatchGetCollection`, `aoss:BatchGetCollectionGroup`

Not a blocker for this PR — the reactive approach works today with any boto3 version and zero new permissions. Just noting the path for when the boto3 minimum is bumped next.

lexical-graph/src/graphrag_toolkit/lexical_graph/config.py
```diff
@@ -291,6 +292,8 @@ class _GraphRAGConfig:
     _enable_cache: Optional[bool] = None
     _metadata_datetime_suffixes: Optional[List[str]] = None
     _opensearch_engine: Optional[str] = None
+    _opensearch_serverless_nextgen: Optional[bool] = None
```
acarbonetto: should we instead use a enum (classic | nextgen) to coincide with the generation field from AOSS?  That would scale better if AOSS chose to include a third value in the generation field.

lexical-graph/src/graphrag_toolkit/lexical_graph/config.py
```diff
@@ -1172,6 +1175,29 @@ def opensearch_engine(self) -> str:
     def opensearch_engine(self, opensearch_engine: str) -> None:
         self._opensearch_engine = opensearch_engine
 
+    @property
+    def opensearch_serverless_nextgen(self) -> Optional[bool]:
+        """None means unset: index_exists() auto-detects NextGen vs Classic by trying
+        Classic first and retrying with NextGen only if the collection rejects it. Set to
+        True/False (or the env var) to force one mapping and skip auto-detection."""
```
acarbonetto: I don't quite understand this works... 
unset: auto-detects NextGen vs Classic by trying Classic first and retrying with NextGen
True: force NextGen?  
False: force Classic?  

It's not obvious.  Maybe switching to an optional enum to force one type would be more obvious?

lexical-graph/tests/unit/storage/vector/test_opensearch_clients.py
```diff
@@ -10,62 +10,12 @@
...
-    sys.modules["opensearchpy"] = fake_opensearch_mod
-    sys.modules["opensearchpy.exceptions"] = fake_exceptions_mod
-    sys.modules["llama_index.vector_stores"] = fake_llama_vs_mod
-    sys.modules["llama_index.vector_stores.opensearch"] = fake_llama_os_mod
-
-    return _NotFoundError
-
-
-_NotFoundError = _install_opensearch_mocks()
+install_opensearch_mocks()
```
acarbonetto, line 18: very nice ❤️

acarbonetto (commented):
Looks good.  I would consider an upgrade to boto3 to receive the endpoint generation and this would simplify the code and avoid the retry mechanism. 

Otherwise, we should consider changing the bool env variable to an enum value.  This would make the interface more intuitive (not sure the force mechanism is going to see much use outside of test).

mykola-pereyma (approved):
LGTM

acarbonetto (approved):
thanks!

PR #397, opened 2026-07-14
[FEATURE] Add OpenSearch (opensearch://) vector store support alongside AOSS

noel-improv:
## Description

Adds an `opensearch://` connection string for an OpenSearch endpoint that isn't Amazon OpenSearch Serverless (AOSS) — a local Docker or Finch container, an EC2 instance, or any other OpenSearch deployment — alongside the existing `aoss://` one. Today the vector store factory only recognizes AOSS endpoints, and the client creation code hardcodes AWS SigV4 signing, so there's no way to point the toolkit at a plain OpenSearch instance without AWS credentials.

## Changes

- `OpenSearchVectorIndexFactory` recognizes an `opensearch://<endpoint>` prefix and routes it through a new `is_local` flag on `OpenSearchIndex`, instead of AOSS detection.
- `create_os_client`/`create_os_async_client` skip AWS SigV4 on this path. They authenticate with HTTP basic auth from `GraphRAGConfig.opensearch_username`/`opensearch_password` (backed by `OPENSEARCH_USERNAME`/`OPENSEARCH_PASSWORD`), or with no auth if neither is set, and log a warning if only one of the two is set.
- `index_exists()` now accepts `client_kwargs`, threaded from `OpenSearchIndex.client_kwargs`, so a plain-HTTP or self-signed endpoint can override `use_ssl`/`verify_certs` (this path previously always forced `use_ssl=True`, which the AOSS-only code never needed to override).
- `is_local` also selects `is_aoss=False` on the underlying vector client, so bulk ingest calls `refresh()` after writes, matching OpenSearch's own behavior instead of AOSS's.
- Reuses the existing Classic knn_vector mapping (`engine: faiss`/`nmslib`) unchanged — no new mapping logic needed.
- README and docs-site updated with the new connection string and its auth behavior.

## Problem

`VectorStoreFactory.for_vector_store()` has no way to target an OpenSearch endpoint that isn't Amazon OpenSearch Serverless; every code path assumes AWS SigV4 and an AOSS endpoint.

Related issue (if any): #

noel-improv/graphrag-toolkit#18 stacks on top of this branch (deterministic AOSS Classic/NextGen generation detection). It's approved and will be rebased onto `main` and opened here once this merges.

## Testing

Unit tests cover the factory's `opensearch://` detection, the basic-auth/no-auth/partial-credential client paths, and `client_kwargs` threading through `index_exists()`. Also ran a live smoke test against two Finch-hosted OpenSearch containers: an unsecured plain-HTTP instance (index creation, bulk ingest, query, confirming `refresh()` fires) and a basic-auth instance with a real admin password (confirming both successful auth and rejection of a wrong password).

## Checklist

---

### DIFF
--- a/docs-site/src/content/docs/lexical-graph/vector-store-opensearch-serverless.mdx
+++ b/docs-site/src/content/docs/lexical-graph/vector-store-opensearch-serverless.mdx
@@ -134,12 +134,12 @@ Field descriptions:
 
 ### Connecting to an OpenSearch vector store
 
-You can use [OpenSearch](https://opensearch.org) directly as a vector store — in a Docker or Finch container, on an EC2 instance, or on your own infrastructure. graphrag-toolkit connects to it with the [opensearch-py](https://pypi.org/project/opensearch-py/) client. Supply a connection string that begins with `opensearch://`, followed by the endpoint:
+You can use [OpenSearch](https://opensearch.org) directly as a vector store — in a Docker or Finch container, on an EC2 instance, or on your own infrastructure. graphrag-toolkit connects to it with the [opensearch-py](https://pypi.org/project/opensearch-py/) client. Pass the endpoint as an `http://` or `https://` URL:
 
 ```python
 from graphrag_toolkit.lexical_graph.storage import VectorStoreFactory
 
-opensearch_connection_info = 'opensearch://localhost:9200'
+opensearch_connection_info = 'https://localhost:9200'
 
 with VectorStoreFactory.for_vector_store(
   opensearch_connection_info,
@@ -148,7 +148,7 @@ with VectorStoreFactory.for_vector_store(
     ...
 ```
 
-The endpoint doesn't need a scheme: `opensearch://localhost:9200` connects over SSL, the toolkit's default (matching OpenSearch, which ships with SSL enabled). Use an explicit `http://` scheme for a plain-HTTP instance: `opensearch://http://localhost:9200`. An endpoint without a port defaults to 9200; for an endpoint on port 443, include the port or an `https://` scheme.
+Any `https://`/`http://` endpoint that isn't an Amazon OpenSearch Serverless domain is treated as a self-managed OpenSearch cluster (use the `aoss://` prefix for Serverless). Use `https://` to connect over SSL, the toolkit's default (matching OpenSearch, which ships with SSL enabled), or `http://` for a plain-HTTP instance. An endpoint without a port defaults to 9200.
 
 A default local OpenSearch (Docker or Finch) uses a self-signed certificate, so the example above fails certificate verification until you either trust its CA or disable verification with `client_kwargs={'verify_certs': False}` (see the self-signed example below). Certificate verification stays on by default.
 
@@ -158,7 +158,7 @@ Any `client_kwargs` you pass to `VectorStoreFactory.for_vector_store` are forwar
 
 ```python
 with VectorStoreFactory.for_vector_store(
-  'opensearch://https://localhost:9200',
+  'https://localhost:9200',
   index_names=['chunk'],
   client_kwargs={'verify_certs': False}
 ) as vector_store:
--- a/docs-site/src/content/docs/lexical-graph/vector-store-opensearch-serverless.mdx
+++ b/docs-site/src/content/docs/lexical-graph/vector-store-opensearch-serverless.mdx
@@ -149,3 +149,16 @@ with VectorStoreFactory.for_vector_store(
 ```
 
 Unlike `aoss://` (Amazon OpenSearch Serverless), an `opensearch://` endpoint never uses AWS SigV4. It connects with HTTP basic auth from the `OPENSEARCH_USERNAME` and `OPENSEARCH_PASSWORD` environment variables (or `GraphRAGConfig.opensearch_username` / `GraphRAGConfig.opensearch_password`), or with no auth if neither is set. If you omit the scheme from the endpoint (e.g. `opensearch://localhost:9200`), it defaults to `https://`.
+
+The `opensearch://` prefix targets OpenSearch instances that don't use AWS authentication (basic auth or no auth). A managed Amazon OpenSearch Service domain that authenticates with IAM is not supported through this prefix yet.
+
+For a local Docker or Finch container over plain HTTP, use `opensearch://http://localhost:9200`. For an endpoint with a self-signed TLS certificate, disable certificate verification by passing `client_kwargs`:
+
+```python
+with VectorStoreFactory.for_vector_store(
+  'opensearch://https://localhost:9200',
+  index_names=['chunk'],
+  client_kwargs={'verify_certs': False}
+) as vector_store:
+    ...
+```
--- a/docs-site/src/content/docs/lexical-graph/vector-store-opensearch-serverless.mdx
+++ b/docs-site/src/content/docs/lexical-graph/vector-store-opensearch-serverless.mdx
@@ -134,7 +134,7 @@ Field descriptions:
 
 ### Connecting to an OpenSearch vector store
 
-You can also use OpenSearch directly as a vector store, instead of Amazon OpenSearch Serverless — in a Docker or Finch container, on an EC2 instance, or on your own infrastructure. graphrag-toolkit connects to it with the [opensearch-py](https://pypi.org/project/opensearch-py/) client. Supply a connection string that begins `opensearch://`, followed by the endpoint:
+You can use [OpenSearch](https://opensearch.org) directly as a vector store — in a Docker or Finch container, on an EC2 instance, or on your own infrastructure. graphrag-toolkit connects to it with the [opensearch-py](https://pypi.org/project/opensearch-py/) client. Supply a connection string that begins with `opensearch://`, followed by the endpoint:
 
 ```python
 from graphrag_toolkit.lexical_graph.storage import VectorStoreFactory
@@ -148,11 +148,11 @@ with VectorStoreFactory.for_vector_store(
     ...
 ```
 
-Unlike `aoss://` (Amazon OpenSearch Serverless), an `opensearch://` endpoint never uses AWS SigV4. It connects with HTTP basic auth from the `OPENSEARCH_USERNAME` and `OPENSEARCH_PASSWORD` environment variables (or `GraphRAGConfig.opensearch_username` / `GraphRAGConfig.opensearch_password`), or with no auth if neither is set. OpenSearch runs with SSL enabled by default, so if you omit the scheme (e.g. `opensearch://localhost:9200`) the endpoint defaults to `https://`.
+The endpoint doesn't need a scheme: `opensearch://localhost:9200` connects over SSL, the toolkit's default (matching OpenSearch, which ships with SSL enabled). Use an explicit `http://` scheme for a plain-HTTP instance: `opensearch://http://localhost:9200`. An endpoint without a port defaults to 9200; for an endpoint on port 443, include the port or an `https://` scheme.
 
-The `opensearch://` prefix targets OpenSearch instances that don't use AWS authentication (basic auth or no auth). A managed Amazon OpenSearch Service domain that authenticates with IAM is not supported through this prefix yet ([tracked in #407](https://github.com/awslabs/graphrag-toolkit/issues/407)).
+The client authenticates with HTTP basic auth from the `OPENSEARCH_USERNAME` and `OPENSEARCH_PASSWORD` environment variables (or `GraphRAGConfig.opensearch_username` / `GraphRAGConfig.opensearch_password`), or with no auth if neither is set.
 
-For a local Docker or Finch container over plain HTTP, use `opensearch://http://localhost:9200`. For an endpoint with a self-signed TLS certificate, disable certificate verification by passing `client_kwargs`:
+Any `client_kwargs` you pass to `VectorStoreFactory.for_vector_store` are forwarded to the opensearch-py client, so all of its [connection options](https://opensearch-project.github.io/opensearch-py/api-ref/clients/opensearch_client.html) — certificate verification, CA bundles, timeouts — are available. For example, for an endpoint with a self-signed TLS certificate:
 
 ```python
 with VectorStoreFactory.for_vector_store(
@@ -162,3 +162,7 @@ with VectorStoreFactory.for_vector_store(
 ) as vector_store:
     ...
 ```
+
+#### Limitations
+
+Authentication methods other than basic auth, including SigV4 for IAM-authenticated domains, are not supported yet ([#407](https://github.com/awslabs/graphrag-toolkit/issues/407)).
--- a/docs-site/src/content/docs/lexical-graph/vector-store-opensearch-serverless.mdx
+++ b/docs-site/src/content/docs/lexical-graph/vector-store-opensearch-serverless.mdx
@@ -9,6 +9,7 @@ title: OpenSearch Serverless Vector Store
   - [Creating an OpenSearch Serverless vector store](#creating-an-opensearch-serverless-vector-store)
   - [Amazon OpenSearch Serverless and custom document IDs](#amazon-opensearch-serverless-and-custom-document-ids)
     - [Verify and repair an Amazon OpenSearch Serverless vector store](#verify-and-repair-an-amazon-opensearch-serverless-vector-store)
+  - [Self-managed OpenSearch](#self-managed-opensearch)
 
 ### Overview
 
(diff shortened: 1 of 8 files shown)


mykola-pereyma (commented):
Few suggestions:

1. Could we rename `is_local` to `is_serverless` (inverted, default True)? The flag means "use AOSS SigV4 auth" vs "use basic auth or no auth" — not whether the instance is local. An EC2-hosted OpenSearch cluster is not local but would use this path. `is_serverless` matches AWS terminology and reads naturally at call sites.

2. The docs page and README should mention how to connect over plain HTTP or with self-signed certs — local Docker users will hit SSL errors with the https default. Something like: "For plain HTTP, use `opensearch://http://localhost:9200`. For self-signed TLS, pass `client_kwargs={"verify_certs": False}` to the factory."

3. Worth adding a note that `opensearch://` is for instances without AWS authentication (basic auth or no auth). A user with a managed OpenSearch Service domain (non-serverless but still on AWS) might try this prefix and get auth failures — a sentence clarifying that managed OpenSearch Service with IAM auth is not yet supported would save confusion.

docs-site/src/content/docs/lexical-graph/vector-store-opensearch-serverless.mdx
```diff
@@ -9,6 +9,7 @@ title: OpenSearch Serverless Vector Store
   - [Creating an OpenSearch Serverless vector store](#creating-an-opensearch-serverless-vector-store)
```
acarbonetto:
```suggestion
  - [Creating an Amazon OpenSearch Serverless vector store](#creating-an-opensearch-serverless-vector-store)
```

docs-site/src/content/docs/lexical-graph/vector-store-opensearch-serverless.mdx
```diff
@@ -9,6 +9,7 @@ title: OpenSearch Serverless Vector Store
...
   - [Amazon OpenSearch Serverless and custom document IDs](#amazon-opensearch-serverless-and-custom-document-ids)
     - [Verify and repair an Amazon OpenSearch Serverless vector store](#verify-and-repair-an-amazon-opensearch-serverless-vector-store)
+  - [OpenSearch](#opensearch)
```
acarbonetto:
```suggestion
  - [Connecting to an OpenSearch vector store](#opensearch)
```

docs-site/src/content/docs/lexical-graph/vector-store-opensearch-serverless.mdx
```diff
@@ -130,3 +131,21 @@ Field descriptions:
 | `num_docs` | Number of documents in a specific tenant vector index |
 | `num_deleted` | Number of documents deleted from a specific tenant vector index (indicative number only if `dry_run` is `true`) |
 | `num_unindexed` | Number of nodes that have not been indexed in a specific tenant vector index |
+
+### OpenSearch
```
acarbonetto: update anchor above too
```suggestion
### Connecting to an OpenSearch vector store 
```

docs-site/src/content/docs/lexical-graph/vector-store-opensearch-serverless.mdx
```diff
@@ -130,3 +131,21 @@ Field descriptions:
...
+
+You can also use OpenSearch directly as a vector store, instead of Amazon OpenSearch Serverless — running locally in a Docker or Finch container, on an EC2 instance, or on your own infrastructure. Supply a connection string that begins `opensearch://`, followed by the endpoint:
```
acarbonetto: link to opensearch.org

lexical-graph/src/graphrag_toolkit/lexical_graph/storage/vector/opensearch_vector_indexes.py
```diff
@@ -115,20 +115,35 @@ class DummyAuth:
...
+    signed with credentials from a pre-configured session and region. When is_local is
+    True, this instead connects to an OpenSearch endpoint that isn't Amazon OpenSearch
+    Serverless (AOSS): HTTP basic auth from
+    GraphRAGConfig.opensearch_username/opensearch_password if both are set, otherwise
+    no auth. The AWS session and region are never touched on this path.
 
     Args:
         endpoint: str
             The OpenSearch endpoint URL to connect to.
+        is_local: bool
```
acarbonetto:
```suggestion
        is_sigv4_auth: bool
```

lexical-graph/src/graphrag_toolkit/lexical_graph/storage/vector/opensearch_vector_index_factory.py
```diff
@@ -10,6 +10,7 @@
 
 OPENSEARCH_SERVERLESS = 'aoss://'
 OPENSEARCH_SERVERLESS_DNS = 'aoss.amazonaws.com'
+OPENSEARCH_LOCAL = 'opensearch://'
```
acarbonetto: OPENSEARCH doesn't necessary use localhost (if you're running it locally, it will)
We should move away from "local == open source", since it's not necessarily a local instance.

docs-site/src/content/docs/lexical-graph/vector-store-opensearch-serverless.mdx
acarbonetto: link to https://pypi.org/project/opensearch-py/ since we're using the opensearch-py client to connect.

docs-site/src/content/docs/lexical-graph/vector-store-opensearch-serverless.mdx
```diff
@@ -130,3 +131,21 @@ Field descriptions:
...
+
+```python
+from graphrag_toolkit.lexical_graph.storage import VectorStoreFactory
+
+opensearch_connection_info = 'opensearch://http://localhost:9200'
```
acarbonetto: by default, OpenSearch runs with SSL turned on (https://localhost:9200).

lexical-graph/src/graphrag_toolkit/lexical_graph/storage/vector/opensearch_vector_index_factory.py
```diff
@@ -43,19 +44,25 @@ def try_create(self, index_names:List[str], vector_index_info:str, **kwargs) ->
...
         endpoint = None
+        is_local = False
         if vector_index_info.startswith(OPENSEARCH_SERVERLESS):
             endpoint = vector_index_info[len(OPENSEARCH_SERVERLESS):]
             if not endpoint.startswith('https://') and not endpoint.startswith('http://'):
                 endpoint = f'https://{endpoint}'
+        elif vector_index_info.startswith(OPENSEARCH_LOCAL):
+            endpoint = vector_index_info[len(OPENSEARCH_LOCAL):]
+            if not endpoint.startswith('https://') and not endpoint.startswith('http://'):
+                endpoint = f'https://{endpoint}'
```
acarbonetto: according to https://opensearch-project.github.io/opensearch-py/api-ref/clients/opensearch_client.html endpoint doesn't _need_ https:// or http://.  We could leave it off and reduce the chance of mistakes.

lexical-graph/src/graphrag_toolkit/lexical_graph/storage/vector/opensearch_vector_indexes.py
```diff
@@ -138,12 +153,15 @@ def create_os_client(endpoint, **kwargs):
             A configured client instance for interacting with OpenSearch.
 
     """
-    session = GraphRAGConfig.session
-    region = GraphRAGConfig.aws_region
-    credentials = session.get_credentials()
-    service = 'aoss'
+    if is_local:
+        auth = _local_http_auth()
```
acarbonetto: (username, password) is one way to connect, but I think we need to consider https:// connections too.  IF we can only support http for now, let's mark that as a limitation and we can raise a ticket later to support https auth.

acarbonetto (commented):
looks good - I agree that we need to move away from Open Source == Local.  OpenSearch can be deployed in more ways than just a localhost.  IF we only want to support localhost (for now) and raise a follow-up ticket to support more options later, that's fine.  But the language in the code should be updated now.

mykola-pereyma (approved):
LGTM

docs-site/src/content/docs/lexical-graph/vector-store-opensearch-serverless.mdx
```diff
@@ -130,3 +131,34 @@ Field descriptions:
 | `num_docs` | Number of documents in a specific tenant vector index |
 | `num_deleted` | Number of documents deleted from a specific tenant vector index (indicative number only if `dry_run` is `true`) |
 | `num_unindexed` | Number of nodes that have not been indexed in a specific tenant vector index |
+
+### Connecting to an OpenSearch vector store
+
+You can also use OpenSearch directly as a vector store, instead of Amazon OpenSearch Serverless — in a Docker or Finch container, on an EC2 instance, or on your own infrastructure. graphrag-toolkit connects to it with the [opensearch-py](https://pypi.org/project/opensearch-py/) client. Supply a connection string that begins `opensearch://`, followed by the endpoint:
```
acarbonetto:
```suggestion
You can use [OpenSearch](opensearch.org) directly as a vector store — in a Docker container, on an EC2 instance, or on your own infrastructure. graphrag-toolkit connects to it with the [opensearch-py](https://pypi.org/project/opensearch-py/) client. Supply a connection string that begins with `opensearch://`, followed by the endpoint:
```

acarbonetto: Question: why do we require the "opensearch://" string?  Doesn't this get removed before being sent to opensearch-py?

docs-site/src/content/docs/lexical-graph/vector-store-opensearch-serverless.mdx
```diff
@@ -130,3 +131,34 @@ Field descriptions:
...
+  index_names=['chunk']
+) as vector_store:
+    ...
+```
+
+Unlike `aoss://` (Amazon OpenSearch Serverless), an `opensearch://` endpoint never uses AWS SigV4. It connects with HTTP basic auth from the `OPENSEARCH_USERNAME` and `OPENSEARCH_PASSWORD` environment variables (or `GraphRAGConfig.opensearch_username` / `GraphRAGConfig.opensearch_password`), or with no auth if neither is set. OpenSearch runs with SSL enabled by default, so if you omit the scheme (e.g. `opensearch://localhost:9200`) the endpoint defaults to `https://`.
+
+The `opensearch://` prefix targets OpenSearch instances that don't use AWS authentication (basic auth or no auth). A managed Amazon OpenSearch Service domain that authenticates with IAM is not supported through this prefix yet ([tracked in #407](https://github.com/awslabs/graphrag-toolkit/issues/407)).
+
+For a local Docker or Finch container over plain HTTP, use `opensearch://http://localhost:9200`. For an endpoint with a self-signed TLS certificate, disable certificate verification by passing `client_kwargs`:
```
acarbonetto: A lot of the above is supplied by the opensearch-py documentation, since we are mostly just forwarding certs and auth to opensearch-py there's no need to duplicate documentation.  

I would suggest removing a lot of the above documentation, and remove any reference to Amazon (opensearch is not defined by not being an Amazon product).  We can include a "limitations" section to discuss how we don't yet support https:// endpoints or any auth other than basic auth (please reference https://github.com/awslabs/graphrag-toolkit/issues/407 or other new issues to fix these limitations).

acarbonetto: Actually, we do support https:// with certs, by passing client_kwargs.  Good.  We can document this is it differs from the opensearch-py documentation.  However, if its the same documentation & examples, just reference opensearch-py documentation directly.

lexical-graph/src/graphrag_toolkit/lexical_graph/storage/vector/opensearch_vector_index_factory.py
```diff
@@ -43,19 +44,25 @@ def try_create(self, index_names:List[str], vector_index_info:str, **kwargs) ->
                 names and endpoint configuration, or None if no suitable endpoint is found.
         """
         endpoint = None
+        is_sigv4_auth = True
         if vector_index_info.startswith(OPENSEARCH_SERVERLESS):
             endpoint = vector_index_info[len(OPENSEARCH_SERVERLESS):]
             if not endpoint.startswith('https://') and not endpoint.startswith('http://'):
                 endpoint = f'https://{endpoint}'
+        elif vector_index_info.startswith(OPENSEARCH):
```
acarbonetto: consider remove this constraint on the endpoint.  If it starts with aoss:// - it's an amazon opensearch serverless endpoint.  Otherwise, it's for opensearch endpoint.

lexical-graph/src/graphrag_toolkit/lexical_graph/storage/vector/opensearch_vector_indexes.py
```diff
@@ -139,12 +154,15 @@ def create_os_client(endpoint, **kwargs):
             A configured client instance for interacting with OpenSearch.
 
     """
-    session = GraphRAGConfig.session
-    region = GraphRAGConfig.aws_region
-    credentials = session.get_credentials()
-    service = 'aoss'
+    if not is_sigv4_auth:
```
acarbonetto, line 159: add a comment here to https://github.com/awslabs/graphrag-toolkit/issues/407 since this will need to be cleaned up for more use cases.  We should support more auth options than just sigv4 and basic auth.

lexical-graph/src/graphrag_toolkit/lexical_graph/storage/vector/opensearch_vector_indexes.py
```diff
@@ -162,28 +180,36 @@ def create_os_client(endpoint, **kwargs):
...
             ``pool_maxsize`` defaults to :data:`DEFAULT_POOL_MAXSIZE` when unset.
 
     Returns:
         AsyncOpenSearch: An instantiated asynchronous OpenSearch client.
     """
-    session = GraphRAGConfig.session
-    region = GraphRAGConfig.aws_region
-    credentials = session.get_credentials()
-    service = 'aoss'
+    if not is_sigv4_auth:
```
acarbonetto, line 213: add a comment here to #407

lexical-graph/src/graphrag_toolkit/lexical_graph/storage/vector/opensearch_vector_indexes.py
```diff
@@ -660,20 +699,21 @@ def client(self) -> OpensearchVectorClient:
                 self._client = None
 
         if not self._client:
-            if index_exists(self.endpoint, self.underlying_index_name(), self.dimensions, self.writeable):
+            if index_exists(self.endpoint, self.underlying_index_name(), self.dimensions, self.writeable, is_sigv4_auth=self.is_sigv4_auth, client_kwargs=self.client_kwargs):
```
acarbonetto: this line is more readable if you save the result to a variable.
```
   does_index_exist = index_exists(self.endpoint, self.underlying_index_name(), self.dimensions, self.writeable, is_sigv4_auth=self.is_sigv4_auth, client_kwargs=self.client_kwargs)
  if does_index_exist: 
    self._client = create_opensearch_vector_client(
    ...
```

docs-site/src/content/docs/lexical-graph/vector-store-opensearch-serverless.mdx
noel-improv: You are right, the prefix is stripped before `opensearch-py` sees anything; fixing

lexical-graph/src/graphrag_toolkit/lexical_graph/storage/vector/opensearch_vector_index_factory.py
noel-improv: That branch predates this PR and keeps working for anyone passing a bare AOSS domain without the `aoss://` prefix, so I've left the routing as-is here to avoid a behaviour change. I can simplify it to `aoss:// = AOSS /` `everything-else = OpenSearch` in a separate cleanup if you'd like

lexical-graph/src/graphrag_toolkit/lexical_graph/storage/vector/opensearch_vector_indexes.py
noel-improv: Done, # TODO is on the sync builder and the async one below.

lexical-graph/src/graphrag_toolkit/lexical_graph/storage/vector/opensearch_vector_index_factory.py
acarbonetto: > I can simplify it to aoss:// = AOSS / everything-else = OpenSearch in a separate cleanup if you'd like

Yes.  This. No change for the current aoss endpoints.

oussamahansal (approved):

acarbonetto (changes requested):
Want to verify this manually

Issue #299, opened 2026-06-01
Dependency conflict: aiobotocore pins botocore to narrow ranges incompatible with bedrock-agentcore


ravikamz:
**Body:**
  ```markdown
  ## Description

  `graphrag-lexical-graph` depends on `llama-index-embeddings-bedrock` → `aioboto3` → `aiobotocore` which pins `botocore` to narrow micro-version
  ranges (e.g., `>=1.38.23,<1.38.28`). This conflicts with applications using `bedrock-agentcore>=1.9.1` which requires `botocore>=1.43.0`.

  This makes it impossible to install `graphrag-lexical-graph` in the same Python environment as `bedrock-agentcore` using standard dependency
  resolution.

  ## Dependency Chain

  graphrag-lexical-graph==3.18.3
    └── llama-index-embeddings-bedrock==0.8.0
         └── aioboto3
              └── aiobotocore==2.25.1
                   └── botocore>=1.38.23,<1.38.28  ← CONFLICTS

  bedrock-agentcore==1.9.1
    └── botocore>=1.43.0  ← REQUIRES NEWER

  ## Error Message (uv resolver)

  graphrag-lexical-graph==3.18.3 depends on one of:
      botocore>=1.38.23,<1.38.28
      botocore>=1.39.9,<1.39.12
      botocore>=1.40.15,<1.40.19
      botocore>=1.40.37,<1.40.62

  And because bedrock-agentcore==1.9.1 depends on botocore>=1.43.0,
  graphrag-lexical-graph==3.18.3 and bedrock-agentcore are incompatible.

  ## Use Case

  Integrating graphrag-toolkit retrieval into an application that uses Bedrock AgentCore Runtime (e.g., a serve layer that hosts both the AgentCore app
   and the graphrag retriever in the same process).

  ## Workaround

  Install with lowest resolution:

  ```bash
  pip install --resolution lowest-direct graphrag-lexical-graph==3.18.3

  Or with uv:

  uv pip install graphrag-lexical-graph==3.18.3 --resolution lowest-direct

  This picks compatible versions of aiobotocore that work with both packages.

  Suggested Fix

  Consider making llama-index-embeddings-bedrock an optional dependency (extras) since not all users need it — users who already have their own
  embedding implementation (direct Bedrock calls) don't need the llama-index embedding wrapper and its transitive dependency chain.

  Environment

  - graphrag-lexical-graph: 3.18.3
  - bedrock-agentcore: 1.9.1
[...]

PR #400, opened 2026-07-15
[FIX] Integration test fail on boto3/botocore dependencies

acarbonetto:
## Description

[FIX] Integration tests are brittle and will fail if there is a version dependency mismatch between boto3 and botocore.  Unfortunately, we are bringing in a lot of dependencies for integration tests that potentially cause this mismatch. 

## Changes

- Use a requirement.txt file to install test dependencies. 
- Add unit tests to check version mismatches to find errors early. 

## Problem

Related issue (if any): N/A

## Testing

- [X] Unit tests added/updated
- [X] Integration tests added (as appropriate)
- [X] Existing tests pass (`pytest`)
- [X] Tested manually (describe below)

## Checklist

- [X] Code follows existing style and conventions
- [X] License headers present on new files
- [X] Documentation updated (if applicable)
- [X] No breaking changes (or clearly documented)

---

### DIFF
--- a/.github/workflows/byokg-rag-tests.yml
+++ b/.github/workflows/byokg-rag-tests.yml
@@ -74,7 +74,7 @@ jobs:
         uses: astral-sh/setup-uv@11f9893b081a58869d3b5fccaea48c9e9e46f990 # v8.3.2
 
       - name: Create virtual environment
-        run: uv venv .venv
+        run: uv venv --seed .venv
 
       - name: Install dependencies
         run: |
--- a/integration-tests/requirements-integ-test.txt
+++ b/integration-tests/requirements-integ-test.txt
@@ -8,3 +8,4 @@ psycopg2-binary>=2.9.0
 pgvector>=0.2.0
 sentence-transformers>=2.2.0
 torch>=2.0.0
+aiobotocore>=2.5.0
--- a/integration-tests/requirements-integ-test.txt
+++ b/integration-tests/requirements-integ-test.txt
@@ -1,0 +1,10 @@
+neo4j>=5.0.0
+opensearch-py>=2.4.0
+llama-index-vector-stores-opensearch>=0.1.0
+llama-index-readers-web>=0.1.0
+llama-index-readers-file>=0.1.0
+llama-index-readers-s3>=0.1.0
+psycopg2-binary>=2.9.0
+pgvector>=0.2.0
+sentence-transformers>=2.2.0
+torch>=2.0.0
--- a/integration-tests/requirements-integ-test.txt
+++ b/integration-tests/requirements-integ-test.txt
@@ -1,9 +1,9 @@
 neo4j>=5.0.0
-opensearch-py>=2.4.0
-llama-index-vector-stores-opensearch>=0.1.0
-llama-index-readers-web>=0.1.0
-llama-index-readers-file>=0.1.0
-llama-index-readers-s3>=0.1.0
+opensearch-py>=3.0.0
+llama-index-vector-stores-opensearch>=1.0.0
+llama-index-readers-web>=0.3.0
+llama-index-readers-file>=0.4.0
+llama-index-readers-s3>=0.4.0
 psycopg2-binary>=2.9.0
 pgvector>=0.2.0
 sentence-transformers>=2.2.0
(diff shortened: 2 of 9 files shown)


.github/workflows/byokg-rag-tests.yml
```diff
@@ -74,7 +74,7 @@ jobs:
         uses: astral-sh/setup-uv@11f9893b081a58869d3b5fccaea48c9e9e46f990 # v8.3.2
 
       - name: Create virtual environment
-        run: uv venv .venv
+        run: uv venv --seed .venv
```
acarbonetto, line 77: this seeds the venv with pip and setuptools at creation time which makes pip available for tests

lexical-graph/tests/unit/test_integ_dependency_compatibility.py
```diff
@@ -0,0 +1,117 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
mykola-pereyma, line 1: The test file appears identical in both byokg-rag and lexical-graph (same blob SHA). Was this intentional? If one gets updated later, how do we keep them in sync?

mykola-pereyma (commented):
The single-pass install approach makes sense — resolving all deps together prevents the mismatch. Few questions before approving:

lexical-graph/tests/unit/test_integ_dependency_compatibility.py
```diff
@@ -0,0 +1,117 @@
...
+        f.write("\n".join(packages))
+        tmp_req = f.name
+
+    tmp_report = tmp_req + ".json"
+
+    try:
+        result = subprocess.run(
+            [
+                sys.executable, "-m", "pip", "install",
+                "--dry-run", "--quiet", "--ignore-installed",
```
mykola-pereyma, line 72: This calls pip install --dry-run which hits PyPI for metadata. How long does this take in CI? Is there a risk of flaky failures if PyPI is slow or unreachable during the test run?

integration-tests/requirements-integ-test.txt
```diff
@@ -0,0 +1,11 @@
...
+opensearch-py>=3.0.0
+llama-index-vector-stores-opensearch>=1.0.0
+llama-index-readers-web>=0.3.0
+llama-index-readers-file>=0.4.0
+llama-index-readers-s3>=0.4.0
+psycopg2-binary>=2.9.0
+pgvector>=0.2.0
+sentence-transformers>=2.2.0
+torch>=2.0.0
+aiobotocore>=2.5.0
```
mykola-pereyma, line 11: Is aiobotocore directly needed by the integration tests, or is it pulled in transitively by something else (s3fs)? Asking because aiobotocore constrains botocore to a narrow range (e.g. <1.40.62) which is the same pattern that caused issue #299. The single-pass resolver handles it, but if we do not actually need aiobotocore explicitly here, removing it would eliminate the constraint.

integration-tests/requirements-integ-test.txt
```diff
@@ -0,0 +1,11 @@
+neo4j>=5.0.0
```
mykola-pereyma, line 1: The requirements use open lower bounds (>=X.Y.Z) with no upper bounds. Is the version-alignment test sufficient to catch breakage if a new release of one dep becomes incompatible with another? Or should we consider pinning at least the known-problematic packages (aiobotocore, botocore, boto3)?

acarbonetto: The lower bounds is sufficient and requires less maintenance.  We can pin exact versions, but that would require an update each time a new pacakge is released. 

If a package update causes a dependency mismatch, we have unit tests to cover that.

integration-tests/requirements-integ-test.txt
acarbonetto: We need it for s3fs.  Otherwise the dependencies cause a mismatch.

lexical-graph/tests/unit/test_integ_dependency_compatibility.py
acarbonetto: We don't have cross-package tests (I considered creating one for integration tests).  I don't anticipate a lot of churn in these tests, so this seemed like the best way forward for the moment.

lexical-graph/tests/unit/test_integ_dependency_compatibility.py
acarbonetto: I hadn't considered that.  It's kind of required for the test to be successful.  
As PyPI is an external dependency, this shouldn't cause a PR to fail.  Let's cross that bridge if this becomes a consistent problem.

mykola-pereyma (approved):

Issue #401, opened 2026-07-15
Feature: S3-backed Checkpoint for serverless/containerized environments


evanerwee:
### Package

lexical-graph

### Problem statement

The existing `Checkpoint` class (`graphrag_toolkit.lexical_graph.indexing.build.checkpoint`) uses local filesystem operations (`os.path.exists`, `os.makedirs`, `open`) to store checkpoint markers. This means checkpointing **does not work** in serverless or containerized environments where local disk is ephemeral:

- **ECS Fargate** — container restarts lose `/tmp` contents
- **AWS Lambda** — ephemeral `/tmp` cleared between invocations  
- **EKS pods** — ephemeral storage unless PVC is configured
- **SQS-driven workers** — message retry may land on a different container

The toolkit already supports S3 for document storage (`S3BasedDocs`), vector storage (`S3VectorIndexes`), and batch inference I/O (`BatchConfig`). The checkpoint is the **only** component that cannot run without persistent local disk.

### Proposed solution

An `S3Checkpoint` class that provides the **same interface** as `Checkpoint` (`add_filter`, `add_writer`, `enabled`) but uses zero-byte S3 objects as markers instead of local files.

```python
from graphrag_toolkit.lexical_graph.indexing.build import S3Checkpoint

checkpoint = S3Checkpoint(
    checkpoint_name="enrichment-run-001",
    bucket_name="my-pipeline-bucket",
    key_prefix="checkpoints/tenant-a"
)

# Same API as Checkpoint — drop-in replacement
graph_index.extract(docs, handler=extracted_docs, checkpoint=checkpoint)
graph_index.build(docs, checkpoint=checkpoint)
```

### Mapping (local → S3)

| Local `Checkpoint` | `S3Checkpoint` equivalent | 
|---|---|
| `os.path.exists(path)` | `s3.head_object()` — catch 404 |
| `open(path, "a"); os.utime(path)` | `s3.put_object(Body=b"")` |
| `os.makedirs(dir)` | No-op (S3 has no directories) |

### Features

| Feature | Description |
|---|---|
| **Drop-in interface** | Same `add_filter` / `add_writer` / `enabled` as `Checkpoint` |
| **Zero-byte markers** | Minimal cost (HEAD + PUT, no data transfer) |
| **Survives restarts** | Markers persist in S3 across container restarts |
| **Safe defaults** | On S3 errors, includes node (re-processes rather than skips) |
[...]

Issue #325, opened 2026-06-16
[FEATURE] Document-level extraction manifest for distributed resume and incremental ingestion


mykola-pereyma:
> **⚠️ DRAFT — This issue is under discussion and not yet ready for implementation.**

## Problem Statement

The existing `Checkpoint`/`CheckpointFilter` tracks per-node completion via local filesystem savepoints:
```python
# CheckpointFilter.checkpoint_does_not_exist():
node_checkpoint_path = join(self.checkpoint_dir, tenant_node_id)
if os.path.exists(node_checkpoint_path): ...
```

Limitations at 1M+ document scale:

1. **Local filesystem only** — savepoints are files at `{output_dir}/save_points/{checkpoint_name}/{node_id}`. Doesn't work across distributed workers/containers.
2. **Node-level, not document-level** — tracks individual chunks. A document with 5 chunks = 5 savepoint files. Can't atomically answer "is this document fully extracted?"
3. **No failure state** — missing savepoint = "not done." No distinction between "never attempted" and "attempted, failed, needs retry."
4. **O(N) scan at resume** — at 1M docs × 5 chunks = 5M savepoint files. `os.path.exists()` per node at startup is slow.

### Why This Is Needed

When extraction or build fails mid-run (OOM, throttle, timeout, spot termination), the pipeline must re-process everything from scratch. At 1M+ documents that means re-calling Bedrock for documents already successfully extracted — burning time and money. The state store enables **resumability**: on restart, only incomplete documents are processed.

### What Already Exists

- `Checkpoint` + `CheckpointFilter` — per-node local savepoints ✅
- `S3BasedDocs` — S3 I/O for extraction output ✅
- `ExtractionPipeline.extraction_filters` — metadata-based filtering ✅
- `GraphStoreFactory.for_graph_store(uri)` — URI-based factory pattern ✅

## Proposed Solution: `DocumentStateStore`

A backend-agnostic abstraction that tracks per-document completion across both extraction and build phases:

```python
class DocumentStateStore(ABC):
    """Tracks per-document completion status across extraction/build phases."""

    @abstractmethod
    async def get_status(self, doc_id: str) -> Optional[DocStatus]:
        ...

    @abstractmethod
    async def set_status(self, doc_id: str, status: DocStatus, *, expected: Optional[DocStatus] = None) -> bool:
        """Conditional write. Returns False if expected != current (optimistic lock)."""
        ...

    @abstractmethod
[...]

PR #402, opened 2026-07-15
feat: add S3Checkpoint for serverless/containerized environments

evanerwee:
## Description

Add S3-backed checkpoint (`S3Checkpoint`) that stores markers as zero-byte S3 objects instead of local files. Enables checkpointing in ECS Fargate, Lambda, and EKS environments where local disk is ephemeral.

## Changes

- New `S3Checkpoint` class — drop-in replacement for `Checkpoint` with same interface
- `S3CheckpointFilter` — uses `HEAD` request instead of `os.path.exists()`  
- `S3CheckpointWriter` — uses `PUT` zero-byte object instead of `open(path, "a")`
- Export added to `__init__.py`
- 15 unit tests

## Problem

Related issue: #401

The existing `Checkpoint` class uses local filesystem markers (`os.path.exists`, `os.makedirs`, `open`). This fails in serverless/containerized environments (Fargate, Lambda, EKS) where local disk is ephemeral and does not survive container restarts or SQS message retries.

## Testing

Manual testing: verified syntax passes `ast.parse()`, imports resolve correctly against the existing module structure.

## Checklist

---

### DIFF
--- a/lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/build/__init__.py
+++ b/lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/build/__init__.py
@@ -6,5 +6,6 @@ from .vector_indexing import VectorIndexing
 from .graph_construction import GraphConstruction
 from .version_manager import VersionManager
 from .checkpoint import Checkpoint
+from .s3_checkpoint import S3Checkpoint
 from .build_filters import BuildFilters, DEFAULT_BUILD_FILTER
 from .delete_sources import DeletePrevVersions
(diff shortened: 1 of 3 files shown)


mykola-pereyma (commented):
Thank you for putting this together — the problem is real and we understand the need for checkpointing in Fargate/Lambda environments.

We have #325 on our implementation roadmap which covers this use case as part of a broader checkpoint/state-store redesign. That design introduces a URI-based factory pattern (matching how GraphStoreFactory and VectorStoreFactory already work) with support for multiple backends (S3, DynamoDB) and document-level progress tracking for distributed workers.

A few things we noticed in this PR that the broader design would address differently:

- The  parameter is accepted but never used (the S3 client comes from GraphRAGConfig)
- No shared interface between Checkpoint and S3Checkpoint — if we add DynamoDB later it would be a third copy
- Per-node HEAD requests in the filter path would be a bottleneck at scale (list_objects_v2 with prefix can batch-check)

We appreciate the contribution and the detailed issue write-up. Would you be open to collaborating on #325 instead? Your experience running this on Fargate would be valuable input for the design — especially around the failure modes and retry patterns you encountered.

Issue #403, opened 2026-07-15
[FEATURE] Add toolkit-native reranker chains with configurable fallback behavior


marora90:
### Package

lexical-graph

### Problem statement

I’d like to make the reranking behavior a little easier to use in production.

Right now traversal search lets you pick one reranker, like tfidf, bedrock etc. 

That works, but in a real deployment it would be preferable to try a better reranker first, like Bedrock, and fall back to the built-in TF-IDF reranker if Bedrock is throttled, unavailable, misconfigured, or returns no usable scores.

### Proposed solution

Smallest possible implementation would be to keep all of the new behavior inside RerankStatements.

New shape can be something like:
```
ProcessorArgs(
    reranker="bedrock,tfidf",
    reranker_fallback_policy=my_policy, #optional
)
```

RerankStatements would add a private parser/helper that turns the configured value into an ordered list

Instead of the existing dispatch block, it would loop over the reranker chain
https://github.com/awslabs/graphrag-toolkit/blob/64422d797c776fdb4f68c1025622d66ed8b226f9/lexical-graph/src/graphrag_toolkit/lexical_graph/retrieval/processors/rerank_statements.py#L250-L258

something like:
```
for reranker in chain:
    try:
        scored_values = scorer(...)
        if scored_values:
            return scored_values

        if not self._should_fallback(reranker, scored_values=scored_values):
            return scored_values

    except Exception as e:
        if not has_next_reranker or not self._should_fallback(reranker, error=e):
            raise
        logger.warning("Reranker failed; trying fallback", exc_info=True)
```

where user code can decide whether to fall back on throttling, timeouts, empty scores, etc.

The current defaults will be maintained including support for a single reranker along with a basic default policy that mimics the behavior today.

### Alternatives considered

A more involved implementation would be to introduce a small toolkit-owned reranker abstraction inside retrieval, then adapt statement and topic reranking to use it.

Rough shape:
```
retrieval/reranking/
  base.py        # shared dataclasses/protocols
  chain.py       # ordered fallback runner
  factory.py     # turns "bedrock,tfidf" into reranker objects
  providers/
    tfidf.py
    bedrock.py
    sentence_transformer.py
    noop.py
```

The useful contract could be like:
`score(query, texts) -> scores`

The existing rerankers can then become providers:
```
TfidfReranker
BedrockReranker
SentenceTransformerReranker
NoopReranker
```

[...]

PR #406, opened 2026-07-16
[FEATURE] Add reranker fallback chains to statement reranking

marora90:
## Description

This PR adds ordered fallback chains for final statement reranking while keeping entity reranking unchanged.

Users can continue to configure a single reranker:

```python
reranker='tfidf'
```

or provide an ordered chain:
```
reranker=['bedrock', 'tfidf']
``` 

If an earlier statement reranker raises and the fallback policy permits recovery, the next strategy is attempted.

## Changes

- Accept an ordered list of final statement rerankers.
- Add reranker_fallback_policy, called as:
```
  policy(reranker, *, error=None) -> bool
```
- Support none as either:
    - the only strategy, disabling statement reranking; or
    - the final fallback, returning the original unreranked statements once reached.

- Validate list-based reranker chains.
- Improve fallback logging:
    - successful reranking logs at DEBUG;
    - fallback warnings name the failed and next strategies;
    - warnings retain the exception type and traceback;
    - empty score maps log at ERROR.

- Clarify the statement/entity reranking distinction in the public documentation.
- Make the throttling-policy example robust to exceptions whose response attribute is not a dictionary.

## Problem

Model-based rerankers can fail because of transient provider errors such as throttling. 

Previously, callers had to either accept that failure or implement fallback outside the retrieval pipeline.

This change allows callers to prefer a model-based reranker while retaining a local fallback.

Fallback remains stateless and is evaluated per query. A reranker that failed for one query will be attempted again for the next query.

Related issue: #403

## Decisions to review

A few behaviors are intentional but worth confirming during review:

### 1. Reranker chains only apply to statements

`reranker` and `reranker_fallback_policy` control final statement reranking. Entity reranking remains independent and
always uses TF-IDF, even when statement reranking is disabled with `None` or `none`.

### 2. Fallback allows any exception by default

Without a custom policy, any failure from a non-final reranker advances to the next strategy. This includes transient
errors such as throttling, but also configuration, credential, dependency, or implementation errors.

A custom policy can restrict which failures allow fallback. Exceptions from the final reranker—and exceptions raised by
the policy itself—always propagate.

### 3. Falling back to `none` still requires policy approval

For `['bedrock', 'none']`, the Bedrock failure must be approved by the fallback policy before `none` is reached. If
denied, the original exception propagates.

Once reached, `none` performs no scoring and cannot fail.

### 4. Empty score maps do not trigger fallback

An empty score map is treated as a completed scorer response. It is logged at `ERROR`, no fallback is attempted, and all
statements are dropped.

This preserves existing behavior, but is worth confirming because it may be surprising in a chain.

### 5. Existing validation behavior is preserved

`ProcessorArgs` remains permissive, with validation performed by `RerankStatements`. Lists validate reranker names and
the placement of `none`; legacy single-string values retain their existing, less-strict behavior for compatibility.

### 6. Fallback state is not remembered

Fallback is evaluated independently for every query. A failed provider is retried on the next query rather than being
temporarily disabled, so persistent failures may add latency until the underlying issue is resolved.

## Testing

## Checklist

---

### DIFF
--- a/docs-site/src/content/docs/lexical-graph/traversal-based-search-configuration.mdx
+++ b/docs-site/src/content/docs/lexical-graph/traversal-based-search-configuration.mdx
@@ -174,20 +174,55 @@ ___
 
 ### Reranking strategy
 
-Traversal-based search incorporates reranking at two key points during the retrieval process:
+Traversal-based search incorporates two independent reranking stages:
 
-  - When generating entity network contexts, both entities and entity networks are reranked
-  - Before finalizing search results, the complete set of statements undergoes reranking
+  - When generating entity network contexts, entities are always reranked with TF-IDF
+  - Before finalizing search results, the complete set of statements can be reranked with a configured strategy
 
-Reranking is managed through a single parameter:
+The `reranker` and `reranker_fallback_policy` parameters control only the final statement reranking stage. They do not change entity reranking.
 
 #####  `reranker`
 
 Parameters options:
 
   - `model`: Uses a LlamaIndex-based `SentenceReranker` to rerank all statements in the result set
-  - `tfidf` (default): Applies a term frequency-inverse document frequency measure to rank statements 
-  - `None`: Disables the reranking feature completely
+  - `bedrock`: Uses the configured Amazon Bedrock reranking model to rerank all statements in the result set
+  - `tfidf` (default): Applies a term frequency-inverse document frequency measure to rank statements
+  - `None` or `none`: Disables final statement reranking; entity reranking still uses TF-IDF
+
+You can also pass an ordered list of rerankers to try them in sequence. This is useful when you want to prefer a model-based reranker, but fall back to the built-in `tfidf` reranker if the first reranker raises an exception.
+
+```python
+query_engine = LexicalGraphQueryEngine.for_traversal_based_search(
+    graph_store,
+    vector_store,
+    reranker=['bedrock', 'tfidf']
+)
+```
+
+Reranker names in a chain are validated before reranking runs; unknown names raise a `ValueError` on the first query. By default, chained rerankers fall back to the next strategy on any exception. To restrict this to specific failures, provide a `reranker_fallback_policy` function; the reranker falls back only when the function returns `True`. An exception from the final reranker in a chain always propagates to the caller, just as with a single reranker.
+
+Use `none` by itself to disable statement reranking, or as the final item in a chain to return unreranked statements when an earlier reranker fails. The preceding failure must be allowed by `reranker_fallback_policy` before the chain can reach `none`; otherwise, the original exception propagates. Once reached, `none` performs no scoring and cannot raise.
+
+```python
+def fallback_on_throttling(reranker, *, error=None):
+    response = getattr(error, 'response', None)
+    if not isinstance(response, dict):
+        return False
+    code = response.get('Error', {}).get('Code')
+    return code in {'ThrottlingException', 'TooManyRequestsException'}
+
+query_engine = LexicalGraphQueryEngine.for_traversal_based_search(
+    graph_store,
+    vector_store,
+    reranker=['bedrock', 'tfidf'],
+    reranker_fallback_policy=fallback_on_throttling
+)
+```
+
+Fallback is decided per query, so a reranker that fails on one query is attempted again on the next; transient conditions such as throttling recover automatically once they clear server-side.
+
+A reranker that returns an empty score map has completed without raising, so the result is terminal: the empty map is logged as an error, no fallback is attempted, and all statements are dropped.
 
 The tfidf-based option is significantly faster than the model-based approach. To use the model reranker, you must first install the following additional dependencies:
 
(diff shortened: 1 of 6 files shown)


docs-site/src/content/docs/lexical-graph/traversal-based-search-configuration.mdx
```diff
@@ -174,20 +174,55 @@ ___
...
 #####  `reranker`
 
 Parameters options:
 
   - `model`: Uses a LlamaIndex-based `SentenceReranker` to rerank all statements in the result set
-  - `tfidf` (default): Applies a term frequency-inverse document frequency measure to rank statements 
-  - `None`: Disables the reranking feature completely
+  - `bedrock`: Uses the configured Amazon Bedrock reranking model to rerank all statements in the result set
+  - `tfidf` (default): Applies a term frequency-inverse document frequency measure to rank statements
+  - `None` or `none`: Disables final statement reranking; entity reranking still uses TF-IDF
```
acarbonetto, line 191: I think it would be best if we can take advantage of the `None` type to identify the end of the list. Then we don't need to add a robust String checked (none or None or NONE or Nil, etc...)

lexical-graph/src/graphrag_toolkit/lexical_graph/retrieval/processors/rerank_statements.py
```diff
@@ -205,13 +219,80 @@ def rerank_text(text_query, text_sources, num_results, model_package_arn):
...
+        reranker_registry = {
+            'model': self._score_values,
+            'tfidf': self._score_values_with_tfidf,
+            'bedrock': self._score_values_with_bedrock
+        }
+
+        for i, reranker in enumerate(reranker_chain):
+            has_next = i < len(reranker_chain) - 1
+
+            if reranker == 'none':
```
acarbonetto, line 258: better to use None here and check false-y values

lexical-graph/src/graphrag_toolkit/lexical_graph/retrieval/processors/rerank_statements.py
```diff
@@ -224,9 +305,14 @@ def _process_results(self, search_results:SearchResultCollection, query:QueryBun
             topic are reranked based on the relevance scores computed by the specified reranking method.
 
         Raises:
-            Any exception raised during the reranking process will propagate to the caller.
+            Exceptions raised by the final (or only) reranker propagate to the
+            caller. Earlier rerankers fall back when the configured policy
+            permits it (by default, on any exception). Reaching a final
+            ``'none'`` also requires the policy to permit the preceding failure;
+            ``'none'`` itself performs no scoring and cannot raise.
```
acarbonetto, line 312: also accepts 'None'

lexical-graph/src/graphrag_toolkit/lexical_graph/retrieval/processors/rerank_statements.py
```diff
@@ -224,9 +305,14 @@ def _process_results(self, search_results:SearchResultCollection, query:QueryBun
...
         """
-        if not self.args.reranker or self.args.reranker.lower() == 'none':
+        reranker_chain = self.reranker_chain
+        if not reranker_chain or reranker_chain[0] == 'none':
```
acarbonetto, line 315: also accepts 'None'

lexical-graph/src/graphrag_toolkit/lexical_graph/retrieval/processors/rerank_statements.py
```diff
@@ -268,21 +351,7 @@ def _process_results(self, search_results:SearchResultCollection, query:QueryBun
...
-
-            Attributes:
-                No attributes are defined directly for this class.
-
-            Methods:
-                _process_results: Processes the search results to rerank the statements based
-                    on their scores.
-
-            """
+            """Apply scores to one topic, dropping statements without scores."""
```
acarbonetto, line 354: please revert and update

lexical-graph/src/graphrag_toolkit/lexical_graph/retrieval/processors/rerank_statements.py
```diff
@@ -295,24 +364,8 @@ def rerank_statements(topic:Topic, source_str:str):
...
-            Methods:
-                _process_results: Processes a collection of search results and applies the reranking
-                to individual entries based on the provided query.
-
-            Args:
-                index (int): Position index of the search result to rerank within the collection.
-                search_result (SearchResult): Individual search result object containing the
-                    data to be reranked.
-            """
+            """Apply statement reranking to one search result."""
```
acarbonetto, line 367: please revert and update

acarbonetto (commented):
Some minor comments. 
Can we split this into the two PRs: 
1. re-ranker fallback
2. adding the bedrock reranking strategy
That being said, we should really consider adding a custom reranker model.  The current strategy doesn't scale well.  Instead, we can add an interface and allow users to deploy their own business logic without needing contribute directly.  See: #435

PR #416, opened 2026-07-20
[benchmark]: Add multi-retriever benchmarking, Nova 2 Lite support, and PGA bio/stat split

oussamahansal:
## Description

Adds multi-retriever benchmark infrastructure, fixes Nova 2 Lite extraction compatibility,
and introduces PGA bio/stat split evaluation.

## Changes

- Fix null handling in batch extractors for Nova 2 Lite (returns None for some docs)
- Add topic-beam-chunk_only and topic_beam_search retrievers to benchmark harness
- Fix topic-beam-chunk_only to use SemanticGuidedChunkRetriever (was silently failing)
- Add 3K token budget (per_topic_cap) and increase max_search_results from 5 to 10
- Add PGA bio/stat split evaluation support (BENCHMARK_DATASET env var)
- Add multi-retriever loop (run_all_retrievers.sh) for single-command full benchmark
- Fix boto3/botocore version pins to use ranges for notebook compatibility

## Problem

Related issue (if any): #

## Testing

## Checklist

---


integration-tests/test-scripts/on-start.sh
```diff
@@ -27,7 +27,7 @@ python -m spacy download en_core_web_sm
 # Pin boto3/botocore/aiobotocore to compatible versions. aiobotocore constrains
 # botocore to a narrow range; without pinning, other installs can pull a boto3
 # that imports symbols missing from the allowed botocore (e.g. DocumentModifiedShape).
-pip install boto3==1.43.0 botocore==1.43.0 aiobotocore==3.7.0
+pip install "boto3>=1.40.61,<1.44" "botocore>=1.40.61,<1.44" "aiobotocore>=2.21,<2.27"
```
mykola-pereyma: The range `>=1.40.61,<1.44` spans ~200 boto3 releases. If a future 1.43.x introduces a breaking change or new dependency conflict, benchmarks will break silently. Would a tighter pin like `>=1.43.0,<1.44` (or even `~=1.43.0`) give enough flexibility for notebook compatibility without the wide lower bound?

benchmarks/utils/retriever_factory.py
```diff
@@ -18,12 +18,16 @@
 from graphrag_toolkit.lexical_graph.retrieval.retrievers import (
     ChunkBasedSearch,
     ChunkBasedSemanticSearch,
+    ChunkCosineSimilaritySearch,
     EntityBasedSearch,
     EntityNetworkSearch,
     RerankingBeamGraphSearch,
+    SemanticChunkBeamGraphSearch,
```
mykola-pereyma, line 25: This change (`max_search_results: 5 -> 10`) is in the shared `_SUB_RETRIEVER_PROCESSOR_ARGS` dict, so it affects ALL retrievers — not just the new ones. Is this intentional? If so, existing benchmark baselines will shift and need re-evaluation. If this is only needed for the new beam-search retrievers, could it be passed as a retriever-specific override instead?

mykola-pereyma (commented):
A few questions:

1. **Three concerns bundled in one PR** — The Nova 2 Lite null fix is a library-level change, the benchmark infra is test tooling, and the boto3 range is dependency management. Would it be possible to split into separate PRs? Makes review/revert cleaner if something regresses.

2. See inline comment on `on-start.sh` about the boto3 version range width.

3. See inline comment on `retriever_factory.py` about the `max_search_results` scope change.

4. `batch_topic_extractor_sync.py` is missing a newline at end of file — minor but may trigger linting.

benchmarks/utils/retriever_factory.py
oussamahansal: Yes this is intentional, and it was done for the sonnet 4.6 and nova 2 benchmarks

benchmarks/scripts/benchmark_query.py
```diff
@@ -39,8 +41,10 @@
 
 def load_qa_pairs(data_dir: str, dataset: str, qa_files: List[str], limit: Optional[int] = None):
     pairs = []
+    # pga_bio and pga_stat share the 'pga' data directory
+    data_subdir = 'pga' if dataset.startswith('pga') else dataset
```
noel-improv, line 45: **[Sugg]:** This remaps `pga_bio`/`pga_stat` → `source-data/pga/` for the loader, but `sync_benchmark_data_from_s3(dataset, ...)` at line 94 still gets the raw `pga_bio` and syncs into `source-data/pga_bio/`. might we want to apply the same `'pga' if dataset.startswith('pga')` mapping to the sync call?

benchmarks/scripts/run_all_retrievers.sh
```diff
@@ -0,0 +1,104 @@
...
+    QUERY_CLASS="benchmark_query.PgaBenchmarkQuery"
+    EVAL_CLASS="benchmark_evaluate.PgaBenchmarkEvaluate"
+    ;;
+  wikihow)
+    EXTRACT_CLASS="benchmark_extract.WikihowBenchmarkExtract"
+    BUILD_CLASS="benchmark_build.WikihowBenchmarkBuild"
+    QUERY_CLASS="benchmark_query.WikihowBenchmarkQuery"
+    EVAL_CLASS="benchmark_evaluate.WikihowBenchmarkEvaluate"
+    ;;
+  *)
```
noel-improv, line 54: **[Question/comment]:** `BENCHMARK_DATASET=pga_bio` is the only way to select the split, but it falls through here to `exit 1` since the case only accepts `cuad/concurrentqa/pga/wikihow`. So the split can't run through the all-retrievers loop. Intended combination, or out of scope for now?

noel-improv (commented):
Nice addition, the multi-retriever loop and the topic-beam engines are a good expansion of the benchmark surface, just a few inline notes below.

oussamahansal:
closing this PR: I will separate it into 3 PRs

Issue #418, opened 2026-07-23
[BUG] S3BasedDocs(for_jsonl=True) deadlocks extraction: forked worker inherits held BufferedWriter lock


noel-improv:
### Package version

3.18.5

### Package

lexical-graph

### Python version

3.10.20

### Operating System

SageMaker Notebook

### Description

Running extraction with `S3BasedDocs(for_jsonl=True)` as the handler hangs partway through the write phase and never recovers. There is no exception and no traceback; the process sits idle until it is killed.

A pipeline worker, created by `fork()`, blocks on its first log call inside `logging`'s `StreamHandler.flush()`. It has already acquired the handler lock, so the block is on the underlying log-file `BufferedWriter` lock, which the worker inherited in a locked state across the fork. The parent then blocks in `ProcessPoolExecutor.shutdown()` joining a worker that can never finish.

`logging` reinitialises its own module and handler locks across fork via `os.register_at_fork`, but it cannot reinitialise the lock inside the stream. A parent thread that was mid-write to the log at fork time therefore leaves that lock held with no owner, and the child blocks on it forever.

The `for_jsonl=True` path triggers this because its uploader keeps a long-lived thread that logs once per document, so a thread is almost always inside a logging call when the pool forks its workers. The default `for_jsonl=False` path keeps no such long-lived logging thread and rarely forks while the lock is held.

Expected: the run completes and reads the documents back. Actual: it deadlocks during the write phase, deterministically at 1,000 uploaded objects (the uploader's queue size), and has to be killed.

### Steps to reproduce

```python
from graphrag_toolkit.lexical_graph.indexing.load import S3BasedDocs

# Handler that writes one JSONL object per source document.
extracted_docs = S3BasedDocs(
    region=REGION,
    bucket_name=BUCKET,
    key_prefix=KEY_PREFIX,
    collection_id=None,
    for_jsonl=True,          # the affected path
)

# Run over a corpus large enough to keep the uploader's producer thread busy;
# 5,000 documents (~16,500 chunks) reproduced it reliably.
graph_index.extract(docs, handler=extracted_docs, show_progress=True)

# The run stalls during the write phase with no error and must be killed.
# Confirm the wedged worker with:
#   pip install py-spy && py-spy dump --pid <worker-pid>
```

### Error output / stack trace

```shell
# Forked worker: blocked on its first log call (BufferedWriter flush)
Thread (idle): "MainThread"
[...]

PR #419, opened 2026-07-23
[FIX] Use spawn start method for pipeline ProcessPoolExecutor

noel-improv:
## Description

`run_pipeline` forks its `ProcessPoolExecutor` workers using the platform default start method (`fork` on Linux). A forked worker inherits any lock a non-forking parent thread held, in a locked state. When a per-document logging thread holds the log file's `BufferedWriter` lock at fork time, the worker deadlocks on its first log call and the extraction hangs with no error. This switches the pool to the `spawn` start method, which starts workers from a clean interpreter and removes the inherited-lock hazard.

## Changes

- `pipeline_utils.py`: create the `ProcessPoolExecutor` with `mp_context=multiprocessing.get_context('spawn')` instead of the default fork context, and add the `import multiprocessing`.
- Add a comment explaining why fork is unsafe here.

## Problem

Extraction using `S3BasedDocs(for_jsonl=True)` hangs partway through the write phase with no exception and no traceback. `py-spy` shows a forked worker blocked in `logging`'s `StreamHandler.flush()` at `self.stream.flush()`. It already acquired the handler lock, so the block is the underlying log-file `BufferedWriter` lock. The parent then blocks in `ProcessPoolExecutor.shutdown()` joining a worker that can never finish.

The child inherits the buffer lock in a held state because a parent thread was mid-write to the log at fork time. `logging` reinitialises its own module and handler locks across fork via `os.register_at_fork` but cannot reach the lock inside the stream, so the child blocks on its first `flush()`.

The `for_jsonl=True` path triggers this because its uploader keeps a long-lived thread that logs once per document, so a thread is almost always inside a logging call when the pool forks. The default `for_jsonl=False` path keeps no such thread and rarely forks while the lock is held.

Related issue (if any): #418

## Testing

Reproduced the hang twice on a SageMaker notebook (Python 3.10.20) running a 5,000-document extraction with `S3BasedDocs(for_jsonl=True)`. Both runs froze at 1,000 uploaded objects with the forked worker at 0% CPU, and the `py-spy` dumps identify the `BufferedWriter` flush. With this change on the same notebook, the run passed 1,000 objects and completed all 5,000 documents. This is one clean run, and the failure is a fork-timing race, so a single pass does not rule out recurrence; it does remove the mechanism the dumps show.

## Checklist

No new files. `spawn` re-imports the module tree per worker, a small per-worker startup cost, and requires the `transform` closure to be picklable, which it is in the current pipeline.

---

### DIFF
--- a/lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/utils/pipeline_utils.py
+++ b/lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/utils/pipeline_utils.py
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 from pipe import Pipe
+import multiprocessing
 from concurrent.futures import ProcessPoolExecutor
 from functools import partial
 from typing import List, Optional, Sequence, Any, cast, Callable, Generator, Union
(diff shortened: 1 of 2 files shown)


mykola-pereyma (approved):
Clean fix — spawn eliminates the entire class of fork-inherited lock deadlocks. One minor suggestion: consider adding a note in the `run_pipeline` docstring that the `transform` callable must be picklable, since that's now an API contract (spawn serializes args to workers via pickle).

acarbonetto (approved):

PR #420, opened 2026-07-23
[FIX] Parallelise per-document chunk listing in S3ChunkDownloader

noel-improv:
## Description

`S3ChunkDownloader.download` reads extracted documents back from S3. It lists each document's chunks with a separate `list_objects_v2` call, and those calls ran one at a time on the main thread. This runs them on a thread pool instead, so the listing no longer serialises. Documents still yield in the same order.

## Changes

- List each document's chunk keys concurrently, sized by `extraction_num_threads_per_worker` (the setting the chunk downloads already use). Documents yield in prefix order. Listing and downloading use separate pools so listings prefetch while the current document downloads.
- Bound the listing with a sliding window of at most `extraction_num_threads_per_worker` documents in flight, refilled one per document consumed, so listing prefetches ahead without reading the whole collection.
- Download each document's chunks lazily, when the consumer reaches that document, rather than dispatching them at listing time. At most one document's chunk data is resident at a time (the same bound as the original serial code), and abandoning the generator early dispatches no downloads for unconsumed documents.
- Read `chunk_page.get('Contents', [])` instead of `chunk_page['Contents']`, so a prefix that lists empty returns no keys instead of raising `KeyError`.
- Tests: document and chunk order preservation, a `Barrier`-based test that the listing runs concurrently, a bounded-window test driven from a consumer thread, and a guard that no later document's chunks download while the consumer is stalled on the first document.

## Problem

Reading extracted documents back from S3 was slow. On a 5,000-document collection (16,536 chunk objects) the readback took 578s against 3s from local disk. The per-document listing accounted for 255.7s of that, running as 5,000 sequential round-trips on the main thread while the downloads already used a thread pool.

Related issue (if any): #

## Testing

169 tests in `tests/unit/indexing/load/` pass, including the four added here. The concurrency test fails with `BrokenBarrierError` against serial listing, and the lazy-dispatch guard fails against eager download dispatch (later documents' chunks download while the consumer is stalled on document 0), so they guard the behaviour rather than passing vacuously.

Measured against the same collection, listing isolated from downloads:

| threads | listing time |
|---|---|
| serial | 255.7s |
| 16 | 16.5s |
| 32 | 8.3s |
| 64 | 4.4s |

Listing is I/O-bound, so raising `extraction_num_threads_per_worker` above its default of 4 increases the gain; at the default it is smaller. This covers the listing part of readback. The chunk downloads (~322s) are unchanged.

## Checklist

Public behaviour is unchanged: the same `SourceDocument` objects yield in the same order. No new files.

---

### DIFF
--- a/lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/load/s3_based_docs.py
+++ b/lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/load/s3_based_docs.py
@@ -10,9 +10,10 @@ import threading
 import uuid
 import concurrent.futures
 
+from collections import deque
 from os.path import join
 from datetime import datetime
-from itertools import repeat
+from itertools import repeat, islice
 from threading import Semaphore
 from typing import List, Any, Generator, Optional, Dict, Callable
 
@@ -295,32 +296,44 @@ class S3ChunkDownloader(BaseComponent):
 
         num_threads = GraphRAGConfig.extraction_num_threads_per_worker
 
-        def _list_chunk_keys(source_doc_prefix):
-            chunk_pages = s3_client.get_paginator('list_objects_v2').paginate(
-                Bucket=self.bucket_name, Prefix=source_doc_prefix
+        # download_executor is the outer context so it outlives list_executor,
+        # whose tasks submit downloads onto it.
+        with concurrent.futures.ThreadPoolExecutor(max_workers=num_threads) as download_executor, \
+             concurrent.futures.ThreadPoolExecutor(max_workers=num_threads) as list_executor:
+
+            def _list_and_dispatch(source_doc_prefix):
+                # List a document's chunks (reusing the stateless paginator) and
+                # dispatch their downloads, so consecutive documents overlap.
+                chunk_pages = paginator.paginate(Bucket=self.bucket_name, Prefix=source_doc_prefix)
+                chunk_keys = [
+                    chunk_obj['Key']
+                    for chunk_page in chunk_pages
+                    for chunk_obj in chunk_page.get('Contents', [])
+                ]
+                return [
+                    download_executor.submit(self._download_chunk, chunk_key, s3_client)
+                    for chunk_key in chunk_keys
+                ]
+
+            # Bounded sliding window: at most num_threads listings in flight, so
+            # peak memory tracks the window, not the whole collection.
+            remaining_prefixes = iter(source_doc_prefixes)
+            in_flight = deque(
+                (source_doc_prefix, list_executor.submit(_list_and_dispatch, source_doc_prefix))
+                for source_doc_prefix in islice(remaining_prefixes, num_threads)
             )
-            return [
-                chunk_obj['Key']
-                for chunk_page in chunk_pages
-                for chunk_obj in chunk_page.get('Contents', [])
-            ]
 
-        # List each document's chunks concurrently. The per-document
-        # list_objects_v2 calls previously ran one at a time on the main thread
-        # and were a large part of readback. executor.map keeps yield order.
-        with concurrent.futures.ThreadPoolExecutor(max_workers=num_threads) as list_executor, \
-             concurrent.futures.ThreadPoolExecutor(max_workers=num_threads) as download_executor:
+            while in_flight:
+                source_doc_prefix, listing = in_flight.popleft()
+                download_futures = listing.result()
 
-            for source_doc_prefix, chunk_keys in zip(
-                source_doc_prefixes,
-                list_executor.map(_list_chunk_keys, source_doc_prefixes)
-            ):
+                next_prefix = next(remaining_prefixes, None)
+                if next_prefix is not None:
+                    in_flight.append(
+                        (next_prefix, list_executor.submit(_list_and_dispatch, next_prefix))
+                    )
 
-                nodes = list(download_executor.map(
-                    self._download_chunk,
-                    chunk_keys,
-                    repeat(s3_client)
-                ))
+                nodes = [download_future.result() for download_future in download_futures]
 
                 logger.debug(f'Yielding source document [source: {source_doc_prefix}, num_nodes: {len(nodes)}]')
 
--- a/lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/load/s3_based_docs.py
+++ b/lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/load/s3_based_docs.py
@@ -293,26 +293,37 @@ class S3ChunkDownloader(BaseComponent):
 
         logger.debug(f'Started getting source documents from S3 [bucket: {self.bucket_name}, collection_path: {collection_path}, num_prefixes: {len(source_doc_prefixes)}]')
 
-        with concurrent.futures.ThreadPoolExecutor(max_workers=GraphRAGConfig.extraction_num_threads_per_worker) as executor:
+        num_threads = GraphRAGConfig.extraction_num_threads_per_worker
 
-            for source_doc_prefix in source_doc_prefixes:
-                
-                chunk_pages = paginator.paginate(Bucket=self.bucket_name, Prefix=source_doc_prefix)
-                
-                chunk_keys = [
-                    chunk_obj['Key']
-                    for chunk_page in chunk_pages
-                    for chunk_obj in chunk_page['Contents'] 
-                ]
+        def _list_chunk_keys(source_doc_prefix):
+            chunk_pages = s3_client.get_paginator('list_objects_v2').paginate(
+                Bucket=self.bucket_name, Prefix=source_doc_prefix
+            )
+            return [
+                chunk_obj['Key']
+                for chunk_page in chunk_pages
+                for chunk_obj in chunk_page.get('Contents', [])
+            ]
 
-                nodes = list(executor.map(
+        # List each document's chunks concurrently. The per-document
+        # list_objects_v2 calls previously ran one at a time on the main thread
+        # and were a large part of readback. executor.map keeps yield order.
+        with concurrent.futures.ThreadPoolExecutor(max_workers=num_threads) as list_executor, \
+             concurrent.futures.ThreadPoolExecutor(max_workers=num_threads) as download_executor:
+
+            for source_doc_prefix, chunk_keys in zip(
+                source_doc_prefixes,
+                list_executor.map(_list_chunk_keys, source_doc_prefixes)
+            ):
+
+                nodes = list(download_executor.map(
                     self._download_chunk,
                     chunk_keys,
                     repeat(s3_client)
                 ))
 
                 logger.debug(f'Yielding source document [source: {source_doc_prefix}, num_nodes: {len(nodes)}]')
-            
+
                 yield SourceDocument(nodes=nodes)
 
 class S3ChunkUploader(BaseComponent):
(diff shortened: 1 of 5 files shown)


lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/load/s3_based_docs.py
```diff
@@ -293,26 +294,50 @@ def download(self):
 
         logger.debug(f'Started getting source documents from S3 [bucket: {self.bucket_name}, collection_path: {collection_path}, num_prefixes: {len(source_doc_prefixes)}]')
 
-        with concurrent.futures.ThreadPoolExecutor(max_workers=GraphRAGConfig.extraction_num_threads_per_worker) as executor:
+        num_threads = GraphRAGConfig.extraction_num_threads_per_worker
 
-            for source_doc_prefix in source_doc_prefixes:
-                
+        with concurrent.futures.ThreadPoolExecutor(max_workers=num_threads) as download_executor, \
```
acarbonetto, line 308: Connection pool exhaustion at recommended thread counts: 

  This is the strongest concern because it directly undercuts the PR's own tuning advice. Listing now runs concurrently with downloading, so peak S3 connections double to 2*num_threads. But the s3 client (config.py:458) is built with session.client('s3') and no max_pool_connections, so it defaults to 10. The PR benchmarks recommend extraction_num_threads_per_worker of 32–64 → up to 128 concurrent connections against a pool of 10, producing "Connection pool is full, discarding connection" warnings and connection thrashing that erodes the intended speedup.

  **Fix**: set botocore.config.Config(max_pool_connections=...) sized to at least 2*num_threads on the s3 client.

acarbonetto, line 308: **Executors can outlive early generator abandonment**

  The two ThreadPoolExecutors are opened in a with block inside the download() generator, so they're only shut down when GeneratorExit is thrown at the yield — i.e. on explicit .close() or GC of the generator. If a consumer stops early(for doc in download(): break), cleanup relies on GC finalizing the generator. That's usually prompt under CPython refcounting, but if a reference cycle keeps the generator alive (or on a non-refcounting runtime), the 2*num_threads worker threads and any in-flight list_objects_v2 calls persist for an unbounded time. The PR's "abandoning the generator early dispatches no downloads" claim holds for downloads but not for these threads.

**Suggested fix** — make close deterministic at the consumer rather than GC-dependent. The only caller (line 502) is itself a re-yielding generator, so wrap the inner generator so its own GeneratorExit propagates the close down:

```
  import contextlib

  with contextlib.closing(self._downloader.download()) as docs:
      for doc in docs:
          doc_count += 1
          yield doc
```

  This unwinds the two with blocks and calls shutdown(wait=True) deterministically on early exit. Since download() holds thread pools across yield, worth adding a docstring note that it must be fully consumed or closed(contextlib.closing/with) if it's ever called directly.

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/load/s3_based_docs.py
```diff
@@ -293,26 +294,50 @@ def download(self):
...
+                next_prefix = next(remaining_prefixes, None)
+                if next_prefix is not None:
+                    in_flight.append(
+                        (next_prefix, list_executor.submit(_list_chunk_keys, next_prefix))
+                    )
+
+                # Download the current document's chunks only, then yield, so at
+                # most one document's chunk data is resident at a time and an
+                # abandoned generator dispatches no downloads for unconsumed docs.
+                nodes = list(download_executor.map(
```
acarbonetto, line 342: **Add tests for the two uncovered error/lifecycle paths in S3ChunkDownloader.download()**

  The concurrent-listing tests cover the happy-path invariants well (order preservation, listing concurrency, bounded window, lazy per-document downloads). Two paths introduced by the overlap change are currently untested:

  1. A listing call that raises. When _list_chunk_keys fails, listing.result() re-raises and the two with blocks unwind via shutdown(wait=True). Nothing asserts the error surfaces to the consumer (vs. being swallowed) or that already-yielded documents are unaffected. Suggested test: make paginate raise for one prefix's chunk-listing call and assert list(downloader.download()) propagates that exception.

  2. Early generator abandonment. No test covers a consumer that stops early (break / partial consumption). This is the path where the executors' cleanup relies on GeneratorExit being delivered. Suggested test: consume one document then break, and assert cleanup happened deterministically — e.g. via contextlib.closing(...) around download(), assert no chunks of later documents were downloaded after the break, and (if a cleanup fix lands) that the executors were shut down.

  Both are small additions to the existing TestS3ChunkDownloaderParallelListing harness, which already has the background-consumer + Event scaffolding needed to drive a mid-stream stall or break. Test 1 also exercises the finding on delayed error propagation, so it does double duty.

acarbonetto: Here are both tests, written to drop into the existing TestS3ChunkDownloaderParallelListing class and matching its style (same mock_config patch, paginate side-effect, background-consumer scaffolding).
  
  Case 1 — a listing call that raises

```
      @patch('graphrag_toolkit.lexical_graph.indexing.load.s3_based_docs.GraphRAGConfig')
      def test_listing_error_propagates_to_consumer(self, mock_config):
          """A failure while listing a document's chunks must surface to the
          consumer rather than being swallowed by executor shutdown, and must not
          corrupt documents already yielded before the failing one."""
          prefixes = ['p/c/doc-0/', 'p/c/doc-1/', 'p/c/doc-2/']

          def paginate(**kwargs):
              if kwargs.get('Delimiter') == '/':
                  return [{'CommonPrefixes': [{'Prefix': p} for p in prefixes]}]
              if kwargs['Prefix'] == 'p/c/doc-1/':
                  raise RuntimeError('transient S3 listing error')
              return [{'Contents': [{'Key': kwargs['Prefix'] + 'c0.json'}]}]

          mock_s3 = MagicMock()
          mock_config.s3 = mock_s3
          mock_config.extraction_num_threads_per_worker = 2
          paginator = MagicMock()
          paginator.paginate.side_effect = paginate
          mock_s3.get_paginator.return_value = paginator

          downloader = S3ChunkDownloader(
              key_prefix='p', collection_id='c', bucket_name='b',
              fn=lambda node: node,
          )

          with patch.object(
              S3ChunkDownloader, '_download_chunk',
              side_effect=lambda key, client: TextNode(id_=key, text=''),
          ):
              gen = downloader.download()
              # doc-0 lists and downloads cleanly and is yielded.
              first = next(gen)
              assert [n.id_ for n in first.nodes] == ['p/c/doc-0/c0.json']
              # doc-1's listing raised; the error must surface, not be swallowed.
              with pytest.raises(RuntimeError, match='transient S3 listing error'):
                  next(gen)
```

  Case 2 — early generator abandonment

```  
      @patch('graphrag_toolkit.lexical_graph.indexing.load.s3_based_docs.GraphRAGConfig')
      def test_early_abandonment_dispatches_no_further_downloads(self, mock_config):
          """A consumer that stops after one document must not cause later
          documents' chunks to download, and closing the generator must tear the
          executors down deterministically rather than relying on GC."""
          import contextlib

          num_threads = 2
          num_docs = 10
          prefixes = [f'p/c/doc-{i}/' for i in range(num_docs)]
          layout = {p: [p + 'c0.json', p + 'c1.json'] for p in prefixes}
          downloaded = []
          downloaded_lock = threading.Lock()

          def paginate(**kwargs):
              if kwargs.get('Delimiter') == '/':
                  return [{'CommonPrefixes': [{'Prefix': p} for p in prefixes]}]
              return [{'Contents': [{'Key': k} for k in layout[kwargs['Prefix']]]}]

          mock_s3 = MagicMock()
          mock_config.s3 = mock_s3
          mock_config.extraction_num_threads_per_worker = num_threads
          paginator = MagicMock()
          paginator.paginate.side_effect = paginate
          mock_s3.get_paginator.return_value = paginator
  
          downloader = S3ChunkDownloader(
              key_prefix='p', collection_id='c', bucket_name='b',
              fn=lambda node: node,
          )
  
          def _download(key, client):
              with downloaded_lock:
                  downloaded.append(key)
              return TextNode(id_=key, text='')

          # closing() delivers GeneratorExit deterministically on block exit,
          # unwinding the executors' `with` blocks without waiting for GC.
          with patch.object(S3ChunkDownloader, '_download_chunk', side_effect=_download):
              with contextlib.closing(downloader.download()) as gen:
                  first = next(gen)
                  assert [n.id_ for n in first.nodes] == layout['p/c/doc-0/']
                  # abandon the generator here — no further next() calls.

          # Give any leaked worker threads a chance to (wrongly) run ahead.
          time.sleep(0.2)
          with downloaded_lock:
              later_doc_downloads = [
                  key for key in downloaded if not key.startswith('p/c/doc-0/')
              ]
          assert later_doc_downloads == [], (
              f'downloads dispatched for unconsumed documents after abandonment: '
              f'{later_doc_downloads}'
          )
```

**Notes on the two tests**

  - Case 1 drives the generator manually with next() so it can assert doc-0 is yielded before doc-1's listing failure surfaces — that's what proves the error propagates rather than being swallowed during shutdown(wait=True). pytest is already imported in the file.
  - Case 2 passes against the current code because downloads are already lazy per-document (the executors get GeneratorExit at the yield). Its real value is as a regression guard: it locks in that early abandonment dispatches nothing further, and the contextlib.closing wrapper is exactly the caller-side fix suggested for finding #2 — so if you land that fix at line 502, this test documents and protects it.
  - If you want Case 2 to also assert the executors were actually shut down (not just that no downloads fired), that requires capturing the executor instances — e.g. patch concurrent.futures.ThreadPoolExecutor with a wrapper that records instances, then assert _shutdown is True on each after the with block. That's more invasive and couples to internals; the download-dispatch assertion above is the lighter, behavior-focused check I'd start with.

acarbonetto (commented):
Honestly, this fix looks pretty good.  

If you only update the fix to set the 2*num_threads for the max_pool_connections, then I think we would be good.  

Adding extra test harness would be a nice add too.

acarbonetto (approved):

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/load/s3_based_docs.py
noel-improv: Fixed in d4dfeb4d. The S3 client now gets `max_pool_connections` at 2 × `extraction_num_threads_per_worker`, floored at botocore's 10 so the default 4-thread case doesn't drop to 8. Other services keep their defaults.

I'd measured and confirmed the same thing from the read side before seeing this. On a 50-object batch, 64 threads against a pool of 10 takes 321.6ms; the same threads with a pool of 64 take 99.1ms. Below 32 threads the pool makes no difference.

Naming note: `DEFAULT_MAX_POOL_CONNECTIONS` already exists in `neptune_graph_stores.py`, I used `BOTOCORE_DEFAULT_MAX_POOL_CONNECTIONS`

noel-improv: Fixed also flagging one thing: I reverted the fix to check the test caught it and everything still passed. Under CPython the break drops the outer generator, clearing its frame drops the inner one and refcounting unwinds both with blocks anyway.

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/load/s3_based_docs.py
noel-improv: Listing-error test added but dropped the early-abandonment one as it passed with the `contextlib.closing` fix reverted, would have sat in the suite looking like coverage for the lifecycle change without being it. I can re-add if you there's a good case.

acarbonetto: all good.  these are really minor edge cases.

acarbonetto:
did you want to merge this to a feature branch & benchmark before we merge to `main`?

PR #421, opened 2026-07-23
[FEATURE] Add S3 doc-store option to WikiHow extraction benchmark

noel-improv:
## Description

The extraction benchmark hardcoded `FileBasedDocs` (local disk), so it could not measure the cost of S3-backed document storage. This adds a `BENCHMARK_DOC_STORE` switch in the shared `run_benchmark_extract` path, so any dataset (ConcurrentQA, WikiHow, PGA) can use it. It defaults to local disk, so with the variable unset every dataset runs as before. The scale runs below use WikiHow.

## Changes

- `benchmark_extract.py`: read `BENCHMARK_DOC_STORE` (`file` default, or `s3`). When `s3`, use `S3BasedDocs` with `BENCHMARK_S3_JSONL` selecting the per-chunk or JSONL path, and `collection_id=None` (timestamp per run) so repeated runs land in separate collections. Log the resolved store, flag, and collection id.
- `benchmark_extract.py`: when `s3` is selected, validate `AWS_REGION_NAME`/`S3_RESULTS_BUCKET`/`S3_RESULTS_PREFIX` are set and raise a `ValueError` naming any that are missing, instead of a bare `KeyError` (these vars are not otherwise required in non-batch prototype runs).
- `build-tests.sh`: propagate `BENCHMARK_DOC_STORE` and `BENCHMARK_S3_JSONL` into `.env.testing` so they reach the notebook.

## Problem

`benchmark_extract` built `FileBasedDocs` unconditionally, so there was no way to benchmark extraction against S3. The switch lives in the shared helper, not one dataset's test, because the document store is a generic `NodeHandler` swap that any run should be able to opt into. The `build-tests.sh` change is needed too: without it the variables stay in the caller's shell, the notebook never sees them, and the run silently falls back to `FileBasedDocs`.

Related issue (if any): #

## Testing

Ran the WikiHow 5,000-doc extraction on a SageMaker notebook across all three settings: `file` (33m 33s), `s3` with `BENCHMARK_S3_JSONL=false` (42m 21s), and `s3` with `BENCHMARK_S3_JSONL=true`. Each run logged the expected `Doc store:` line and produced the expected object layout (16,536 per-chunk objects vs one JSONL per source document). The default (`file`, unset) reproduces the prior behaviour unchanged.

## Checklist

No new files (existing headers unchanged). Default behaviour is identical to before when the new variables are unset.

---


benchmarks/scripts/benchmark_extract.py
```diff
@@ -69,9 +69,33 @@ def run_benchmark_extract(handler: IntegrationTestHandler,
...
+        missing = [
+            var for var in ('AWS_REGION_NAME', 'S3_RESULTS_BUCKET', 'S3_RESULTS_PREFIX')
+            if not os.environ.get(var)
+        ]
+        if missing:
+            raise ValueError(
+                f"BENCHMARK_DOC_STORE=s3 requires {', '.join(missing)} to be set"
+            )
+        extracted_docs = S3BasedDocs(
+            region=os.environ['AWS_REGION_NAME'],
```
acarbonetto, line 121: Not used by S3BasedDocs. 

region is passed from AWS_REGION_NAME into S3BasedDocs, which stores it as an attribute but never uses it. All actual S3 calls go through GraphRAGConfig.s3, whose region resolves from os.environ['AWS_REGION'] or the boto default (config.py:532) — never AWS_REGION_NAME. Since build-tests.sh only exports AWS_REGION_NAME, if the boto default region differs from the bucket's region, uploads/downloads hit the wrong region and fail with redirect/mismatch errors — even though the benchmark "validated" that AWS_REGION_NAME was set. This risk is shared with the existing batch code path.

Set it once in `run_benchmark_extract`, right where the other `GraphRAGConfig.*` values are already being configured (lines 54–58), rather than trying to teach S3BasedDocs to honour its own region attribute (which would diverge from how every other client resolves region):

```
      GraphRAGConfig.extraction_llm = os.environ.get(
          'TEST_EXTRACTION_LLM', 'us.anthropic.claude-sonnet-4-6'
      )
      GraphRAGConfig.extraction_batch_size = 15000
      GraphRAGConfig.extraction_num_workers = 2

      # AWS_REGION_NAME is the region the benchmark harness exports; feed it into
      # GraphRAGConfig so every boto3 client (s3 doc-store, bedrock) targets it.
      # GraphRAGConfig.aws_region otherwise resolves from AWS_REGION / boto default,
      # which may not match the bucket's region.
      aws_region = os.environ.get('AWS_REGION_NAME')
      if aws_region:
          GraphRAGConfig.aws_region = aws_region
```

integration-tests/test-scripts/graphrag_toolkit_tests/benchmark_extract.py
acarbonetto: num_extracted = sum(1 for _ in extracted_docs) invokes S3BasedDocs.__iter__, which lists and downloads every chunk/JSONL object back from S3 (16,536 objects on the WikiHow run) solely to compute a count. This adds a full second pass of S3 GET traffic and latency right after extraction already wrote them. A list-only count (or capturing the count during upload) would avoid the re-download.

integration-tests/build-tests.sh
```diff
@@ -378,6 +378,12 @@ fi
 if [[ "$BENCHMARK_IS_PROTOTYPE" ]]; then
 	echo "export BENCHMARK_IS_PROTOTYPE=$BENCHMARK_IS_PROTOTYPE" >> lexical-graph-examples/.env.testing
 fi
+if [[ "$BENCHMARK_DOC_STORE" ]]; then
```
acarbonetto, line 410: please document new environment variables

integration-tests/build-tests.sh
```diff
@@ -378,6 +378,12 @@ fi
...
+	echo "export BENCHMARK_DOC_STORE=$BENCHMARK_DOC_STORE" >> lexical-graph-examples/.env.testing
+fi
+if [[ "$BENCHMARK_S3_JSONL" ]]; then
```
acarbonetto, line 413: please document new environment variables

integration-tests/test-scripts/graphrag_toolkit_tests/benchmark_extract.py
```diff
@@ -69,9 +69,33 @@ def run_benchmark_extract(handler: IntegrationTestHandler,
...
-    extracted_docs = FileBasedDocs(
-        docs_directory=os.path.join(data_dir, dataset_name, 'extracted'),
-        collection_id=dataset_name
+    doc_store = os.environ.get('BENCHMARK_DOC_STORE', 'file').lower()
+    if doc_store == 's3':
+        missing = [
+            var for var in ('AWS_REGION_NAME', 'S3_RESULTS_BUCKET', 'S3_RESULTS_PREFIX')
+            if not os.environ.get(var)
+        ]
+        if missing:
```
acarbonetto: WikihowBenchmarkExtract runs with use_batch defaulting to True, and the BatchConfig block (lines 62–69) dereferences the same env vars with bracket access (os.environ[...]) first. If any is missing, a bare `KeyError` fires there before the new ValueError runs — so the clearer message only helps non-batch prototype datasets, not the documented WikiHow scale run the PR targets.  

Add validation before lines 60 to ensure that the right environments are set (if use_batch is set or not). Something like: 
```

      doc_store = os.environ.get('BENCHMARK_DOC_STORE', 'file').lower()
  
      # Validate all required env vars up front, before anything dereferences them,
      # so both the batch path and the S3 doc-store path get a clear error rather
      # than a bare KeyError (whichever runs first).
      required_vars = set()
      if use_batch:
          required_vars.update(
              ('AWS_REGION_NAME', 'S3_RESULTS_BUCKET', 'S3_RESULTS_PREFIX', 'BATCH_INFERENCE_ROLE')
          )
      if doc_store == 's3':
          required_vars.update(('AWS_REGION_NAME', 'S3_RESULTS_BUCKET', 'S3_RESULTS_PREFIX'))

      missing = sorted(var for var in required_vars if not os.environ.get(var))
      if missing:
          raise ValueError(
              f"use_batch={use_batch}, BENCHMARK_DOC_STORE={doc_store} "
              f"requires {', '.join(missing)} to be set"
          )

```

acarbonetto (commented):
Looks good.  A couple of small comments to address.

benchmarks/scripts/benchmark_extract.py
noel-improv: Fixed. `GraphRAGConfig.aws_region` is now set from `AWS_REGION_NAME` alongside the other config values. Confirmed the region argument is stored and never read on both `BatchConfig` and `S3BasedDocs`.

integration-tests/test-scripts/graphrag_toolkit_tests/benchmark_extract.py
noel-improv: Fixed. Now a delimiter listing

integration-tests/build-tests.sh
noel-improv: Done - both env.template copies and the benchmarks README, matching the existing format.

noel-improv:
@acarbonetto Heads up: main moved these files from `integration-tests/test-scripts/graphrag_toolkit_tests/` to `benchmarks/scripts/`, so your comments anchor to old paths. The changes are in `benchmarks/scripts/benchmark_extract.py` now.

oussamahansal (approved):
LGTM

Issue #324, opened 2026-06-16
[FEATURE] Pluggable external chunk text store to reduce Neptune Analytics memory usage


mykola-pereyma:
> **Status: in progress.** Interface in #436, wiring in #437. The design below is the original proposal; where the implementation diverged, it says so.

## Problem Statement

`ChunkGraphBuilder` stores raw chunk text as a graph property:
```python
'chunk.value = params.text'
```

At retrieval time, chunk text is accessed at these specific locations:
- `keyword_vss_provider.py:81` — `RETURN c.value AS content` (explicit content fetch for keyword search)
- `statement_enhancement.py:103` — `node.node.metadata['chunk']['value']` (LLM synthesis context)
- `traversal_based_base_retriever.py:136` — only when `include_chunk_details=True` (opt-in, **defaults to False**)
- `update_chunk_metadata.py:18` — `chunk.metadata.pop('value', None)` (moves value from metadata dict)

**Note:** The main traversal-based retriever does NOT fetch chunk text by default (`include_chunk_details=False`). The memory cost is incurred regardless — Neptune keeps all node properties in-memory even when not queried.

For 5M chunks averaging 500 tokens (~2KB text), this consumes ~10GB of Neptune Analytics graph memory for text that is only needed at final answer synthesis.

**Key insight:** The traversal-based retriever already sets `value: NULL` in its projection — the architecture already treats chunk text as a late-binding fetch. This confirms external storage is architecturally aligned.

## Proposed Solution

1. **`ChunkStore` protocol (backend-agnostic interface):**
```python
class ChunkStore(ABC):
    # get() is concrete: it defaults to get_batch([chunk_id]).get(chunk_id),
    # so a backend only implements get_batch.
    def get(self, chunk_id: str) -> Optional[str]: ...
    @abstractmethod
    def put(self, chunk_id: str, text: str) -> None: ...
    @abstractmethod
    def get_batch(self, chunk_ids: List[str]) -> Dict[str, str]: ...
```

`get_batch()` matches on chunk id alone and does not require the `__EXTRACTED_FROM__` link `get_chunks_query` requires, since `put()` never writes that link. Any backend has to honor this.

2. **Reference implementations:**
   - `S3ChunkStore` — S3 objects keyed by chunk_id, batch get via ThreadPoolExecutor
   - `InGraphChunkStore` — current behavior (reads `chunk.value`), for backward compat
   - Interface supports any object store (DynamoDB, Redis, local FS, third-party)

[...]

PR #436, opened 2026-07-27
[FEATURE] Add pluggable ChunkStore interface for chunk text storage

noel-improv:
## Description

Chunk text is stored inline as a `chunk.value` property on every `__Chunk__` node. Neptune keeps all node properties in memory whether or not a query reads them, so at scale this costs memory for text that is only needed at answer synthesis. This PR adds a `ChunkStore` interface for chunk text and an `InGraphChunkStore` that keeps today's behavior as the default. Nothing is wired to it yet, so there is no behavior change. Wiring the write and read paths (#437) and adding an S3 backend are follow-ups that merge after this.

## Changes

- `ChunkStore`, an abstract interface for chunk text. `put` and `get_batch` are abstract; `get` defaults to `get_batch([chunk_id]).get(chunk_id)`, so a backend only implements `get_batch`. `put()` has a TODO noting there is no batched write yet: `ChunkGraphBuilder` calls `put()` once per chunk while reads already batch, so writes are where the remaining ingestion cost sits.
- `InGraphChunkStore`, which reads and writes `chunk.value` on `__Chunk__` nodes. `get_batch()` matches on chunk id alone. It does not require the `__EXTRACTED_FROM__` link to a `__Source__` that `get_chunks_query` requires, because `put()` never creates that link: requiring it on read would make text written through the store unreadable from the same store.
- `ChunkStoreFactory`, a registration pattern for backends. Registered factories are tried in registration order. If none match and no `chunk_info` was given, the in-graph store is the default; a non-empty `chunk_info` nothing recognizes raises, so a typo'd URI fails loudly. This differs from `GraphStoreFactory`, which has no default and raises for empty `graph_info` too.
- `try_create` is annotated `Optional[ChunkStore]`, since it returns `None` when a factory doesn't recognize `chunk_info`. `GraphStoreFactoryMethod` and `VectorIndexFactoryMethod` document returning `None` while annotated otherwise; left alone to keep this PR to one concern.
- A shared `graph_utils.to_params()` helper for the `UNWIND $params` wrapping already used by `GraphBuilder._to_params`, so `InGraphChunkStore.put()` doesn't duplicate it.

## Problem

Related issue (if any): #324

## Testing

Unit tests cover the abstract contract, `InGraphChunkStore` against a mocked graph client, and factory registration and dispatch.

`tests/integration/storage/chunk/test_chunk_store_live_neo4j.py` runs the store against a real Neo4j instance, skipped unless `NEO4J_TEST_URI` is set (the container command is in the file header). Mocked tests assert on the shape of the query string, so they can't catch a query that is well-formed but matches nothing. That is the bug this caught: `put()` wrote a chunk `get_batch()` could not read back. Four of its five cases fail against the pre-fix query.

## Checklist

- [ ] Documentation updated (if applicable) — internal interface, no user-facing config surface until an external backend lands

---

### DIFF
--- a/lexical-graph/src/graphrag_toolkit/lexical_graph/storage/chunk/chunk_store_factory_method.py
+++ b/lexical-graph/src/graphrag_toolkit/lexical_graph/storage/chunk/chunk_store_factory_method.py
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import abc
+from typing import Optional
 
 from graphrag_toolkit.lexical_graph.storage.chunk.chunk_store import ChunkStore
 
@@ -16,7 +17,7 @@ class ChunkStoreFactoryMethod(abc.ABC):
     factory.
     """
     @abc.abstractmethod
-    def try_create(self, chunk_info: str, **kwargs) -> ChunkStore:
+    def try_create(self, chunk_info: str, **kwargs) -> Optional[ChunkStore]:
         """
         Attempt to create a ChunkStore from chunk_info. Return None if this
         factory doesn't recognize chunk_info.
--- a/lexical-graph/src/graphrag_toolkit/lexical_graph/storage/chunk/chunk_store_factory_method.py
+++ b/lexical-graph/src/graphrag_toolkit/lexical_graph/storage/chunk/chunk_store_factory_method.py
@@ -1,0 +1,24 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+import abc
+
+from graphrag_toolkit.lexical_graph.storage.chunk.chunk_store import ChunkStore
+
+
+class ChunkStoreFactoryMethod(abc.ABC):
+    """
+    Factory method pattern for creating ChunkStore instances.
+
+    Mirrors GraphStoreFactoryMethod: subclasses attempt to create a
+    ChunkStore from the given chunk_info, returning None if they don't
+    recognize it so ChunkStoreFactory can move on to the next registered
+    factory.
+    """
+    @abc.abstractmethod
+    def try_create(self, chunk_info: str, **kwargs) -> ChunkStore:
+        """
+        Attempt to create a ChunkStore from chunk_info. Return None if this
+        factory doesn't recognize chunk_info.
+        """
+        raise NotImplementedError
--- a/lexical-graph/src/graphrag_toolkit/lexical_graph/storage/chunk/in_graph_chunk_store.py
+++ b/lexical-graph/src/graphrag_toolkit/lexical_graph/storage/chunk/in_graph_chunk_store.py
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import logging
-from typing import Dict, List, Optional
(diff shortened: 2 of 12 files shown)


acarbonetto:
@noel-improv please update #324.

lexical-graph/src/graphrag_toolkit/lexical_graph/storage/chunk/chunk_store_factory_method.py
```diff
@@ -0,0 +1,24 @@
...
+
+    Mirrors GraphStoreFactoryMethod: subclasses attempt to create a
+    ChunkStore from the given chunk_info, returning None if they don't
+    recognize it so ChunkStoreFactory can move on to the next registered
+    factory.
+    """
+    @abc.abstractmethod
+    def try_create(self, chunk_info: str, **kwargs) -> ChunkStore:
+        """
+        Attempt to create a ChunkStore from chunk_info. Return None if this
```
acarbonetto, line 22: nit: consider updating to return Optional[ChunkStore]
but this would be a bigger fix

lexical-graph/src/graphrag_toolkit/lexical_graph/storage/chunk/in_graph_chunk_store.py
```diff
@@ -0,0 +1,70 @@
...
+
+        rows = self.graph_client.execute_query(query, params)
+
+        return {
+            row['result']['chunk']['chunkId']: row['result']['chunk']['value']
+            for row in rows
+            if row['result']['chunk'].get('value') is not None
+        }
+
+    def put(self, chunk_id: str, text: str) -> None:
```
acarbonetto, line 58: consider adding `put_batch` equivalent
maybe add a TODO comment to do this in the future if it's relevant

acarbonetto (commented):
Looks good - pelase update the PR description for this:

`get_batch` existence semantics differ from `get_chunks_query` — The new read path does not require the `__EXTRACTED_FROM__ → __Source__` link, but the PR body claims it matches `get_chunks_query`. Orphan chunks (source link never created) would start appearing if this is later swapped in as a drop-in replacement, changing retrieval results. The class docstring documents this as deliberate, directly contradicting the PR description.

oussamahansal (approved):
LGTM

lexical-graph/tests/integration/storage/chunk/test_chunk_store_live_neo4j.py
```diff
@@ -0,0 +1,78 @@
...
+    reason='set NEO4J_TEST_URI to a running Neo4j instance to run this live test',
+)
+
+
+@pytest.fixture
+def graph_client():
+    client = GraphStoreFactory.for_graph_store(NEO4J_TEST_URI)
+    client.execute_query('MATCH (n) DETACH DELETE n', {})
+    yield client
+    client.execute_query('MATCH (n) DETACH DELETE n', {})
```
acarbonetto: this could be dangerous to do, even in an integration test.  
consider checking for an empty database first, and failing if its not empty instead.

noel-improv: done, I added the empty db check and tested against a real Neo4j run

PR #437, opened 2026-07-29
[FEATURE] Wire ChunkGraphBuilder and read paths to ChunkStore

noel-improv:
## Description

The `ChunkStore` interface (`InGraphChunkStore`, `ChunkStoreFactory`) landed as an interface-only addition with nothing calling it. This PR wires it into the write path and the two read paths that actually need chunk text today: `ChunkGraphBuilder` (write), `KeywordVSSProvider` (query-time keyword extraction), and `TraversalBasedBaseRetriever`'s `include_chunk_details` branch (synthesis context). `InGraphChunkStore` stays the default, so there's no behavior change for the in-graph backend today, this just gets these callers off direct `chunk.value` graph reads/writes so a future external backend (e.g. S3) can be swapped in without touching them. `chunk_utils.get_chunks_query()` is left as-is; its only caller lives in the deprecated retrievers package.

Stacked on #436 (the `ChunkStore` interface PR, not yet merged) - this diff will include those changes until it merges, then shrink to just this PR's commits.

## Changes

- `ChunkGraphBuilder.build()` writes chunk text via `ChunkStore.put()` instead of an inline `chunk.value` SET, split from the (now separate, conditional) metadata-properties query.
- `KeywordVSSProvider._get_chunk_content()` reads via `ChunkStore.get_batch()` instead of a hand-rolled Cypher query.
- `TraversalBasedBaseRetriever.get_statements_by_topic_and_source()`'s `include_chunk_details` branch reuses the chunk text the main query already returns when present, and falls back to `ChunkStore.get_batch()` only for chunks where it isn't, avoiding a redundant round trip to re-fetch text the query already has.
- `GraphBatchClient` (what graph builders actually receive as their graph client) gains an `execute_query` passthrough, closing a gap where it duck-typed the write side of `GraphStore` but not the read side.
- `ChunkGraphBuilder` no longer lets a `chunk_external_properties` field named `value` overwrite the chunk text `ChunkStore.put()` just wrote.

## Problem

Related issue (if any): #324

## Testing

Unit tests cover the write/read wiring, the `get_batch()` existence-semantics change, the `include_chunk_details` fetch/fallback split, the `GraphBatchClient.execute_query` passthrough, and the metadata `value`-key collision fix. Also added a skip-gated live test (`NEO4J_TEST_URI`) exercising the write path and both read paths against a real Neo4j instance, since mocked tests can't catch a malformed query or a client that doesn't duck-type the way a factory expects, both real bugs this work caught during development before they reached review.

## Checklist

- [ ] Documentation updated (if applicable) - not applicable yet, no user-facing config surface until an external backend exists

`InGraphChunkStore` remains the default backend and its behavior is unchanged for existing users; the `get_batch()` existence-semantics change only affects code paths introduced by this PR and #436, nothing that shipped before them.

---

### DIFF
--- a/lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/build/chunk_graph_builder.py
+++ b/lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/build/chunk_graph_builder.py
@@ -5,6 +5,7 @@ import logging
 from typing import Any
 
 from graphrag_toolkit.lexical_graph.storage.graph import GraphStore
+from graphrag_toolkit.lexical_graph.storage.chunk_store_factory import ChunkStoreFactory
 from graphrag_toolkit.lexical_graph.indexing.build.graph_builder import GraphBuilder
 
 from llama_index.core.schema import BaseNode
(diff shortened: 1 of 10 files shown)


acarbonetto:
Please update the PR description to say `Fixes #324` - which will automatically close the issue when this PR is merged.

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/build/chunk_graph_builder.py
```diff
@@ -57,33 +58,35 @@ def build(self, node:BaseNode, graph_client: GraphStore, **kwargs:Any):
 
             logger.debug(f'Inserting chunk [chunk_id: {chunk_id}]')
 
-            statements_c = [
-                '// insert chunks',
-                'UNWIND $params AS params',
-                f'MERGE (chunk:`__Chunk__`{{{graph_client.node_id("chunkId")}: params.chunk_id}})'
-            ]
+            chunk_store = ChunkStoreFactory.for_chunk_store(graph_store=graph_client)
+            chunk_store.put(chunk_id, node.text)
```
oussamahansal, line 62: is the extra round trip intended here, or should the metadata `SET` be folded back into the `put()` for the in-graph store until S3 actually lands?

oussamahansal (approved):
lgtm

acarbonetto (approved):

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/build/graph_batch_client.py
```diff
@@ -77,7 +77,21 @@ def node_id(self, id_name:str):
...
+        batched, so this passes straight through rather than queuing.
+
+        Args:
+            query: The query to execute.
+            parameters: Query parameters.
+
+        Returns:
+            The graph client's query result.
+        """
+        return self.graph_client.execute_query(query, parameters, **kwargs)
```
acarbonetto: execute_query doesn't accept `kwargs`.  Please fix before merging.

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/build/chunk_graph_builder.py
noel-improv: thanks for flagging this, it's intended: that second query only runs when `chunk_external_properties` is set and metadata can't fold into `put()` since `S3ChunkStore.put()` never touches the graph, the batched-write path is already tracked in the TODO on `ChunkStore.put`

PR #415, opened 2026-07-20
fix: return empty string for empty filter list in parse_metadata_filters_recursive

evanerwee:
## Summary

When `MetadataFilters` has an empty `filters` list, `parse_metadata_filters_recursive` returns `"()"` (from joining an empty list within parentheses). This truthy string bypasses the guard in `VersionManager._get_existing_source_nodes()` at line 70:

```python
if not where_clauses:  # "()" is truthy — guard fails
    return []
```

This causes the VersionManager to generate invalid Cypher:

```cypher
WHERE () AND coalesce(source.__aws__versioning__id_fields__, ...) = $param
```

Neptune Database (1.4.7.0) rejects this with:
```
MalformedQueryException: Invalid input 'A': expected whitespace, comment or a relationship pattern
```

## Fix

Return `''` (empty string) when `filter_strs` is empty, before reaching the condition-based join logic.

```python
if not filter_strs:
    return ''
```

## Testing

- Added 2 unit tests covering empty filter lists with AND and OR conditions
- Verified on Neptune Database 1.4.7.0: the versioning query now works correctly
- Confirmed via direct Neptune query testing that Neptune supports all other Cypher features used by the VersionManager (map literals, coalesce, `__` property names)

## Evidence

Direct Neptune testing results (from issue #408):

| Test | Query Pattern | Result |
|------|--------------|--------|
| Map literal in RETURN | `RETURN {id: id(n)} AS result` | ✅ Works |
| `__aws__` property names | `t.__aws__versioning__id_fields__` | ✅ Works |
| coalesce on versioning properties | `coalesce(t.__aws__versioning__valid_from__, -1)` | ✅ Works |
| **WHERE () AND filter** | `WHERE () AND id(n) IS NOT NULL` | **❌ MalformedQueryException** |
| Full versioning query (properly formed) | Complete pattern with valid WHERE | ✅ Works |

Fixes #408

PR #441, opened 2026-07-31
[FIX] Handle empty metadata filters across query backends

### LINKED ISSUE #423: [BUG] Empty MetadataFilters produces invalid queries in pg_vector, s3_vector, and graph_notebook_visualisation
## Package version

all versions (bug exists on current main)

## Package

lexical-graph

## Python version

all

## Operating System

all (library-level bug, not platform-specific)

## Description

Issue #408 identified that `parse_metadata_filters_recursive` produces invalid output when given an empty filters list (`MetadataFilters(filters=[], condition=FilterCondition.AND)`). PR #415 fixed the primary code path in `graph_utils.py`, but three parallel code paths share the same root cause and remain unfixed. Each independently produces invalid query syntax when receiving empty metadata filters.

#### 1. `pg_vector_indexes.py` — invalid SQL `()`

- **File:** `lexical-graph/src/graphrag_toolkit/lexical_graph/storage/vector/pg_vector_indexes.py` (lines ~143-150)
- **Symptom:** `parse_metadata_filters_recursive` returns `('()', [])` when `filter_strs` is empty. Produces invalid SQL like `WHERE ()`.
- **Fix:** Add `if not filter_strs: return ('', [])` early return before joining filter strings.

#### 2. `s3_vector_indexes.py` — invalid filter document `{"$and": []}`

- **File:** `lexical-graph/src/graphrag_toolkit/lexical_graph/storage/vector/s3_vector_indexes.py` (lines ~105-112)
- **Symptom:** Empty `filter_strs` produces `{"$and": []}` or `{"$or": []}`. Depending on the S3 Select/filter backend, this may be rejected as an invalid filter expression.
- **Fix:** Add `if not filter_strs: return {}` or `return None` early return before constructing the filter document.

#### 3. `graph_notebook_visualisation.py` — invalid Cypher `WHERE  OR ...`

- **File:** `lexical-graph/src/graphrag_toolkit/lexical_graph/visualisation/graph_notebook/graph_notebook_visualisation.py` (line ~178)
- **Symptom:** Even with PR #415's fix applied, the empty-string result `''` is appended to `where_clauses`. The list becomes truthy (contains one empty-string element), so the join produces `WHERE  OR (id(source) in [...])` — invalid Cypher with a leading space/OR.
- **Fix:** Filter falsy values before joining: `where_clauses = [c for c in where_clauses if c]`

## Steps to reproduce

All three paths are triggered the same way — passing an empty filters object:

```python
from llama_index.core.vector_stores.types import MetadataFilters, FilterCondition

empty_filters = MetadataFilters(filters=[], condition=FilterCondition.AND)
```

- **pg_vector:** Call any vector search method on `PgVectorIndexes` with `empty_filters` as the metadata filter argument.
- **s3_vector:** Call any vector search method on `S3VectorIndexes` with `empty_filters` as the metadata filter argument.
- **graph_notebook:** Invoke visualisation with a filter config that resolves to empty filters (e.g., no matching filter keys), causing the empty-string result to enter `where_clauses`.

## Error output

- **pg_vector:** Produces invalid SQL: `WHERE ()`
- **s3_vector:** Produces invalid filter document: `{"$and": []}` or `{"$or": []}`
- **graph_notebook:** Produces invalid Cypher: `WHERE  OR (id(source) in [...])`

## References

- #408 — original bug report for empty MetadataFilters
- #415 — fix for `graph_utils.py` (does not cover these three paths)

Pybsama:
## Description

Handle empty and recursively empty `MetadataFilters` without generating
malformed queries in the PGVector, S3 Vectors, and Graph Notebook paths.

## Changes

- return an empty PG filter fragment when no effective clauses remain, including
  after recursively empty groups are removed
- omit empty nested groups from S3 filter composition while preserving the
  existing unsupported behavior for `FilterCondition.NOT`
- remove empty openCypher clauses before Graph Notebook builds its `WHERE`
  expression
- add regression coverage for top-level and nested AND/OR groups, mixed valid
  filters, the S3 request payload, the final PG SQL, and Graph Notebook queries

The Graph Notebook consumer change relies on the empty-string helper contract
introduced by #415. Its tests stub that contract so this PR does not duplicate
or modify the `graph_utils.py` changes in #415.

## Problem

An empty PG filter currently becomes `()` and is emitted as `WHERE ()`. S3
serializes the same input as `$and: []` or `$or: []` and sends it as a filter.
After the graph helper returns an empty string, Graph Notebook still includes
that string in its clause list, producing either an empty `WHERE` or
`WHERE  OR (...)`.

Recursively empty groups need the same treatment; otherwise a parent group can
still produce a leading boolean operator or contain an empty S3 object.

Fixes #423

## Testing

- [ ] Integration tests added (not needed; the affected request/SQL boundaries
      are exercised with unit-level client and cursor fakes)
- [x] Tested manually (directly inspected the final PG SQL and S3
      `query_vectors` arguments)

Validation performed:

- focused regression suite: 69 passed
- full non-network suite: 1,810 passed, 1 skipped
- dependency-resolution compatibility test: 1 passed
- sdist and wheel builds
- repository license-header check
- `pre-commit run git-secrets --all-files`
- `git diff --check`

## Checklist

- [x] Documentation updated (not applicable; no public API change)
- [x] No breaking changes

---

### DIFF
--- a/lexical-graph/src/graphrag_toolkit/lexical_graph/storage/graph/graph_utils.py
+++ b/lexical-graph/src/graphrag_toolkit/lexical_graph/storage/graph/graph_utils.py
@@ -304,10 +304,15 @@ def parse_metadata_filters_recursive(metadata_filters:MetadataFilters) -> str:
                 raise ValueError(f'Expected MetadataFilters for FilterCondition.NOT, but found MetadataFilter')
             filter_strs.append(metadata_filter_to_opencypher_filter(metadata_filter))
         elif isinstance(metadata_filter, MetadataFilters):
-            filter_strs.append(parse_metadata_filters_recursive(metadata_filter))
+            nested_filter = parse_metadata_filters_recursive(metadata_filter)
+            if nested_filter:
+                filter_strs.append(nested_filter)
         else:
             raise ValueError(f'Invalid metadata filter type: {type(metadata_filter)}')
         
+    if not filter_strs:
+        return ''
+
     if metadata_filters.condition == FilterCondition.NOT:
         return f"(NOT {' '.join(filter_strs)})"
     elif metadata_filters.condition == FilterCondition.AND or metadata_filters.condition == FilterCondition.OR:
@@ -338,4 +343,3 @@ def filter_config_to_opencypher_filters(filter_config:FilterConfig) -> str:
     if filter_config is None or filter_config.source_filters is None:
         return ''
     return parse_metadata_filters_recursive(filter_config.source_filters)
-    
--- a/lexical-graph/src/graphrag_toolkit/lexical_graph/storage/vector/pg_vector_indexes.py
+++ b/lexical-graph/src/graphrag_toolkit/lexical_graph/storage/vector/pg_vector_indexes.py
@@ -138,8 +138,12 @@ def parse_metadata_filters_recursive(metadata_filters:MetadataFilters) -> tuple[
             (fragment, fragment_params) = parse_metadata_filters_recursive(metadata_filter)
         else:
             raise ValueError(f'Invalid metadata filter type: {type(metadata_filter)}')
-        filter_strs.append(fragment)
-        params.extend(fragment_params)
+        if fragment:
+            filter_strs.append(fragment)
+            params.extend(fragment_params)
+
+    if not filter_strs:
+        return ('', [])
 
     if metadata_filters.condition == FilterCondition.NOT:
         return (f"(NOT {' '.join(filter_strs)})", params)
--- a/lexical-graph/src/graphrag_toolkit/lexical_graph/storage/vector/s3_vector_indexes.py
+++ b/lexical-graph/src/graphrag_toolkit/lexical_graph/storage/vector/s3_vector_indexes.py
@@ -102,9 +102,17 @@ def parse_metadata_filters_recursive(metadata_filters:MetadataFilters) -> Dict[s
                 raise ValueError(f'Expected MetadataFilters for FilterCondition.NOT, but found MetadataFilter')
             filter_strs.append(to_s3_filter(metadata_filter))
         elif isinstance(metadata_filter, MetadataFilters):
-            filter_strs.append(json.dumps(parse_metadata_filters_recursive(metadata_filter)))
+            nested_filters = parse_metadata_filters_recursive(metadata_filter)
+            if nested_filters:
+                filter_strs.append(json.dumps(nested_filters))
         else:
             raise ValueError(f'Invalid metadata filter type: {type(metadata_filter)}')
+
+    if not filter_strs and (
+        metadata_filters.condition == FilterCondition.AND
+        or metadata_filters.condition == FilterCondition.OR
+    ):
+        return {}
 
     if metadata_filters.condition == FilterCondition.AND:       
         return json.loads(f'{{"$and": [{",".join(filter_strs)}]}}')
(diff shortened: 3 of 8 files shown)


Pybsama:
Validation update after hardening the nested-filter cases:

- PR head: `890c387ec0bd38dd1c49036d2e259b0f57fca5a0`.
- Current upstream `main`: `218bf37382412cd1ff72d8a5f64755c012976bb4`.
- GitHub could not update the branch because the existing OAuth token lacks the workflow scope needed for upstream workflow-file changes. I did not expand token permissions. A local merge-tree simulation against current `main` completed without conflicts and produced tree `75c267d97d6a93e782a3dc9aa4ce8eae84d8594b`, exactly matching the fully validated review tree.
- TDD exposed a nested-empty gap in the graph helper: the added cases produced 11 failures before the fix; the focused lexical-graph suites then passed 145/145, including real Graph Notebook helper coverage for AND, OR, NOT, nested-only, and mixed empty/valid filters.
- The exact merged tree passed the full Python 3.11 CI command: 2,109 passed, 15 skipped, 65.65% coverage (minimum 56%). The same focused 145 tests passed on Python 3.10 and 3.12.
- Bandit reported 0 medium/high findings, all lexical-graph Python files passed the license-header check, and wheel/sdist builds passed. Wheel SHA-256: `9307144ce8cf310210e00a29994580c9aa93ac4087b125d01733e75556171a37`; sdist SHA-256: `ea0b4d8b8b78bdbbb0dc9f13a980a59d534028c718bb0f33b6320261c6c985f4`.
- The historical Python 3.11 NLTK hardlink/pathsec failure was separately resolved on current `main` by #470; its exact regression path also passed in the merged-tree environment.
- Independent review after the nested fix: 0 critical, 0 important, 0 minor findings; ready to review.

noel-improv (approved):
Thanks @Pybsama PR looks good.

mykola-pereyma (approved):
LGTM. This correctly fixes the empty / recursively-empty `MetadataFilters` handling across all three backends (#423), and I verified each change against current `main`:

- **PGVector**: `if fragment:` guards the fragment and its params together, so values stay bound and there's no placeholder/param misalignment; returns `('', [])` when nothing remains.
- **S3 Vectors** (net-new — `main` has no empty-handling here): empty nested groups are omitted and an empty AND/OR returns `{}`, so no more `$and: []` / `{}`-in-array payloads. The deliberate "NOT unsupported" behavior is preserved and tested.
- **Graph Notebook** (net-new): empty clauses are dropped before building `WHERE`, eliminating the empty `WHERE` and `WHERE  OR (...)` cases.
- **openCypher**: the recursive-empty guard (`if nested_filter:`) is the needed addition on top of the top-level empty check #415 already landed.

Regression coverage is thorough and matches the exact failure modes: top-level and nested empty AND/OR across all three backends, nested-empty-next-to-valid, NOT-of-empty, final PG SQL with no `WHERE`, the S3 request omitting the `filter` kwarg, and Graph Notebook with no leading `OR`. All CI checks are green and the PR merges cleanly (no conflict with the already-merged #415 — git reconciles the shared top-level line via the 3-way merge).

Two non-blocking notes for later, neither affecting correctness:
- The empty-handling is now duplicated across the three parallel renderers. A future refactor could centralize it in a single `prune_empty(MetadataFilters) -> Optional[MetadataFilters]` pass so the renderers assume non-empty groups (it's behavior-identical for openCypher + PG; the one thing to decide would be whether S3's empty-`NOT` should stay a raise or become `{}`). Fine as a follow-up.
- The renderer functions are on the longer side, but that's from the pre-existing nested leaf-render helpers, not this change; the added guards follow the existing collect-then-dispatch convention.

Approving.

PR #456, opened 2026-08-07
[FEATURE] (benchmarks): Add PGA bio/stat split evaluation support

oussamahansal:
## Description

Allow running benchmarks against just the PGA bio or PGA stat subset by setting BENCHMARK_DATASET=pga_bio or BENCHMARK_DATASET=pga_stat.

## Changes

- Add 'pga_bio' and 'pga_stat' entries to QA_FILE_MAP
- Resolve data subdirectory to 'pga' for all pga* datasets in load_qa_pairs() and sync_benchmark_data_from_s3()
- Read dataset_name from BENCHMARK_DATASET env var in PgaBenchmarkQuery

## Problem

Related issue (if any): #

## Testing

- [x] Unit tests added/updated
- [ ] Integration tests added (as appropriate)
- [x] Existing tests pass (`pytest`)
- [x] Tested manually (describe below)

## Checklist

- [x] Code follows existing style and conventions
- [x] License headers present on new files
- [x] Documentation updated (if applicable)
- [x] No breaking changes (or clearly documented)

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #455, opened 2026-08-07
[FEATURE] (benchmarks): Add multi-retriever benchmark harness with beam-search retrievers

oussamahansal:
## Description

 Add multi-retriever benchmark harness with beam-search retrievers
## Changes

- Add support for running all retrievers in a single benchmark pass:
- Add topic-beam-chunk_only and topic_beam_search to retriever factory
- Import ChunkCosineSimilaritySearch, SemanticChunkBeamGraphSearch, SemanticGuidedChunkRetriever, TopicBeamSearch
- Tune shared sub-retriever params: max_search_results=10, max_context_tokens=3000, token_truncation_mode=per_topic_cap
- Create run_all_retrievers.sh loop script (extract+build once, query+evaluate per retriever)
- Create benchmark.concurrentqa.all-retrievers dataset file
- Forward BENCHMARK_ALL_RETRIEVERS and BENCHMARK_DATASET env vars through build-tests.sh to .env.testing
- Conditionally invoke run_all_retrievers.sh from run_test_suite.sh

## Problem

Related issue (if any): #

## Testing

- [ ] Unit tests added/updated
- [ ] Integration tests added (as appropriate)
- [x] Existing tests pass (`pytest`)
- [x] Tested manually (describe below)

## Checklist

- [x] Code follows existing style and conventions
- [x] License headers present on new files
- [x] Documentation updated (if applicable)
- [x] No breaking changes (or clearly documented)

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #457, opened 2026-08-07
[FEATURE] (benchmarks): Upload benchmark results to S3 after evaluation

oussamahansal:
Uses S3_RESULTS_BUCKET and S3_RESULTS_PREFIX env vars (same as integration_test_handler.py). No-ops with a warning when bucket is unset.

## Description

Add upload_benchmark_results_to_s3() and upload_all_benchmark_results_to_s3() to benchmarks/utils/s3_utils.py using boto3. The per-retriever upload is called at the end of run_benchmark_evaluate() after scores are written but before assertions run, ensuring results persist even if assertions fail.
Use S3_RESULTS_BUCKET and S3_RESULTS_PREFIX env vars (same as integration_test_handler.py). No-ops with a warning when bucket is unset.

## Changes

- upload_benchmark_results_to_s3(dataset, retriever_id, results_dir='benchmark-results') — uploads the retriever-specific directory to s3://<bucket>/<prefix>/benchmark-results/<dataset>/<retriever_id>/. No-ops with a warning if S3_RESULTS_BUCKET is unset. Uses AES256 server-side encryption and sets application/json for .json/.jsonl, text/plain otherwise.
- upload_all_benchmark_results_to_s3(results_dir='benchmark-results') — walks all <dataset>/<retriever>/ subdirs and uploads each.

## Problem

Related issue (if any): #

## Testing

## Checklist

---

### DIFF
--- a/benchmarks/scripts/benchmark_evaluate.py
+++ b/benchmarks/scripts/benchmark_evaluate.py
@@ -20,6 +20,7 @@ from typing import Dict, Any, Optional, List
 from benchmarks.scripts.integration_test_base import IntegrationTestBase
 from benchmarks.scripts.integration_test_handler import IntegrationTestHandler
 from benchmarks.utils.run_evaluation import evaluate_responses
+from benchmarks.utils.s3_utils import upload_benchmark_results_to_s3
 
 logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger(__name__)
@@ -93,6 +94,9 @@ def run_benchmark_evaluate(handler: IntegrationTestHandler, params: Dict[str, An
         handler.add_output(metric, score)
 
     params['benchmark_scores'] = scores
+
+    # Persist results to S3 before assertions so they survive assertion failures
+    upload_benchmark_results_to_s3(dataset, retriever_id)
 
     class BenchmarkEvaluateAssertions(unittest.TestCase):
         @classmethod
--- a/benchmarks/scripts/benchmark_evaluate.py
+++ b/benchmarks/scripts/benchmark_evaluate.py
@@ -96,7 +96,10 @@ def run_benchmark_evaluate(handler: IntegrationTestHandler, params: Dict[str, An
     params['benchmark_scores'] = scores
 
     # Persist results to S3 before assertions so they survive assertion failures
-    upload_benchmark_results_to_s3(dataset, retriever_id)
+    upload_benchmark_results_to_s3(
+        local_dir=results_dir,
+        s3_sub_path=f'{dataset}/{retriever_id}',
+    )
 
     class BenchmarkEvaluateAssertions(unittest.TestCase):
         @classmethod
--- a/benchmarks/utils/s3_utils.py
+++ b/benchmarks/utils/s3_utils.py
@@ -3,6 +3,9 @@
 import os
 import subprocess
 import logging
+import mimetypes
+
+import boto3
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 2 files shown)


benchmarks/scripts/benchmark_evaluate.py
```diff
@@ -94,6 +95,9 @@ def run_benchmark_evaluate(handler: IntegrationTestHandler, params: Dict[str, An
 
     params['benchmark_scores'] = scores
 
+    # Persist results to S3 before assertions so they survive assertion failures
+    upload_benchmark_results_to_s3(dataset, retriever_id)
```
noel-improv: **[Blocking]:** No try/except around `upload_file`, so a credentials or AccessDenied failure takes down an otherwise-passing run the exception lands at `test_suite.py:245-247` and `handler.succeeded()` never fires. Confirmed locally: with no creds it raises `NoCredentialsError`. Wrap and log?

benchmarks/utils/s3_utils.py
```diff
@@ -37,3 +40,113 @@ def sync_benchmark_data_from_s3(dataset: str, data_dir: str):
...
+    bucket = os.environ.get('S3_RESULTS_BUCKET')
+    if not bucket:
+        logger.warning(
+            'S3_RESULTS_BUCKET not set — skipping benchmark results upload to S3'
+        )
+        return
+
+    prefix = os.environ.get('S3_RESULTS_PREFIX', '').strip('/')
+
+    local_dir = os.path.join(results_dir, dataset, retriever_id)
```
noel-improv: **[Blocking]:** This rebuilds `benchmark-results/<dataset>/<retriever_id>` rather than using the `results_dir` already computed at `benchmark_evaluate.py:69`. With #456 in, a `pga_bio` run writes to `benchmark-results/pga_bio/` while evaluate still passes `dataset='pga'`, so this probes the wrong dir, hits the `isdir` guard and logs `nothing to upload`, a silent no-op that reads as success. Pass `results_dir` through instead?

benchmarks/utils/s3_utils.py
```diff
@@ -37,3 +40,113 @@ def sync_benchmark_data_from_s3(dataset: str, data_dir: str):
...
+    if not os.path.isdir(local_dir):
+        logger.warning(f'Results directory does not exist: {local_dir} — nothing to upload')
+        return
+
+    s3_key_base = '/'.join(
```
noel-improv, line 92: **[Question/Sugg]:** No run identifier in the key, and `S3_PREFIX` is stable per graph name, so a second run against the same graph overwrites the first. `integration_test_handler.py:179` namespaces by `test_run` maybe worth matching that here?

benchmarks/utils/s3_utils.py
```diff
@@ -37,3 +40,113 @@ def sync_benchmark_data_from_s3(dataset: str, data_dir: str):
...
+            )
+            uploaded_count += 1
+
+    s3_uri = f's3://{bucket}/{s3_key_base}/'
+    logger.info(
+        f'Upload complete: {uploaded_count} file(s) uploaded to {s3_uri}'
+    )
+
+
+def upload_all_benchmark_results_to_s3(results_dir: str = 'benchmark-results') -> None:
```
noel-improv: **[Nit]:** No callers anywhere, and the docstring points at `run_all_retrievers.sh`, which is in #455 rather than this branch. Drop it or land it with the caller that needs it?

benchmarks/utils/s3_utils.py
oussamahansal: I intentionally didn't namespace by test_run here because the use case is different — integration_test_handler tracks per-CI-run results for regression dashboards, but the benchmark results are meant to be the "latest canonical" output for a given graph+retriever combination (overwriting is the desired behavior when re-running against the same graph after a retriever fix). If you do want versioning, the cleanest option would be an optional S3_BENCHMARK_RUN_ID env var that, when set, gets inserted into the key: <prefix>/benchmark-results/<run_id>/<dataset>/<retriever>/. That keeps it opt-in and backward compatible.

benchmarks/utils/s3_utils.py
oussamahansal: Keeping it here intentionally — upload_all_benchmark_results_to_s3 depends on upload_benchmark_results_to_s3 which lives in this same file, so it makes sense to land them together.

noel-improv (approved):
lgtm

acarbonetto (approved):

Issue #458, opened 2026-08-07
[BUG] AdvancedPDFReaderProvider raises AttributeError: extract_tables removed from PDFReaderConfig


noel-improv:
### Package version

3.18.5

### Package

lexical-graph

### Python version

3.12.13

### Operating System

macOS

### Description

`AdvancedPDFReaderProvider` raises `AttributeError` on construction, so the advanced PDF reader is unusable for anyone with `pymupdf` installed.

#449 removed `extract_tables: bool = True` from `PDFReaderConfig`, but `AdvancedPDFReaderProvider` still reads `config.extract_tables` in three places, including `__init__`. Constructing the provider fails immediately.

The removal looks unintentional. #449 adds `LlamaIndexPluginReaderConfig` to the same file, nothing in the new config references `extract_tables`, and the PR description states it contains no unrelated changes.

**Why CI did not catch it**

The tests covering this provider begin with `pytest.importorskip("pymupdf")`. The workflow installs only `requirements.txt`, which does not include `pymupdf` since it is an optional dependency. All 10 tests skip and the job reports success.

The config test that does run in CI (`TestPDFReaderConfig`) asserts `return_full_document` and `metadata_fn`, but never covered `extract_tables`, so nothing upstream observed the field being deleted.

`pymupdf` is the only `importorskip`'d dependency in the suite, so this gap is limited to the PDF provider.

### Steps to reproduce

```python
# On main (f03f5703), in an environment with pymupdf installed:
from graphrag_toolkit.lexical_graph.indexing.load.readers.reader_provider_config import PDFReaderConfig
from graphrag_toolkit.lexical_graph.indexing.load.readers.providers.advanced_pdf_reader_provider import (
    AdvancedPDFReaderProvider,
)

AdvancedPDFReaderProvider(PDFReaderConfig())  # AttributeError

# Or via the existing tests, which pass on 9599d87e and fail on f03f5703:
#   pytest lexical-graph/tests/unit/indexing/load/readers/providers/test_advanced_pdf_table_extraction.py
#   -> 10 failed
```

### Error output / stack trace

```shell
tests/unit/indexing/load/readers/providers/test_advanced_pdf_table_extraction.py:68: in test_extracts_table_as_markdown_preserving_rows_and_cols
    docs = AdvancedPDFReaderProvider(PDFReaderConfig()).read(pdf)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #449, opened 2026-08-03
feat: Generic LlamaIndex Reader Plugin (clean, 4 files only)

evanerwee:
Clean PR as requested in #390 — contains only the 4 plugin files:

- `providers/__init__.py` (registration)
- `providers/llama_index_plugin_reader_provider.py` (implementation)
- `reader_provider_config.py` (config updates)
- `tests/test_llama_index_plugin_reader_provider.py` (test)

No unrelated changes. Supersedes #390.

---
**What it does:**
Allows using any LlamaIndex reader package as an extraction source via a generic plugin provider. Includes input validation, security hardening, and comprehensive tests.

PR #459, opened 2026-08-07
[FIX] Restore extract_tables on PDFReaderConfig

noel-improv:
## Description

`AdvancedPDFReaderProvider` raised `AttributeError` on construction because #449 removed `extract_tables` from `PDFReaderConfig` while the provider still reads it. This restores the field and makes the gap that hid it visible to CI.

## Changes

- Restore `extract_tables: bool = True` on `PDFReaderConfig`, in its original position
- Cover the field in the existing default and custom-value config tests, and add a field-presence test that needs no `pymupdf`
- Install `pymupdf` in the test job so the PDF provider tests run instead of skipping

## Problem

Related issue: #458

The provider reads `config.extract_tables` in three places, including `__init__`, so any user of the advanced PDF reader hit the error immediately.

CI stayed green because the provider's tests open with `pytest.importorskip("pymupdf")` and the workflow installs only `requirements.txt`, which omits the optional dependency. All ten skipped. The config test that does run asserted `return_full_document` and `metadata_fn` but never `extract_tables`, so nothing upstream observed the deletion.

`pymupdf` is the only `importorskip`'d dependency in the suite, so the second commit closes the gap rather than papering over one case.

## Testing

The 10 previously failing tests in `test_advanced_pdf_table_extraction.py` pass. Reader tests: 162 passed. Full unit suite: 1831 passed, with the one pre-existing `test_integ_dependency_compatibility` error.

I checked the new coverage fails for the right reason by deleting the field again: 3 tests fail, and they run without `pymupdf`, so CI would have caught the original removal.

Two notes for the reviewer. The 10 provider tests were verified against a clean checkout of `main` at `f03f5703` before the fix, so the failure is not local to my environment. Separately, `test_in_memory_graph_store` has two hypothesis property tests that fail intermittently on unmodified `main` — roughly one run in three here — which is unrelated to this change and worth its own issue.

## Checklist

### DIFF
--- a/.github/workflows/lexical-graph-tests.yml
+++ b/.github/workflows/lexical-graph-tests.yml
@@ -77,17 +77,9 @@ jobs:
         run: uv venv --seed .venv
 
       - name: Install dependencies
-        # pymupdf stays optional at runtime, but the PDF provider tests
-        # importorskip it: without it here they skip and the job passes green,
-        # which is how #449 removed a field they depend on unnoticed.
         run: |
           uv pip install --python .venv/bin/python \
-            pytest \
-            pytest-cov \
-            pytest-mock \
-            pytest-asyncio \
-            hypothesis \
-            pymupdf \
+            -e '.[test]' \
             -r src/graphrag_toolkit/lexical_graph/requirements.txt
 
       - name: Run tests
--- a/.github/workflows/lexical-graph-tests.yml
+++ b/.github/workflows/lexical-graph-tests.yml
@@ -77,10 +77,12 @@ jobs:
         run: uv venv --seed .venv
 
       - name: Install dependencies
+        # Separate commands: requirements.txt opens with `--only-binary :all:`,
+        # which would block building the local package if combined.
         run: |
           uv pip install --python .venv/bin/python \
-            -e '.[test]' \
             -r src/graphrag_toolkit/lexical_graph/requirements.txt
+          uv pip install --python .venv/bin/python -e '.[test]'
 
       - name: Run tests
         id: run_tests
--- a/.github/workflows/lexical-graph-tests.yml
+++ b/.github/workflows/lexical-graph-tests.yml
@@ -77,6 +77,9 @@ jobs:
         run: uv venv --seed .venv
 
       - name: Install dependencies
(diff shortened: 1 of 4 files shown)


.github/workflows/lexical-graph-tests.yml
```diff
@@ -77,13 +77,17 @@ jobs:
...
+        # importorskip it: without it here they skip and the job passes green,
+        # which is how #449 removed a field they depend on unnoticed.
         run: |
           uv pip install --python .venv/bin/python \
             pytest \
             pytest-cov \
             pytest-mock \
             pytest-asyncio \
             hypothesis \
+            pymupdf \
```
acarbonetto: please add to pyproject.toml: https://github.com/noel-improv/graphrag-toolkit/blob/17f505b0f6718ccf40e07a43a332bb5ab9ee53bf/lexical-graph/pyproject.toml#L26

acarbonetto (approved):
please update the pyproject.toml optional imports too

lexical-graph/tests/unit/indexing/load/readers/test_reader_provider_config.py
```diff
@@ -55,23 +55,31 @@ class TestPDFReaderConfig:
...
+            extract_tables=False,
             metadata_fn=custom_fn
         )
-        
+
         assert config.return_full_document is True
+        assert config.extract_tables is False
         assert config.metadata_fn == custom_fn
 
+    def test_extract_tables_field_is_present(self):
```
acarbonetto: The new test_extract_tables_field_is_present guards nothing new: the two tests updated in the same diff already raise AttributeError if the field is removed.

.github/workflows/lexical-graph-tests.yml
acarbonetto: make sure to pin the version in pyproject.toml to make CI reproduceable.

acarbonetto: consider updating this command to use the optional dependencies: 
```
    uv pip install --python .venv/bin/python \
      -e .[test] \
      -r src/graphrag_toolkit/lexical_graph/requirements.txt
```

PR #465, opened 2026-08-11
[FEATURE] Add S3-backed chunk store behind the ChunkStore interface

noel-improv:
## Description

Adds `S3ChunkStore`, the first non-graph backend for the `ChunkStore` interface from #436/#437. Chunk text moves to S3 and off the `__Chunk__` node, so graph memory stops scaling with text volume.

Opt in with `GraphRAGConfig.chunk_store` or the `CHUNK_STORE` environment variable:

```
s3://bucket/prefix?kmsKeyArn=...
```

Leaving it unset keeps text on the graph node, so nothing changes for anyone who doesn't configure it.

**This PR stacks on #436 and #437** and shows their 7 commits. Its own change is the last two, across 14 files.

**Two design points worth reading before the diff**, because each one is visible as an absence rather than as code:

- Nothing is written to the graph node. The S3 key is derived from the chunk id (`{prefix}/{chunk_id}.txt`), so a read already knows every key it needs. There is no reference property to store, index, or leave unindexed.
- Nothing lists the bucket. #420 parallelises key *discovery* in `S3ChunkDownloader`, which is extraction-phase document storage rather than chunk-text storage despite the similar name. Reads here are concurrent GETs in `get_batch`, sized from `extraction_num_threads_per_worker`.

`S3ChunkStore` also takes an optional `fallback` store, wired to `InGraphChunkStore` by the factory. Chunks S3 doesn't hold are read from the graph, so text written before you set `chunk_store` stays readable and existing graphs need no migration step. Errors other than a missing key propagate rather than falling back, so an expired-credentials or access-denied problem doesn't quietly read as an empty bucket.

## Changes

- `storage/chunk/s3_chunk_store.py` — new backend: deterministic key, concurrent batched reads, `AES256` or `aws:kms` server-side encryption, dual-read fallback
- `storage/chunk_store_factory.py` — `S3ChunkStoreFactory` for `s3://` connection strings, mirroring `S3VectorIndexFactory`; registry seeded at import
- `config.py` — `chunk_store` property, env-backed via `CHUNK_STORE`, defaulting to `None`
- The three call sites now pass `GraphRAGConfig.chunk_store` through to the factory
- `configuration.mdx` — parameter table entry and a section on chunk text storage

## Problem

Chunk text stored as `chunk.value` dominates graph memory on a large corpus, and every traversal touching a chunk node carries it.

Measured on a 500-chunk collection: graph-resident chunk text went from 1000 KiB to zero, with a query path identical to the in-graph baseline.

Two things I'd rather state than have found in review:

- **Writes are serial.** `ChunkGraphBuilder` calls `put()` once per chunk with no batching. Measured over 5000 chunks that's 96.6 ms/chunk against 8.0 ms/chunk in-graph. Batched reads at raised concurrency run 2.14 ms/chunk. There's a `TODO` on `put()` recording this; a `put_batch` on the interface is the follow-up, and it belongs with the interface rather than here.
- **Read concurrency depends on #420.** `get_batch` fans out to `extraction_num_threads_per_worker` GETs, and botocore's connection pool defaults to 10. #420 sizes the pool to the thread count and is already in `main`; this branch was cut before that merge and picks it up on rebase. Benchmarking these reads without it measures the pool, not the store.

This lands at 14 files against the usual leanness guideline of five. That was a deliberate call to keep the backend, its wiring, and its tests reviewable as one unit rather than split across three PRs that are meaningless apart.

Related issue (if any): #324

## Testing

Unit tests cover the round trip, batch reads, missing keys omitted rather than raised, the fallback consulted only for what S3 lacked, empty chunk text treated as a hit rather than a miss, large payloads, `AES256` vs `aws:kms` selection, and separate cases for expired credentials, throttling past botocore's retries, and access denied not being swallowed by the fallback.

`tests/integration/storage/chunk/test_s3_chunk_store_live.py` runs a live round trip and a dual-read case against real S3, gated on an environment variable.

122 tests pass across the chunk store, factory, builder, retriever, and config files. The wider unit run shows failures in `version_manager`, `vector_batch_client`, `vector_indexing`, and the fact/statement/topic builders; those reproduce identically on a clean `main` checkout and are unrelated to this change.

Manual: live runs against S3 in `us-west-2` with Neo4j 5, covering the memory result above, read latency across chunk sizes from 512 B to 128 KiB, and the mixed-residency dual-read case. Reads get slower as chunks migrate into S3, not faster, because a miss returns a small 404 while a hit downloads a body — so there's no urgency to backfill existing graphs.

## Checklist

---

### DIFF
--- a/docs-site/src/content/docs/lexical-graph/configuration.mdx
+++ b/docs-site/src/content/docs/lexical-graph/configuration.mdx
@@ -70,6 +70,12 @@ The following parameter applies only when using Amazon OpenSearch Serverless as 
 | ------------- | ------------- | ------------- | ------------- |
 | `opensearch_engine` | OpenSearch kNN engine | `nmslib` | `OPENSEARCH_ENGINE` |
 
+The following parameter controls where chunk text is stored:
+
+| Parameter | Description | Default | Environment Variable |
+| ------------- | ------------- | ------------- | ------------- |
+| `chunk_store` | Connection string for the store that holds chunk text (see [Chunk text storage](#chunk-text-storage)) | *None* | `CHUNK_STORE` |
+
 The following parameters configure local filesystem paths for container/EKS deployments:
 
 | Parameter | Description | Default | Environment Variable |
@@ -203,6 +209,26 @@ The lexical-graph uses microbatching to progress source data through the extract
   - In the build stage, chunks from the extract stage are broken down into smaller *indexable* nodes representing sources, chunks, topics, statements and facts. These indexable nodes are then processed by the graph construction and vector indexing handlers.
 
 The `batch_writes_enabled` configuration parameter determines whether all of the indexable nodes derived from a batch of incoming chunks are written to the graph and vector stores singly, or as a bulk operation. Bulk/batch operations tend to improve the throughput of the build stage, at the expense of some additonal latency with regard to this data becoming available to query.
+
+#### Chunk text storage
+
+By default, chunk text is stored as a `value` property on each `__Chunk__` node in the graph. That keeps everything in one place, but on a large corpus the text dominates graph memory, and every traversal that touches a chunk node carries it.
+
+Set `chunk_store` to move chunk text to S3 instead:
+
+```python
+GraphRAGConfig.chunk_store = 's3://my-bucket/chunks'
+```
+
+The connection string takes an optional prefix and an optional `kmsKeyArn` query parameter. Objects are server-side encrypted with `AES256` by default, or with `aws:kms` when a key is supplied:
+
+```python
+GraphRAGConfig.chunk_store = 's3://my-bucket/chunks?kmsKeyArn=arn:aws:kms:us-east-1:111122223333:key/1234abcd'
+```
+
+Each chunk is written to a key derived from its chunk id, so reads never list the bucket. Reads that don't find a chunk in S3 fall back to the graph, which means text written before you set `chunk_store` is still returned. Existing graphs keep working with no migration step; chunks written from that point on go to S3.
+
+Leave `chunk_store` unset to keep chunk text on the graph node.
 
 #### Caching Amazon Bedrock LLM responses
 
--- a/docs-site/src/content/docs/lexical-graph/configuration.mdx
+++ b/docs-site/src/content/docs/lexical-graph/configuration.mdx
@@ -212,7 +212,7 @@ The `batch_writes_enabled` configuration parameter determines whether all of the
 
 #### Chunk text storage
 
-By default, chunk text is stored as a `value` property on each `__Chunk__` node in the graph. That keeps everything in one place, but on a large corpus the text dominates graph memory, and every traversal that touches a chunk node carries it.
+By default, chunk text is stored as a `value` property on each `__Chunk__` node in the graph. That keeps everything in one place, but for more optimal storage, use the chunk_store option to store the text in an external source.
 
 Set `chunk_store` to move chunk text to S3 instead:
 
@@ -220,7 +220,7 @@ Set `chunk_store` to move chunk text to S3 instead:
 GraphRAGConfig.chunk_store = 's3://my-bucket/chunks'
 ```
 
-The connection string takes an optional prefix and an optional `kmsKeyArn` query parameter. Objects are server-side encrypted with `AES256` by default, or with `aws:kms` when a key is supplied:
+The connection string takes an optional prefix (`s3://my-bucket` and `s3://my-bucket/chunks` are both valid; with a prefix, objects are written under `chunks/`), and a `?` separates the optional `kmsKeyArn` query parameter. Objects are server-side encrypted with `AES256` by default, or with `aws:kms` when a key is supplied:
 
 ```python
 GraphRAGConfig.chunk_store = 's3://my-bucket/chunks?kmsKeyArn=arn:aws:kms:us-east-1:111122223333:key/1234abcd'
--- a/docs-site/src/content/docs/lexical-graph/configuration.mdx
+++ b/docs-site/src/content/docs/lexical-graph/configuration.mdx
@@ -74,7 +74,7 @@ The following parameter controls where chunk text is stored:
 
 | Parameter | Description | Default | Environment Variable |
 | ------------- | ------------- | ------------- | ------------- |
-| `chunk_store` | Connection string for the store that holds chunk text (see [Chunk text storage](#chunk-text-storage)) | *None* | `CHUNK_STORE` |
+| `s3_chunk_store` | S3 connection string for the store that holds chunk text (see [Chunk text storage](#chunk-text-storage)) | *None* | `S3_CHUNK_STORE` |
 
 The following parameters configure local filesystem paths for container/EKS deployments:
 
@@ -212,23 +212,23 @@ The `batch_writes_enabled` configuration parameter determines whether all of the
 
 #### Chunk text storage
 
-By default, chunk text is stored as a `value` property on each `__Chunk__` node in the graph. That keeps everything in one place, but for more optimal storage, use the chunk_store option to store the text in an external source.
+By default, chunk text is stored as a `value` property on each `__Chunk__` node in the graph. That keeps everything in one place, but for more optimal storage, use the `s3_chunk_store` option to store the text in an external source.
 
-Set `chunk_store` to move chunk text to S3 instead:
-
-```python
-GraphRAGConfig.chunk_store = 's3://my-bucket/chunks'
-```
-
-The connection string takes an optional prefix (`s3://my-bucket` and `s3://my-bucket/chunks` are both valid; with a prefix, objects are written under `chunks/`), and a `?` separates the optional `kmsKeyArn` query parameter. Objects are server-side encrypted with `AES256` by default, or with `aws:kms` when a key is supplied:
+Set `s3_chunk_store` to move chunk text to S3 instead:
 
 ```python
-GraphRAGConfig.chunk_store = 's3://my-bucket/chunks?kmsKeyArn=arn:aws:kms:us-east-1:111122223333:key/1234abcd'
+GraphRAGConfig.s3_chunk_store = 's3://my-bucket/chunks'
 ```
 
-Each chunk is written to a key derived from its chunk id, so reads never list the bucket. Reads that don't find a chunk in S3 fall back to the graph, which means text written before you set `chunk_store` is still returned. Existing graphs keep working with no migration step; chunks written from that point on go to S3.
+The connection string takes an optional prefix (`s3://my-bucket` and `s3://my-bucket/chunks` are both valid; with a prefix, objects are written under `chunks/`), and a `?` separates the optional `kmsKeyArn` query parameter. Objects are encrypted with `aws:kms` when a key is supplied:
 
-Leave `chunk_store` unset to keep chunk text on the graph node.
+```python
+GraphRAGConfig.s3_chunk_store = 's3://my-bucket/chunks?kmsKeyArn=arn:aws:kms:us-east-1:111122223333:key/1234abcd'
+```
+
+Each chunk is written to a key derived from its chunk id, so reads never list the bucket. Reads that don't find a chunk in S3 fall back to the graph, which means text written before you set `s3_chunk_store` is still returned. Existing graphs keep working with no migration step; chunks written from that point on go to S3.
+
+Leave `s3_chunk_store` unset to keep chunk text on the graph node.
 
 #### Caching Amazon Bedrock LLM responses
 
--- a/lexical-graph/src/graphrag_toolkit/lexical_graph/config.py
+++ b/lexical-graph/src/graphrag_toolkit/lexical_graph/config.py
@@ -348,7 +348,7 @@ class _GraphRAGConfig:
     _opensearch_serverless_nextgen_compression: Optional[str] = None
     _enable_versioning = None
     _chunk_external_properties: Optional[Dict[str, str]] = None
-    _chunk_store: Optional[str] = None
+    _s3_chunk_store: Optional[str] = None
     _local_output_dir: Optional[str] = None
     _log_output_dir: Optional[str] = None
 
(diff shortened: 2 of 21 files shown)


docs-site/src/content/docs/lexical-graph/configuration.mdx
```diff
@@ -204,6 +210,26 @@ The lexical-graph uses microbatching to progress source data through the extract
...
+
+By default, chunk text is stored as a `value` property on each `__Chunk__` node in the graph. That keeps everything in one place, but on a large corpus the text dominates graph memory, and every traversal that touches a chunk node carries it.
+
+Set `chunk_store` to move chunk text to S3 instead:
+
+```python
+GraphRAGConfig.chunk_store = 's3://my-bucket/chunks'
+```
+
+The connection string takes an optional prefix and an optional `kmsKeyArn` query parameter. Objects are server-side encrypted with `AES256` by default, or with `aws:kms` when a key is supplied:
```
acarbonetto: Provide an example with "an optional prefix".  It is optional?

acarbonetto: Mention that parameters are separated with a "?"

docs-site/src/content/docs/lexical-graph/configuration.mdx
```diff
@@ -204,6 +210,26 @@ The lexical-graph uses microbatching to progress source data through the extract
...
+
+```python
+GraphRAGConfig.chunk_store = 's3://my-bucket/chunks?kmsKeyArn=arn:aws:kms:us-east-1:111122223333:key/1234abcd'
+```
+
+Each chunk is written to a key derived from its chunk id, so reads never list the bucket. Reads that don't find a chunk in S3 fall back to the graph, which means text written before you set `chunk_store` is still returned. Existing graphs keep working with no migration step; chunks written from that point on go to S3.
```
acarbonetto: By default, we should read from the graph (since S3 lookup is slower).  If the chunk is not specified, then it should look up the chuck_store reference (S3 in this case).

docs-site/src/content/docs/lexical-graph/configuration.mdx
```diff
@@ -70,6 +70,12 @@ The following parameter applies only when using Amazon OpenSearch Serverless as
 | ------------- | ------------- | ------------- | ------------- |
 | `opensearch_engine` | OpenSearch kNN engine | `nmslib` | `OPENSEARCH_ENGINE` |
 
+The following parameter controls where chunk text is stored:
+
+| Parameter | Description | Default | Environment Variable |
+| ------------- | ------------- | ------------- | ------------- |
+| `chunk_store` | Connection string for the store that holds chunk text (see [Chunk text storage](#chunk-text-storage)) | *None* | `CHUNK_STORE` |
```
acarbonetto: does this scale to _other_ locations?  What if we want to use something like OpenSearch as the chunk_store, how do you envision that would work?

docs-site/src/content/docs/lexical-graph/configuration.mdx
```diff
@@ -204,6 +210,26 @@ The lexical-graph uses microbatching to progress source data through the extract
 
 The `batch_writes_enabled` configuration parameter determines whether all of the indexable nodes derived from a batch of incoming chunks are written to the graph and vector stores singly, or as a bulk operation. Bulk/batch operations tend to improve the throughput of the build stage, at the expense of some additonal latency with regard to this data becoming available to query.
 
+#### Chunk text storage
+
+By default, chunk text is stored as a `value` property on each `__Chunk__` node in the graph. That keeps everything in one place, but on a large corpus the text dominates graph memory, and every traversal that touches a chunk node carries it.
```
acarbonetto:
```suggestion
By default, chunk text is stored as a `value` property on each `__Chunk__` node in the graph. That keeps everything in one place, but for more optimal storage, use the chunk_store option to store the text in an external source.
```

docs-site/src/content/docs/lexical-graph/configuration.mdx
acarbonetto: > Objects are server-side encrypted with `AES256` by default

Is this defined on the S3 side?

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/build/chunk_graph_builder.py
```diff
@@ -57,33 +59,35 @@ def build(self, node:BaseNode, graph_client: GraphStore, **kwargs:Any):
 
             logger.debug(f'Inserting chunk [chunk_id: {chunk_id}]')
 
-            statements_c = [
-                '// insert chunks',
-                'UNWIND $params AS params',
-                f'MERGE (chunk:`__Chunk__`{{{graph_client.node_id("chunkId")}: params.chunk_id}})'
-            ]
+            chunk_store = ChunkStoreFactory.for_chunk_store(GraphRAGConfig.chunk_store, graph_store=graph_client)
+            chunk_store.put(chunk_id, node.text)
```
acarbonetto: During batch indexing, chunk_store.put() writes to S3 immediately, while the __Chunk__ node, __EXTRACTED_FROM__ edge, and metadata queries are only queued and flushed later. If apply_batch_operations() fails for that batch (or the process dies after the put), S3 holds orphaned chunk text with no corresponding graph node/edges, and the write never honored the max_attempts=5 retry passed to put(). The in-graph backend can't hit this — its put() rides the same flush.

lexical-graph/src/graphrag_toolkit/lexical_graph/storage/chunk/chunk_store.py
```diff
@@ -0,0 +1,40 @@
...
+    (the graph itself, S3, or any other object store).
+    """
+
+    def get(self, chunk_id: str) -> Optional[str]:
+        """
+        Return the text for a single chunk, or None if it isn't found.
+        """
+        return self.get_batch([chunk_id]).get(chunk_id)
+
+    # TODO: add a put_batch to this interface. ChunkGraphBuilder calls put()
```
acarbonetto: I think we should implement this TODO now - since we want to be consistent with the graph calls. Also, this would help us with our batch benchmarking.

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/build/chunk_graph_builder.py
```diff
@@ -57,33 +59,35 @@ def build(self, node:BaseNode, graph_client: GraphStore, **kwargs:Any):
 
             logger.debug(f'Inserting chunk [chunk_id: {chunk_id}]')
 
-            statements_c = [
-                '// insert chunks',
-                'UNWIND $params AS params',
-                f'MERGE (chunk:`__Chunk__`{{{graph_client.node_id("chunkId")}: params.chunk_id}})'
-            ]
+            chunk_store = ChunkStoreFactory.for_chunk_store(GraphRAGConfig.chunk_store, graph_store=graph_client)
```
acarbonetto: Builds a fresh S3ChunkStore (+ in-graph fallback) for every node. Other consumers (KeywordVSSProvider, TraversalBasedBaseRetriever) resolve the store once at construction. The builder should too.

lexical-graph/src/graphrag_toolkit/lexical_graph/storage/chunk/s3_chunk_store.py
```diff
@@ -0,0 +1,105 @@
...
+        return response['Body'].read().decode('UTF-8')
+
+    def get_batch(self, chunk_ids: List[str]) -> Dict[str, str]:
+        if not chunk_ids:
+            return {}
+
+        s3_client = GraphRAGConfig.s3
+        found = {}
+
+        with concurrent.futures.ThreadPoolExecutor(max_workers=GraphRAGConfig.extraction_num_threads_per_worker) as executor:
```
acarbonetto: get_batch runs on the retrieval read path but sizes its pool by extraction_num_threads_per_worker — an extraction-time config with no relation to retrieval concurrency. Plus pool create/teardown on every call.

acarbonetto (changes requested):
to consider:

lexical-graph/src/graphrag_toolkit/lexical_graph/storage/chunk_store_factory.py
```diff
@@ -0,0 +1,137 @@
...
+class S3ChunkStoreFactory(ChunkStoreFactoryMethod):
+    """
+    Creates an S3ChunkStore from an `s3://bucket/prefix` connection string,
+    optionally with a `?kmsKeyArn=` query parameter.
+
+    The store is given an InGraphChunkStore fallback when a graph_store is
+    available, so chunk text written before a migration - still inline on the
+    graph node - is still readable.
+    """
+    def try_create(self, chunk_info: str, **kwargs) -> Optional[ChunkStore]:
```
oussamahansal: `s3:///prefix` (no bucket) passes `startswith('s3://')`, `parse_s3_connection_string` returns `bucket_name=None`, and we build a broken `S3ChunkStore(None)`. it  will only trigger on a malformed connection string we should probably validate the bucket

lexical-graph/src/graphrag_toolkit/lexical_graph/storage/chunk_store_factory.py
```diff
@@ -0,0 +1,137 @@
...
+
+    prefix = parsed.path[1:] if parsed.path else None
+    if prefix:
+        while prefix.endswith('/'):
+            prefix = prefix[:-1]
+    prefix = prefix if prefix else None
+
+    kms_key_arn = parse_qs(parsed.query).get('kmsKeyArn', None) if parsed.query else None
+    if kms_key_arn:
+        kms_key_arn = kms_key_arn if not isinstance(kms_key_arn, list) else kms_key_arn[0]
```
oussamahansal: `parse_qs().get('kmsKeyArn')` always returns a list, so the `not isinstance(..., list)` branch is dead. Just take `[0]`

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/build/graph_batch_client.py
```diff
@@ -77,7 +77,21 @@ def node_id(self, id_name:str):
             Any: The ID of the node as returned by the `graph_client`.
         """
         return self.graph_client.node_id(id_name)
-    
+
+    def execute_query(self, query, parameters={}, **kwargs):
```
oussamahansal: `build()` only calls `put()`, and read paths hold the real GraphStore so nothing calls `InGraphChunkStore.get_batch() `with a batch client. Is this reachable?

oussamahansal (changes requested):

lexical-graph/src/graphrag_toolkit/lexical_graph/storage/chunk/chunk_store.py
noel-improv: Implemented in `cbaba46a` and wired in `77dc9546`

oussamahansal (approved):
LGTM

docs-site/src/content/docs/lexical-graph/configuration.mdx
noel-improv: Half agree. No stored reference exists to look up, keys derive from the chunk id. Graph-first wins during migration but costs a wasted query on every read forever after, and migration measured as the cheap end. Keeping S3-first with the fallback; we can revisit when a migration-heavy deployment shows up

docs-site/src/content/docs/lexical-graph/configuration.mdx
noel-improv: Done

docs-site/src/content/docs/lexical-graph/configuration.mdx
```diff
@@ -204,6 +210,26 @@ The lexical-graph uses microbatching to progress source data through the extract
...
+
+Set `chunk_store` to move chunk text to S3 instead:
```
acarbonetto: we should name this variable appropriately (it should have "S3" or "S3_uri" or "S3_arn" as part of the name of the variable.  For example: `chunk_store_s3_arn` to explicitly mention the type.

lexical-graph/src/graphrag_toolkit/lexical_graph/storage/chunk_store_factory.py
```diff
@@ -3,15 +3,46 @@
...
 
 logger = logging.getLogger(__name__)
 
 ChunkStoreType = Union[str, ChunkStore]
 ChunkStoreFactoryMethodType = Union[ChunkStoreFactoryMethod, Type[ChunkStoreFactoryMethod]]
 
+S3 = 's3://'
+
+
+def parse_s3_connection_string(connection_string):
```
acarbonetto: this should be moved to the s3_chunk_store since it's only relevant to s3 chunk storage.

lexical-graph/src/graphrag_toolkit/lexical_graph/storage/chunk_store_factory.py
```diff
@@ -3,15 +3,46 @@
...
 from graphrag_toolkit.lexical_graph.storage.chunk import ChunkStore, ChunkStoreFactoryMethod
 from graphrag_toolkit.lexical_graph.storage.chunk.in_graph_chunk_store import InGraphChunkStore
+from graphrag_toolkit.lexical_graph.storage.chunk.s3_chunk_store import S3ChunkStore
 
 logger = logging.getLogger(__name__)
 
 ChunkStoreType = Union[str, ChunkStore]
 ChunkStoreFactoryMethodType = Union[ChunkStoreFactoryMethod, Type[ChunkStoreFactoryMethod]]
 
+S3 = 's3://'
```
acarbonetto: move to s3_chunk_store

lexical-graph/src/graphrag_toolkit/lexical_graph/storage/chunk_store_factory.py
```diff
@@ -31,7 +37,39 @@ def try_create(self, chunk_info: str, **kwargs) -> Optional[ChunkStore]:
         return InGraphChunkStore(graph_store)
 
 
-_chunk_store_factories: Dict[str, ChunkStoreFactoryMethod] = {}
+class S3ChunkStoreFactory(ChunkStoreFactoryMethod):
```
acarbonetto: are we going to include all the chunk store factories in the same file, or should be split this into separate files (like we do for the chunk stores themselves). e.g. s3_chunk_store_factory.py

lexical-graph/src/graphrag_toolkit/lexical_graph/config.py
```diff
@@ -1216,6 +1217,25 @@ def bedrock_reranking_model(self) -> str:
     def bedrock_reranking_model(self, bedrock_reranking_model: str) -> None:
         self._bedrock_reranking_model = bedrock_reranking_model
 
+    @property
+    def chunk_store(self) -> Optional[str]:
+        """
+        Connection string for the store that holds chunk text, or None to keep
+        it on the graph node as `chunk.value`.
+
+        Set `s3://bucket/prefix` (optionally `?kmsKeyArn=...`) to move chunk
```
acarbonetto, line 1226: this is only relevant for s3 chunk stores, and not a general parameter.  If we do want to keep this here, we must include a 'type' in the name, for example `chunk_store_url` or `chunk_store_endpoint`, etc that is generalized for all storage. 
I'd prefer to move this to the S3 chunk storage as a parameter for S3 chunk store specifically.

acarbonetto (approved):

PR #466, opened 2026-08-13
[FIX] Stop hypothesis generating property keys that shadow the node id

### LINKED ISSUE #460: [BUG] In-memory graph store property tests overwrite the node id with a generated property
### Package version

3.18.5

### Package

lexical-graph

### Python version

3.12.13

### Operating System

macOS

### Description

Four property tests in `tests/unit/storage/graph/test_in_memory_graph_store.py` build their query parameters so that a generated property can overwrite the node id:

```python
parameters = {
    'id': node['id'],
    'label': node['label'],
    **node['properties']
}
```

The `properties` strategy generates arbitrary keys, so it can produce one named `id`. When it does, the splat overwrites `parameters['id']`, the `MERGE` writes a node under the property's value, and the assertions then look up the original `node['id']` and find nothing.

Falsifying example from a local run:

```
node={'id': '0', 'label': 'Document', 'properties': {'id': '1'}}
-> Expected 1 node after 5 additions, got 0
```

The store behaves correctly here. The test hands it one id and then asserts on a different one.

Affected sites, all in `tests/unit/storage/graph/test_in_memory_graph_store.py`:

- line 118, in `test_add_node_idempotence_property` (line 93)
- line 168, in `test_add_node_multiple_times_idempotence_property` (line 146)
- lines 266 and 275, in `test_add_node_with_updated_properties_idempotence_property` (line 244)

**Why this surfaces intermittently**

Hypothesis caches counterexamples in `lexical-graph/.hypothesis/`. Before it finds this input the tests pass; afterwards they fail on every run from the cached example. That makes it look flaky when it is deterministic on either side of the discovery.

Measured locally on unmodified `main`:

| Condition | Result |
| --- | --- |
| With the cached counterexample | 10 of 10 runs fail |
| After `rm -rf lexical-graph/.hypothesis` | 5 of 5 runs pass |

CI starts each run without the database and the `ci` profile caps `max_examples` at 20, so it usually does not reach the input. It will fail whenever a runner happens to generate it.

A fix would keep generated properties from colliding with the reserved keys, either by filtering `id` and `label` out of the strategy or by passing properties as a nested dict rather than splatting them.

### Steps to reproduce

```python
# From lexical-graph/, with the example database cleared:
#   rm -rf .hypothesis
#   HYPOTHESIS_PROFILE=ci PYTHONPATH=src pytest \
#     tests/unit/storage/graph/test_in_memory_graph_store.py
#
# Repeat until hypothesis finds the input; afterwards it fails on every run.
# Or force it directly:

node = {'id': '0', 'label': 'Document', 'properties': {'id': '1'}}

parameters = {
    'id': node['id'],
    'label': node['label'],
    **node['properties'],
}

assert parameters['id'] == node['id']  # fails: 'id' is now '1'
```

```shell
tests/unit/storage/graph/test_in_memory_graph_store.py:176: in test_add_node_multiple_times_idempotence_property
    assert stored_node is not None, "Node not found after additions"
E   AssertionError: Node not found after additions
E   assert None is not None
E   Falsifying example: test_add_node_multiple_times_idempotence_property(
E       node={'id': '0', 'label': 'Document', 'properties': {'id': '1'}},
E   )

tests/unit/storage/graph/test_in_memory_graph_store.py:176: in test_add_node_multiple_times_idempotence_property
    assert store.get_node_count() == 1
E   AssertionError: Expected 1 node after 5 additions, got 0
E   Falsifying example: test_add_node_multiple_times_idempotence_property(
E       node={'id': '0', 'label': 'Document', 'properties': {'id': ''}},
E   )
```

noel-improv:
## Description

The in-memory graph store property tests can generate a property key named `id`, which shadows the node id and makes the test assert on a node that was never written.

## Changes

- Filter `id`, `node_id` and `label` out of the `properties` key strategy in `test_in_memory_graph_store.py`

## Problem

Four tests build their query parameters as `{'id': ..., 'label': ..., **node['properties']}`. The properties strategy generates arbitrary keys, so it can produce one named `id`. The splat then overwrites `parameters['id']`, `MERGE` writes the node under the property's value, and the assertions look up the original id and find nothing. The store is correct; the test hands it one id and asserts on another.

This looks flaky but is not. Hypothesis caches the counterexample, so the tests pass until it finds the input and fail on every run afterwards. CI starts without the database and the `ci` profile caps `max_examples` at 20, so it usually does not reach the input.

Fixing it in the strategy covers all four call sites in one line.

Related issue (if any): Fixes #460

## Testing

`pytest tests/unit/storage/graph/test_in_memory_graph_store.py` passes with the cached counterexample present, and on three further runs after `rm -rf .hypothesis`.

## Checklist

### DIFF
--- a/lexical-graph/tests/unit/storage/graph/test_in_memory_graph_store.py
+++ b/lexical-graph/tests/unit/storage/graph/test_in_memory_graph_store.py
@@ -64,7 +64,7 @@ def node_parameters(node, **extra_properties):
     }
 
 
-def merge_query(label):
+def node_merge_query(label):
     """
     Build the MERGE query for a node with the given label.
 
@@ -158,12 +158,12 @@ def test_add_node_idempotence_property(node):
     parameters = node_parameters(node)
 
     # Add node first time
-    store.execute_query(merge_query(node['label']), parameters)
+    store.execute_query(node_merge_query(node['label']), parameters)
     count_after_first = store.get_node_count()
     node_after_first = store.get_node(node['id'])
     
     # Add same node second time (should be idempotent)
-    store.execute_query(merge_query(node['label']), parameters)
+    store.execute_query(node_merge_query(node['label']), parameters)
     count_after_second = store.get_node_count()
     node_after_second = store.get_node(node['id'])
     
@@ -199,7 +199,7 @@ def test_add_node_multiple_times_idempotence_property(node):
 
     # Add node multiple times
     for _ in range(5):
-        store.execute_query(merge_query(node['label']), parameters)
+        store.execute_query(node_merge_query(node['label']), parameters)
     
     # Property: Should have exactly one node
     assert store.get_node_count() == 1, \
@@ -235,10 +235,10 @@ def test_add_different_nodes_not_idempotent_property(node1, node2):
     store = InMemoryGraphStoreForTesting()
     
     # Add first node
-    store.execute_query(merge_query(node1['label']), node_parameters(node1))
+    store.execute_query(node_merge_query(node1['label']), node_parameters(node1))
 
     # Add second node
-    store.execute_query(merge_query(node2['label']), node_parameters(node2))
+    store.execute_query(node_merge_query(node2['label']), node_parameters(node2))
     
     # Property: Should have two distinct nodes
     assert store.get_node_count() == 2, \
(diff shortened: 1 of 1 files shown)


noel-improv:
Pushed the restructure. Parameters are now `{id, label, properties}` so `SET n += $properties` binds to something, which removes the collision at its root and lets the filter go.

One deliberate difference from your suggestion: the label stays in the query text rather than becoming `$label`. openCypher cannot parameterize a label, so `MERGE (n:$label ...)` would not run against Neptune or Neo4j, and every MERGE in `src/` writes a literal label and binds only values. There is a comment on `merge_query()` saying so. Happy to switch if you disagree.

I left the `**Validates: Requirements 5.6**` docstrings alone. No such document exists in the repo, and it is wider than this file: 7 test files cite 13 different requirement numbers. Looks like a separate cleanup to me rather than part of this fix.

lexical-graph/tests/unit/storage/graph/test_in_memory_graph_store.py
```diff
@@ -36,6 +36,53 @@
...
+    }
+
+    return {
+        'id': node['id'],
+        'label': node['label'],
+        'properties': properties
+    }
+
+
+def merge_query(label):
```
acarbonetto: nit: 
```suggestion
def node_merge_query(label):
```

acarbonetto (approved):

PR #451, opened 2026-08-05
fix: stop FMObservabilityPublisher.close() from leaking its poller thread

AmirF194:
## Description

`FMObservabilityPublisher.close()` set `allow_continue = False` but never touched the poller, so a poller thread survived every `close()` call.

## Problem

`publish_stats()` unconditionally stops the current poller and starts a replacement (lines 581-583 before the fix) and only checks `allow_continue` afterward to decide whether to reschedule itself. The `threading.Timer` scheduled in `__init__`, or an in-flight tick, can still fire after `close()` sets `allow_continue = False`, so the replacement poller it creates is started but never stopped and keeps running for the rest of the process.

The same gap turns two of the file's own tests into a busy loop: `test_publisher_initialization_sets_up_handlers` and `test_publisher_context_manager` mock `Queue` with a plain `Mock`, so `Mock.get()` returns immediately instead of raising `queue.Empty`, and the leaked poller spins instead of idling on a real 1-second block.

Related issue: #450

## Changes

- `close()` now stops the current poller synchronously (under a new `threading.Lock` shared with `publish_stats()`).
- `publish_stats()` only creates and starts a replacement poller while `allow_continue` is still true, checked under the same lock, so a tick that fires after `close()` cannot leave a fresh poller running.
- `test_publisher_initialization_sets_up_handlers` / `test_publisher_context_manager`: gave the mocked queue `get.side_effect = queue.Empty`, matching the `mock_fm_queue` fixture already used elsewhere in the file, so the poller polls instead of spinning.
- Added `test_close_stops_the_current_poller` and `test_publish_stats_does_not_replace_poller_after_close`, which fail on unmodified `main` and pass with the fix.

## Testing

- [x] Unit tests added/updated
- [ ] Integration tests added (as appropriate)
- [x] Existing tests pass (`pytest`)
- [ ] Tested manually (describe below)

New tests fail on `main` (`assert False` on `_discontinue.is_set()`; poller identity check fails because `publish_stats()` always replaces it) and pass on this branch. Full `lexical-graph` suite: 1793 passed, 1 skipped, coverage 61.98% (repo gate is 56%), run in a clean `python:3.12-slim` container with the CI-matching `pytest --cov-config=.coveragerc --cov=graphrag_toolkit.lexical_graph --cov-fail-under=56` invocation. `bandit -r lexical-graph/src/ -ll`: no issues. Not run: the `3.10`/`3.11` legs of the CI matrix (only checked `3.12` locally).

## Checklist

[...]

PR #467, opened 2026-08-13
[FIX] Stop FMObservabilityPublisher leaking its poller thread

### LINKED ISSUE #450: [BUG] FMObservabilityPublisher.close() leaks its poller thread; unit tests busy-loop it on a mocked queue
### Package version

3.18.5

### Package

lexical-graph

### Python version

3.12.13

### Operating System

macOS (macOS-26.5.2-arm64)

### Description

`FMObservabilityPublisher.close()` does not stop its poller thread. It sets `allow_continue = False` and returns (`fm_observability.py:535-547`). The `threading.Timer` scheduled in `__init__` then fires `publish_stats()`, which stops the current poller and starts a replacement at lines 581-583, before the `allow_continue` check at line 584. Every publisher therefore leaves a live `FMObservabilityQueuePoller` behind, and calling `close()` does not change that.

I confirmed this with the real `multiprocessing.Queue` in place, patching only `Settings`:

```
threads still alive after close(): [('Thread-3', True)]
publisher's current poller alive: True
poller._discontinue set: False
```

With a real queue the leaked thread is idle, because `Queue.get(timeout=1)` blocks for a second per iteration. Under the unit tests it is not idle. `tests/unit/utils/test_fm_observability.py` patches `fm_observability.Queue` with a plain `Mock` (lines 443 and 464), so `__init__` assigns a Mock to the module global `_fm_observability_queue` at line 520. The poller loop catches only `queue.Empty` (lines 70-75), and `Mock.get()` returns a truthy child Mock immediately rather than raising, so the one-second poll becomes a busy loop. Every iteration is recorded in the Mock's `mock_calls`, which nothing trims.

Measured with the script below: roughly 410 MB/s and 175k retained calls per second, unbounded, for the rest of the process lifetime. Two tests start such a publisher, `test_publisher_initialization_sets_up_handlers` and `test_publisher_context_manager`.

What this does to a full `pytest lexical-graph/tests` run, measured with a small plugin that samples RSS at each test teardown:

- Every test that runs after `test_fm_observability.py` grows RSS by roughly 150 MB regardless of what the test does. `test_io_utils.py` grows 2 MB when run on its own and 4 GB when it runs after the observability tests.
- Peak RSS for the unit suite was 10.3 GB with coverage off.
- Shutdown is slow, because the GC traverses the retained call list. In the bounded repro below, 20 s of leaking produced 8.4 GB and added about 24 s of interpreter shutdown to a program whose body took 22 s.

Stopping the surviving poller pins RSS flat at +0.0 MB, which is how I confirmed nothing else contributes.

Suggested fix on the production side: `close()` stops the poller, and `publish_stats()` starts a replacement only when `allow_continue` is true. On the test side, the mocked queue needs `get.side_effect = queue.Empty` so the poller polls rather than spins.

### Steps to reproduce

```python
# thread survives close() -- real multiprocessing.Queue, only Settings patched
import threading, time
from unittest.mock import patch, Mock
from graphrag_toolkit.lexical_graph.utils import fm_observability as fmo

before = set(t.name for t in threading.enumerate())
with patch.object(fmo, "Settings") as settings:
    settings.callback_manager = Mock()
    pub = fmo.FMObservabilityPublisher(subscribers=[], interval_seconds=1.0)
    pub.close()

time.sleep(2.5)  # let the publish_stats Timer fire
after = [t for t in threading.enumerate() if t.name not in before]
print("threads still alive after close():", [(t.name, t.is_alive()) for t in after])
print("publisher's current poller alive:", pub.poller.is_alive())
print("poller._discontinue set:", pub.poller._discontinue.is_set())


# unbounded growth -- Queue patched with a Mock, as the unit tests do
import os, subprocess
pid = str(os.getpid())
def rss_mb():
    out = subprocess.run(["ps", "-o", "rss=", "-p", pid], capture_output=True, text=True).stdout
    return int(out.strip()) / 1024

with patch.object(fmo, "Queue") as qcls, patch.object(fmo, "Settings") as settings:
    qcls.return_value = Mock()
    settings.callback_manager = Mock()
    pub = fmo.FMObservabilityPublisher(subscribers=[], interval_seconds=1.0)
    pub.close()

start = rss_mb()
for i in range(5):
    time.sleep(2)
    print(f"t+{2*(i+1):>2}s  rss={rss_mb():8.1f} MB  (+{rss_mb()-start:.1f} MB)  "
          f"mock_calls on queue={len(fmo._fm_observability_queue.mock_calls)}")
```

```shell
$ python repro.py
threads alive after close(): ['MainThread', 'Thread-1', 'Thread-2']
poller thread alive: True
t+ 2s  rss=  1086.4 MB  (+799.6 MB)  mock_calls on queue=353521
t+ 4s  rss=  1899.7 MB  (+1611.5 MB)  mock_calls on queue=697634
t+ 6s  rss=  2674.8 MB  (+2383.4 MB)  mock_calls on queue=1011997
t+ 8s  rss=  3530.1 MB  (+3243.8 MB)  mock_calls on queue=1383564
t+10s  rss=  4397.1 MB  (+4108.3 MB)  mock_calls on queue=1731547

# same repro, leaking for a bounded 20s, then timing process exit
$ time python repro_shutdown.py 20
leaked for 20.0s -> rss 8375 MB, 3318341 retained mock calls
main body done at 17:11:38; measuring shutdown from here
atexit reached at +0.0s
python repro_shutdown.py 20  43.90s user 1.17s system 97% cpu 46.106 total

# per-test RSS at teardown, full unit suite
final rss: 10284 MB   peak(ru_maxrss): 10282 MB
--- top files by cumulative growth ---
+  4019.8 MB  tests/unit/utils/test_io_utils.py
+  1896.1 MB  tests/unit/utils/test_format_helpers.py
+  1846.2 MB  tests/unit/utils/test_llm_cache.py
+   933.9 MB  tests/unit/utils/test_reranker_utils.py
+   814.5 MB  tests/unit/utils/test_fm_observability.py

# the same files run on their own
test_io_utils          final rss:  258 MB
test_format_helpers    final rss:  256 MB
test_llm_cache         final rss:  266 MB
test_reranker_utils    final rss:  257 MB
test_fm_observability  final rss: 1277 MB
```

noel-improv:
## Description

`FMObservabilityPublisher.close()` does not stop its poller thread, so every publisher leaves one running. Under the unit tests that leaked poller busy-loops on a mocked queue and retains every call it makes.

## Changes

- `close()` stops the poller, publishes its final stats, and is idempotent
- `publish_stats()` returns early when the publisher is closed, instead of starting a replacement poller first
- a lock guards both paths, since the Timer thread and the caller of `close()` can race
- the mocked queue in `test_fm_observability.py` raises `queue.Empty`, so the poller polls instead of spinning
- three regression tests: the poller stops on close, no replacement poller appears after the timer fires, final stats publish once

## Problem

`close()` sets `allow_continue = False` and returns. The `threading.Timer` scheduled in `__init__` then fires `publish_stats()`, which stops the current poller and starts a replacement before it checks `allow_continue`. Every publisher leaves a live `FMObservabilityQueuePoller` behind, and calling `close()` does not change that.

With a real queue the leaked thread is idle, because `Queue.get(timeout=1)` blocks for a second per iteration. Under the unit tests it is not idle. The tests patch `fm_observability.Queue` with a plain `Mock`, and `Mock.get()` returns a truthy child immediately rather than raising, so the poll becomes a busy loop. The Mock retains every iteration in `mock_calls`, and nothing trims it.

Related issue (if any): Fixes #450

## Testing

Ran the repro from the issue against both sides. On `main` the poller is alive after `close()` and `_discontinue` is unset; on this branch the poller is stopped and `_discontinue` is set.

Full unit suite on this branch: 1871 passed, peak RSS 443 MB, 43 s. The issue measured 10.3 GB peak for the same suite before the fix. One unrelated error remains locally, `test_boto3_botocore_version_alignment`, which shells out to pip and fails in a uv-created venv.

## Checklist

### DIFF
--- a/lexical-graph/src/graphrag_toolkit/lexical_graph/utils/fm_observability.py
+++ b/lexical-graph/src/graphrag_toolkit/lexical_graph/utils/fm_observability.py
@@ -525,6 +525,7 @@ class FMObservabilityPublisher():
         self.subscribers = subscribers
         self.interval_seconds = interval_seconds
         self.allow_continue = True
+        self._lock = threading.Lock()
         self.poller = FMObservabilityQueuePoller()
         self.poller.start()
 
@@ -534,17 +535,23 @@ class FMObservabilityPublisher():
 
     def close(self):
         """
-        Disables the continuation functionality by updating the internal flag.
+        Stops publishing and shuts down the poller thread.
 
-        This method prevents further continuation of a process by setting the
-        `allow_continue` attribute to `False`. Once invoked, the attribute will
-        indicate that continuation should not occur.
+        Sets `allow_continue` to `False` so that no further poller is scheduled,
+        stops the current poller, and publishes its final stats to the subscribers.
+        Calling this more than once has no additional effect.
 
         Attributes:
             allow_continue (bool): Indicates whether continuation is permitted or not.
                 Set to `False` to disable continuation.
         """
-        self.allow_continue = False
+        with self._lock:
+            if not self.allow_continue:
+                return
+            self.allow_continue = False
+            stats = self.poller.stop()
+        for subscriber in self.subscribers:
+            subscriber.on_new_stats(stats)
 
     def __enter__(self):
         """
@@ -578,16 +585,17 @@ class FMObservabilityPublisher():
         Returns:
             None
         """
-        stats = self.poller.stop()
-        self.poller = FMObservabilityQueuePoller()
-        self.poller.start()
-        if self.allow_continue:
+        with self._lock:
+            if not self.allow_continue:
+                logging.debug('Publisher is closed; not starting a new poller')
+                return
+            stats = self.poller.stop()
+            self.poller = FMObservabilityQueuePoller()
+            self.poller.start()
             logging.debug('Scheduling new poller')
             t = threading.Timer(self.interval_seconds, self.publish_stats)
             t.daemon = True
             t.start()
-        else:
-            logging.debug('Shutting down publisher')
         for subscriber in self.subscribers:
             subscriber.on_new_stats(stats)
 
--- a/lexical-graph/tests/unit/utils/test_fm_observability.py
+++ b/lexical-graph/tests/unit/utils/test_fm_observability.py
@@ -445,6 +445,9 @@ class TestObservabilityIntegration:
     def test_publisher_initialization_sets_up_handlers(self, mock_settings, mock_queue_class):
         """Verify FMObservabilityPublisher sets up handlers on initialization."""
         mock_queue_instance = Mock()
+        # Poll rather than spin: a bare Mock returns a truthy child immediately,
+        # turning the poller's one-second get() into a busy loop.
+        mock_queue_instance.get.side_effect = queue.Empty
         mock_queue_class.return_value = mock_queue_instance
         mock_callback_manager = Mock()
         mock_settings.callback_manager = mock_callback_manager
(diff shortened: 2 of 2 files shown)


noel-improv:
Closing in favour of #451, which reported the same fix first.

Two gaps from this PR are worth carrying over, and I've left them as review comments there: `close()` should be idempotent, and it should publish the final stats returned by `poller.stop()` rather than discarding them.

Thanks @AmirF194 for the offer to stand down — not necessary, you were first.

PR #469, opened 2026-08-13
[FEATURE] (lexical-graph): Auto-tuning batch extraction

### LINKED ISSUE #257: [Feature]: Auto-tuning batch extraction
**Description**

Currently, configuring batch extraction requires users to manually reason about the relationship between batch_size (number of documents to load), average chunks per document, num_workers, and max_batch_size to avoid under-filling batch jobs. Getting this wrong leads to either many small under-utilized batches (slow due to job overhead) or batches that exceed limits.

**Proposal**
 Introduce an auto-tuning where the framework keeps pulling and chunking documents until the specified number of batch jobs are full (up to max_batch_size), then submits them. The user only needs to specify num_workers and max_batch_size — batch_size becomes a derived value.
- Stream/chunk documents incrementally
- Assign chunks to worker buckets
- Stop accepting new documents once all buckets reach max_batch_size, or documents are exhausted. If total chunks fit in fewer batches than num_workers, consolidate to minimize job overhead

oussamahansal:
## Description

Adds an opt-in `auto_tune` mode to batch extraction that removes `batch_size` from the set of knobs a user must hand-tune. Instead of pulling a fixed number of documents per iteration and hoping the resulting chunk count fills Bedrock jobs well, the pipeline streams and chunks documents incrementally, packs the chunks contiguously into jobs of `max_batch_size`, and submits a round of up to `num_workers` jobs once a round fills. Under auto-tuning the user specifies only `num_workers` and `max_batch_size`; `batch_size` becomes derived.

Enable with `BatchConfig(..., auto_tune=True)`. Default is `False` — no behavior change for existing configurations.

## Changes

- `BatchConfig`: new `auto_tune` flag + `__post_init__` validation of `max_batch_size` against the Bedrock bounds.
- `BucketFiller` (new): contiguous, count-based chunk accumulator — mirrors the fixed path's `node_batcher`/`split_nodes` packing (fill to `max_batch_size`, remainder in the last job).
- `ExtractionPipeline`: detects `auto_tune` from the batch extractor's config and branches to `_extract_auto_tuned`; the existing fixed-`batch_size` path is unchanged. Chunking moves to the main process so chunk counts are known before job assignment; the extractor tail still runs across `num_workers` worker processes.
- Docs + integration tests (`BatchExtractAutoTuneToS3`, and an on-demand `BatchExtractSmallBatchToS3` baseline for the fixed-vs-auto-tune comparison).

## Problem

Today, keeping Bedrock batch jobs well-utilized requires reasoning about `batch_size` × avg-chunks-per-doc ÷ `num_workers` vs `max_batch_size`. Get it wrong and you either get many small, under-utilized jobs (slow: each job carries
~10 min of fixed Bedrock overhead, and sub-100-record jobs fall back to synchronous extraction) or jobs that need re-splitting. Auto-tuning derives the document volume so jobs fill to `max_batch_size` automatically.

Related issue (if any): closes #257 

## Testing

## Checklist

---

### DIFF
--- a/docs-site/src/content/docs/lexical-graph/configuring-batch-extraction.mdx
+++ b/docs-site/src/content/docs/lexical-graph/configuring-batch-extraction.mdx
@@ -20,7 +20,10 @@ title: Configuring Batch Extraction
       - [security_group_ids](#security_group_ids)
     - [File management](#file-management)
       - [delete_on_success](#delete_on_success)
+    - [Auto-tuning](#auto-tuning)
+      - [auto_tune](#auto_tune)
   - [Optimizing batch extraction performance](#optimizing-batch-extraction-performance)
+  - [Auto-tuning batch extraction](#auto-tuning-batch-extraction)
 
 ### Overview
 
@@ -78,11 +81,17 @@ An array of security group IDs within your VPC for protecting batch inference jo
 
 Controls whether input and output JSON files are automatically deleted from the local filesystem after successful batch job completion. By default, this is set to `True`. Note that this setting does not affect files stored in S3, which are preserved regardless.
 
+#### Auto-tuning
+
+##### `auto_tune`
+
+When set to `True`, enables [auto-tuning batch extraction](#auto-tuning-batch-extraction). Instead of pulling a fixed number of source documents per iteration, the framework streams and chunks documents incrementally, filling `GraphRAGConfig.extraction_num_workers` buckets up to `max_batch_size` chunks each before submitting a round of batch jobs. For small inputs it consolidates chunks into the fewest jobs needed. With auto-tuning enabled you only need to specify `extraction_num_workers` and `max_batch_size`; `extraction_batch_size` becomes a derived value (if you set it explicitly, it acts as an optional upper bound on the number of documents consumed per round). The default is `False`.
+
 ### Optimizing batch extraction performance
 
 The most important settings for controlling batch extraction performance are:
 
-  - `GraphRAGConfig.extraction_batch_size`: Sets how many source documents go to the extraction pipeline. When calculating this value, consider that the total number of chunks (source documents × average chunks per document) should be sufficient to fill your planned simultaneous batch jobs.
+  - `GraphRAGConfig.extraction_batch_size`: Sets how many source documents go to the extraction pipeline. When calculating this value, consider that the total number of chunks (source documents × average chunks per document) should be sufficient to fill your planned simultaneous batch jobs. If you enable [auto-tuning](#auto-tuning-batch-extraction), you no longer need to calculate this value — the framework derives it for you.
   - `GraphRAGConfig.extraction_num_workers`: Sets how many CPUs run batch jobs simultaneously.
   - `BatchConfig.max_num_concurrent_batches`: Sets how many concurrent batch jobs each worker runs.
   - `BatchConfig.max_batch_size`: Sets the maximum number of chunks per batch job.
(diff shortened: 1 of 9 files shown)


docs-site/src/content/docs/lexical-graph/configuring-batch-extraction.mdx
```diff
@@ -78,11 +81,17 @@ An array of security group IDs within your VPC for protecting batch inference jo
 
 Controls whether input and output JSON files are automatically deleted from the local filesystem after successful batch job completion. By default, this is set to `True`. Note that this setting does not affect files stored in S3, which are preserved regardless.
 
+#### Auto-tuning
+
+##### `auto_tune`
+
+When set to `True`, enables [auto-tuning batch extraction](#auto-tuning-batch-extraction). Instead of pulling a fixed number of source documents per iteration, the framework streams and chunks documents incrementally, filling `GraphRAGConfig.extraction_num_workers` buckets up to `max_batch_size` chunks each before submitting a round of batch jobs. For small inputs it consolidates chunks into the fewest jobs needed. With auto-tuning enabled you only need to specify `extraction_num_workers` and `max_batch_size`; `extraction_batch_size` becomes a derived value (if you set it explicitly, it acts as an optional upper bound on the number of documents consumed per round). The default is `False`.
```
acarbonetto, line 88: please mention how this coordinates with parallelization - if at all.  

I believe this works on a per-thread basis

docs-site/src/content/docs/lexical-graph/configuring-batch-extraction.mdx
```diff
@@ -78,11 +81,17 @@ An array of security group IDs within your VPC for protecting batch inference jo
...
+
 ### Optimizing batch extraction performance
 
 The most important settings for controlling batch extraction performance are:
 
-  - `GraphRAGConfig.extraction_batch_size`: Sets how many source documents go to the extraction pipeline. When calculating this value, consider that the total number of chunks (source documents × average chunks per document) should be sufficient to fill your planned simultaneous batch jobs.
+  - `GraphRAGConfig.extraction_batch_size`: Sets how many source documents go to the extraction pipeline. When calculating this value, consider that the total number of chunks (source documents × average chunks per document) should be sufficient to fill your planned simultaneous batch jobs. If you enable [auto-tuning](#auto-tuning-batch-extraction), you no longer need to calculate this value — the framework derives it for you.
```
acarbonetto:
```suggestion
  - `GraphRAGConfig.extraction_batch_size`: (ignored if `auto_tune` is true) Sets how many source documents go to the extraction pipeline. When calculating this value, consider that the total number of chunks (source documents × average chunks per document) should be sufficient to fill your planned simultaneous batch jobs. If you enable [auto-tuning](#auto-tuning-batch-extraction), this value is determined dynamically.
```

docs-site/src/content/docs/lexical-graph/configuring-batch-extraction.mdx
```diff
@@ -78,11 +81,17 @@ An array of security group IDs within your VPC for protecting batch inference jo
 
 Controls whether input and output JSON files are automatically deleted from the local filesystem after successful batch job completion. By default, this is set to `True`. Note that this setting does not affect files stored in S3, which are preserved regardless.
 
+#### Auto-tuning
+
+##### `auto_tune`
```
acarbonetto, line 86: consider a more explicit name: for example `auto_tune_batch_size`

oussamahansal:
I added `BatchExtractSmallBatchToS3` and `BatchExtractAutoTuneToS3` integration tests so the fixed and auto-tuned paths can be compared directly, and ran both.

- Fuller jobs: average fill is higher when auto tuning(255 vs 147) — each Bedrock job carries more records for the same number of jobs.

- auto-tune keeps job sizes in a tight 239–293 band versus a wide 102–267 spread for the fixed path
- No jobs near the floor, the fixed path leaves several jobs just above the Bedrock 100-record minimum (102, 103), which is the under-fill the round-packing is meant to avoid. The 293-record job on the auto-tune side is the tail-merge exceeding max_batch_size as designed, rather than spilling into an undersized sync-fallback job.

Metric | Small-batch (fixed, batch_size=10) | Auto-tune
-- | -- | --
Batch jobs | 6 | 6
Job sizes | 102, 103, 119, 120, 173, 267 | 239, 249, 250, 250, 250, 293
Avg fill | 147 | 255

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/extract/bucket_filler.py
```diff
@@ -0,0 +1,77 @@
...
+
+        The caller flushes when this is True *before* appending, so a document
+        that fits within a round is not split across rounds -- while the flushed
+        round stays near-full (the buffer is already close to capacity). An empty
+        buffer never overshoots, so an oversized document is still accepted and
+        then drained in full rounds by the caller.
+        """
+        return bool(self._buffer) and (len(self._buffer) + num_chunks > self.round_capacity)
+
+    def drain_round(self) -> List[List[BaseNode]]:
```
acarbonetto, line 62: drain_round can leave a sub-minimum last job that falls back to slow sync extraction (correctness) — BucketFiller.drain_round slices a partial buffer into fixed max_batch_size jobs but, unlike the existing split_nodes, does not merge a sub-minimum remainder into the prior batch. So an overshoot-triggered flush (e.g. an 80-chunk buffer) or a consolidation round (e.g. 210 → [100, 100, 10]) produces a job below Bedrock's 100-record minimum, which _process_nodes then routes to slow chunk-by-chunk synchronous extraction — exactly what auto-tuning claims to avoid. The tail-merge in _extract_auto_tuned only reconciles the final remainder against the held round, not these intra-round under-min jobs. The
  unit tests don't cover the overshoot-flush case, so this passes CI while missing the goal.

acarbonetto: Finding #1 is confirmed — intra-round under-min jobs can and do exist. Here's the proof chain and a clean example.

  The claim only holds if the individual buckets inside a round each reach Bedrock's 100-record gate separately. They do:
1. drain_round (bucket_filler.py:74-77) slices with plain fixed-width slicing and no tail-merge: 
```
return [take[i:i + self.max_batch_size]
          for i in range(0, len(take), self.max_batch_size)]
```
2.  Each round's bucket list goes to _run_extractor_round → run_pipeline (pipeline_utils.py:46), which does p.map(transform, node_batches) — every bucket is transformed as an independent unit
3. Each bucket independently lands in the batch extractor's _process_nodes, whose gate (batch_extractor_base.py:239-242) routes any bucket < BEDROCK_MIN_BATCH_SIZE (100) to the synchronous non-batch extractor.

  So a 10-node bucket sitting next to two 100-node buckets in the same round still hits the sync fallback on its own.

**Cleanest airtight example**
  
  Config: num_workers=3, max_batch_size=100 → round_capacity=300. Corpus: 210 chunks total (e.g. 3 docs × 70 chunks).
  
  Trace through _extract_auto_tuned:
  - Each doc appends 70 chunks. would_overshoot? 140+70=210 > 300 → no. while pending() >= 300? no. So nothing flushes; buffer = 210, held = None.
  - Input exhausts → final_jobs = filler.drain_round() on 210 = [100, 100, 10], final_total = 210.
  - Tail-merge guard (line 708): requires 0 < final_total < BEDROCK_MIN(100). Here final_total = 210, so 210 < 100 is False → no merge.
  - emit_jobs([100, 100, 10], 'exhausted', …) → the 10-node job → _process_nodes → sync fallback.

  Contrast — split_nodes(210, batch_size=100) (the fixed path, batch_inference_utils.py:58-64):
  - i=0: 210-(0+100)=110 ≥ 100 → emit [0:100]
  - i=100: 210-(100+100)=10 < 100 → emit [100:210] = 110 nodes, break
  - Result: [100, 110] — no sub-min job. 
  
  Same input: fixed path → [100, 110] (all batched); auto-tune → [100, 100, 10] (the 10 goes synchronous). That's the divergence, and it's exactly the sub-100 fallback the feature claims to eliminate.

  So the intra-round under-min job is real. The root cause is: drain_round reimplements packing but omits split_nodes' deliberate < BEDROCK_MIN tail-merge (line 59). The clean fix is to route the per-round slicing through split_nodes (or lift its merge rule into drain_round) so the last job of every round — not just the final remainder — absorbs a sub-100 tail.

docs-site/src/content/docs/lexical-graph/configuring-batch-extraction.mdx
acarbonetto: we are happy with `auto_tune`?

acarbonetto (approved):

PR #470, opened 2026-08-13
[FIX] (CI): Pin nltk<3.10.3 to unblock tests

### LINKED ISSUE #468: [BUG] nltk 3.10.3 pathsec hardlink check breaks all CI test runs (uv-hardlinked site-packages)
### Package version

3.18.5 (main, 8e779464)

### Package

lexical-graph

### Python version

3.10, 3.11, 3.12 (CI)

### Operating System

Linux (ubuntu-latest CI runners)

### Description

Every CI run since 2026-08-12 23:44 UTC fails in `tests/unit/indexing/extract/test_chunking.py` with a `PermissionError` raised inside NLTK.

`nltk/pathsec.py` refuses to open any file whose `st_nlink` is greater than 1. llama-index ships its NLTK data inside site-packages, and `uv pip install` hardlinks from its cache into the venv on Linux, so those data files arrive with `st_nlink=2` and NLTK rejects them.

The check is new in nltk 3.10.3, published to PyPI at `2026-08-12T23:44:13Z`. I diffed the two wheels: `pathsec.py` is in both, but the `st_nlink` check appears only in 3.10.3 (4 references, versus 0 in 3.10.2).

Recent runs, independent of branch:

| Run (UTC) | Branch | nltk | Result |
| --- | --- | --- | --- |
| 22:31, 22:38 | main | 3.10.2 | pass |
| 22:35, 22:56 | 2 feature branches | 3.10.2 | pass |
| 00:43 | fix/hypothesis-property-key-collision | 3.10.3 | fail |
| 01:12 | fix/observability-poller-thread-leak | 3.10.3 | fail |

`llama-index-core 0.14.23` requires `nltk>=3.9.3` with no upper bound, and nothing here pins nltk, so every fresh install takes the newest release. macOS escapes it because APFS clonefile leaves `st_nlink` at 1, which is why local runs still pass.

This is probably not CI-only. The failing path is ordinary sentence splitting, so anyone installing lexical-graph with uv on Linux should hit it. I have not reproduced it outside CI.

Pinning `nltk<3.10.3` covers both CI and users. `UV_LINK_MODE=copy` in the test workflows would unblock CI alone.

### Steps to reproduce

On Linux, in a uv-created venv (uv hardlinks from its cache):

```
uv venv --seed .venv
uv pip install --python .venv/bin/python -e '.[test]'   # resolves nltk 3.10.3
PYTHONPATH=src .venv/bin/python -m pytest tests/unit/indexing/extract/test_chunking.py
```

```shell
.venv/lib/python3.11/site-packages/nltk/data.py:444: in open
.venv/lib/python3.11/site-packages/nltk/pathsec.py:796: in open
.venv/lib/python3.11/site-packages/nltk/pathsec.py:721: in _hardened_open
E   PermissionError: Security Violation [pathsec.open]: refusing multiply-linked file '/home/runner/work/graphrag-toolkit/graphrag-toolkit/lexical-graph/.venv/lib/python3.11/site-packages/llama_index/core/_static/nltk_cache/corpora/stopwords/english' (st_nlink=2); a hardlink can point at an outside-root inode (CWE-59)

FAILED tests/unit/indexing/extract/test_chunking.py::TestChunkingBasicBehavior::test_chunking_with_overlap_contains_original
```

oussamahansal:
## Description

nltk 3.10.3 added a `pathsec` check (CWE-59) that refuses to open hardlinked files. Since `uv pip install` hardlinks packages into the venv on Linux, the bundled NLTK data arrives with st_nlink=2 and gets rejected — breaking sentence splitting and failing CI (`test_chunking.py`) on every branch.

## Changes

- Pin `nltk<3.10.3` in lexical-graph requirements — fixes CI **and** end-users who install via uv on Linux.
- Set `UV_LINK_MODE=copy` in the lexical-graph and byokg-rag test workflows as CI defense-in-depth.

## Problem

Related issue (if any): closes #468

## Testing

## Checklist

---

### DIFF
--- a/.github/workflows/byokg-rag-tests.yml
+++ b/.github/workflows/byokg-rag-tests.yml
@@ -77,6 +77,10 @@ jobs:
         run: uv venv --seed .venv
 
       - name: Install dependencies
+        env:
+          # Copy instead of hardlink into the venv so bundled data files have
+          # st_nlink=1; nltk's pathsec check refuses to open multiply-linked files.
+          UV_LINK_MODE: copy
         run: |
           uv pip install --python .venv/bin/python \
             pytest \
--- a/lexical-graph/src/graphrag_toolkit/lexical_graph/requirements.txt
+++ b/lexical-graph/src/graphrag_toolkit/lexical_graph/requirements.txt
@@ -9,6 +9,9 @@ llama-index-embeddings-bedrock>=0.8.2
 llama-index-llms-anthropic>=0.11.8
 llama-index-llms-bedrock-converse>=0.14.16
 lru-dict==1.3.0
+# Pinned below 3.10.3: that release's pathsec hardlink check (CWE-59) rejects
+# uv-hardlinked NLTK data files, breaking sentence splitting. See issue 468.
+nltk<3.10.3
 pipe==2.2
 python-dotenv==1.2.2
 smart_open==7.1.0
--- a/.github/workflows/lexical-graph-tests.yml
+++ b/.github/workflows/lexical-graph-tests.yml
@@ -79,6 +79,10 @@ jobs:
       - name: Install dependencies
         # Separate commands: requirements.txt opens with `--only-binary :all:`,
         # which would block building the local package if combined.
+        env:
+          # Copy instead of hardlink into the venv so bundled data files have
+          # st_nlink=1; nltk's pathsec check refuses to open multiply-linked files.
+          UV_LINK_MODE: copy
         run: |
           uv pip install --python .venv/bin/python \
             -r src/graphrag_toolkit/lexical_graph/requirements.txt


lexical-graph/src/graphrag_toolkit/lexical_graph/requirements.txt
```diff
@@ -9,6 +9,9 @@ llama-index-embeddings-bedrock>=0.8.2
 llama-index-llms-anthropic>=0.11.8
 llama-index-llms-bedrock-converse>=0.14.16
 lru-dict==1.3.0
+# Pinned below 3.10.3: that release's pathsec hardlink check (CWE-59) rejects
+# uv-hardlinked NLTK data files, breaking sentence splitting. See issue 468.
+nltk<3.10.3
```
noel-improv, line 14: **[nit-sugg]**: Would `nltk!=3.10.3` work as well here?

noel-improv (approved):
lgtm, just a single recomm

.github/workflows/byokg-rag-tests.yml
```diff
@@ -77,6 +77,10 @@ jobs:
         run: uv venv --seed .venv
 
       - name: Install dependencies
+        env:
+          # Copy instead of hardlink into the venv so bundled data files have
+          # st_nlink=1; nltk's pathsec check refuses to open multiply-linked files.
+          UV_LINK_MODE: copy
```
acarbonetto, line 83: does this need to be in the README and documentation too to install locally?  
What about the benchmarking scripts, so we need this env when installing dependencies for benchmarking?

acarbonetto:
please raise a follow-up ticket (blocked on nltk update) to revert this change in the future.

oussamahansal:
follow up ticket created: #471

.github/workflows/byokg-rag-tests.yml
oussamahansal: I checked them. The lexical-graph README documents `pip install`, and pip copies files (not hardlinks) on Linux. and for Benchmarking / integration tests, they install via `pip` on SageMaker (`on-start.sh`, `run_test_suite.sh`) which is also unaffected

lexical-graph/src/graphrag_toolkit/lexical_graph/requirements.txt
oussamahansal: yea it would work but this is just a workaround.. the real fix should be upstream

acarbonetto (approved):
only affects `uv` 👍

PR #476, opened 2026-08-14
[FIX] (lexical-graph): fix KeyError 'params' in domain-entity batch insert

oussamahansal:
## Description

- insert_domain_entity passed a flat {'entityId': e_id} dict and a scalar
$entityId query to the batch graph client. When batch writes are enabled,
GraphBatchClient.execute_query_with_retry reads properties['params'], so the
missing key raised KeyError: 'params' and aborted the entire build whenever
include_domain_labels=True.

## Changes

- Align this call site with every other graph builder: use the
UNWIND $params AS params form (binding params.entityId) and wrap the params
via _to_params so they carry the {'params': [...]} shape the batch client
requires. Label escaping and id-as-bound-parameter behaviour are unchanged.

## Problem

_to_params(p) wraps params into the {'params': [p]} shape the batch client requires, and every graph-builder call site uses it — except insert_domain_entity at entity_graph_builder.py:128, which passes a raw {'entityId': e_id} and a scalar $entityId query instead of the UNWIND $params pattern.
Under batch writes, the client does properties['params'] (graph_batch_client.py:141) → no 'params' key → KeyError: 'params', which aborts the entire build. The two queries then time out because there's no graph.

Related issue (if any): #

## Testing

## Checklist

---

### DIFF
--- a/lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/build/entity_graph_builder.py
+++ b/lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/build/entity_graph_builder.py
@@ -124,8 +124,8 @@ class EntityGraphBuilder(GraphBuilder):
                     e_id = entity.entityId
                     e_label = escape_cypher_label(label_from(entity.classification or DEFAULT_CLASSIFICATION))
                     e_comment = f'// awsqid:{e_id}-{e_label}'.replace('\r', ' ').replace('\n', ' ')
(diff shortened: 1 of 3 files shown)


lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/build/entity_graph_builder.py
acarbonetto, line 123: nit: I'll raise a separate issue to fix this...we don't need a unique uuid for each request.  see: #477

acarbonetto (approved):

PR #481, opened 2026-08-17
[BUG]: Validate chunk ids and bound downloads in S3ChunkStore

oussamahansal:
## Description

Chunk ids are interpolated into the S3 key, so validate them before use: reject empty/blank ids and ids containing a path separator or control characters, which would otherwise resolve to an unexpected key outside the configured prefix. Both read and write paths go through _key(), so the check covers get and put.

## Changes

- Validate `chunk_id` in `S3ChunkStore._key()` before building the object key: reject empty/blank ids and ids containing a path separator (`/`, `\`) or control characters. Keeps derived keys inside the configured prefix.
Both `get`/`get_batch` and `put`/`put_batch` route through `_key()`, so the check applies to reads and writes.

## Problem

Related issue (if any): #

## Testing

## Checklist

---

### DIFF
--- a/lexical-graph/src/graphrag_toolkit/lexical_graph/storage/chunk/s3_chunk_store.py
+++ b/lexical-graph/src/graphrag_toolkit/lexical_graph/storage/chunk/s3_chunk_store.py
@@ -3,6 +3,7 @@
 
 import concurrent.futures
 import contextlib
+import re
 from urllib.parse import urlparse, parse_qs
 import logging
 from typing import Dict, List, Optional
@@ -107,24 +108,28 @@ class S3ChunkStore(ChunkStore):
 
         return max(1, min(num_items, configured))
 
+    # Allowlist for chunk ids: an allowlist (vs blocking specific separators)
+    # closes URL-encoding, Unicode, and double-encoding bypasses in one check.
+    # IdGenerator ids (aws::<hex>:<hex>:<hex>, optional tenant segment) fit.
+    _CHUNK_ID_PATTERN = re.compile(r'[A-Za-z0-9:._-]+')
+
     @staticmethod
     def _validate_chunk_id(chunk_id: str) -> None:
         """
-        Reject chunk ids that could escape the configured prefix.
+        Reject any chunk id that isn't a plain [A-Za-z0-9:._-] token.
 
-        S3 keys are literal, so a prefix-scoped IAM policy would still match
-        `prefix/../secret.txt`. But botocore puts path segments on the wire
-        unencoded, so a normalizing proxy or an S3-compatible endpoint reached
-        through `endpoint_url` can collapse `../` before the request gets to
-        S3, escaping the prefix with IAM none the wiser. Blocking separators
-        stops that; real ids are IdGenerator hashes and hold none.
+        Blocking separators is what matters for the key: botocore sends key
+        segments unencoded, so a normalizing proxy or S3-compatible endpoint
+        could collapse `../` before S3/IAM sees it and escape the prefix. The
+        allowlist blocks that and every encoded variant at once.
         """
         if not chunk_id or not chunk_id.strip():
             raise ValueError('chunk_id must be a non-empty string.')
-        if '/' in chunk_id or '\\' in chunk_id:
-            raise ValueError(f'chunk_id must not contain a path separator: {chunk_id!r}')
-        if any(ord(c) < 0x20 or ord(c) == 0x7f for c in chunk_id):
-            raise ValueError(f'chunk_id must not contain control characters: {chunk_id!r}')
+        if not S3ChunkStore._CHUNK_ID_PATTERN.fullmatch(chunk_id):
+            raise ValueError(
+                f'chunk_id contains invalid characters (allowed: alphanumeric, colon, '
+                f'period, underscore, hyphen): {chunk_id!r}'
+            )
 
     def _key(self, chunk_id: str) -> str:
         self._validate_chunk_id(chunk_id)
@@ -162,16 +167,19 @@ class S3ChunkStore(ChunkStore):
         if content_length is not None and content_length > self.max_chunk_bytes:
             return self._skip_oversized(key, content_length)
 
-        # closing() returns the connection to the pool on every path. With a
-        # known length a full read() drains the stream, so botocore verifies the
-        # byte count against Content-Length and a truncated transfer raises
-        # rather than returning short text; with no declared length, bound the
-        # read so an unknown-size object still can't exhaust memory.
+        # Always bound the read: memory safety must not depend on a spoofable
+        # Content-Length. closing() returns the connection to the pool.
         with contextlib.closing(response['Body']) as body:
-            data = body.read() if content_length is not None else body.read(self.max_chunk_bytes + 1)
+            data = body.read(self.max_chunk_bytes + 1)
 
         if len(data) > self.max_chunk_bytes:
             return self._skip_oversized(key)
+        # Content-Length is kept only as a secondary integrity signal.
+        if content_length is not None and len(data) != content_length:
+            logger.warning(
+                f'Content-Length mismatch [key: {key}, declared: {content_length}, '
+                f'actual: {len(data)}]'
+            )
         return data.decode('UTF-8')
 
     def get_batch(self, chunk_ids: List[str]) -> Dict[str, str]:
--- a/lexical-graph/src/graphrag_toolkit/lexical_graph/storage/chunk/s3_chunk_store.py
+++ b/lexical-graph/src/graphrag_toolkit/lexical_graph/storage/chunk/s3_chunk_store.py
@@ -153,7 +153,15 @@ class S3ChunkStore(ChunkStore):
         return None
 
     def _get_chunk(self, chunk_id: str, s3_client) -> Optional[str]:
-        key = self._key(chunk_id)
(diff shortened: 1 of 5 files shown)


lexical-graph/src/graphrag_toolkit/lexical_graph/storage/chunk/s3_chunk_store.py
```diff
@@ -94,17 +100,43 @@ def _num_workers(self, num_items: int) -> int:
...
         return max(1, min(num_items, configured))
 
+    @staticmethod
+    def _validate_chunk_id(chunk_id: str) -> None:
+        """
+        Reject chunk ids that could escape the configured prefix.
+
+        The key is f'{prefix}/{chunk_id}.txt', so a separator in the id opens a
+        new path segment and lets a crafted id touch any *.txt object in the
+        bucket. Real ids are IdGenerator hashes and hold no separators.
```
noel-improv: The check is worth having, but not for the reason given here. S3 keys are literal, so `prefix/../secret.txt` is its own object and still starts with `prefix/` and a prefix-scoped IAM policy matches it too. What botocore does do is put the segments on the wire unencoded:

```
'prefix/../secret.txt' -> https://bkt.s3.amazonaws.com/prefix/../secret.txt
```

(botocore 1.40.61). So the case that bites is a normalizing proxy or one of the S3-compatible endpoints we support through `endpoint_url`, where the collapse happens before S3 sees it and IAM never gets a say. Same check, and it reads as deliberate instead of speculative.

lexical-graph/src/graphrag_toolkit/lexical_graph/storage/chunk/s3_chunk_store.py
```diff
@@ -94,17 +100,43 @@ def _num_workers(self, num_items: int) -> int:
...
             raise
-        return response['Body'].read().decode('UTF-8')
+        # Read one byte past the cap to detect an over-limit object without
+        # pulling the whole thing into memory.
+        body = response['Body'].read(self.max_chunk_bytes + 1)
+        if len(body) > self.max_chunk_bytes:
+            raise ValueError(
+                f'Chunk text at [bucket: {self.bucket_name}, key: {key}] exceeds the '
+                f'maximum of {self.max_chunk_bytes} bytes.'
+            )
```
noel-improv: The old `.read()` with no argument drained the stream and verified content length on the way out. Bounded reads don't, so on the raise path `response['Body']` goes out of scope unclosed and the connection never returns to the pool. `get_batch` runs these across a thread pool, so it compounds. `contextlib.closing` around the body, or a close before the raise.

lexical-graph/src/graphrag_toolkit/lexical_graph/storage/chunk/s3_chunk_store.py
```diff
@@ -94,17 +100,43 @@ def _num_workers(self, num_items: int) -> int:
...
+        return body.decode('UTF-8')
 
     def get_batch(self, chunk_ids: List[str]) -> Dict[str, str]:
```
noel-improv, line 195: Downstream of the cap: `future.result()` at line 156 re-raises, so the `ValueError` leaves `get_batch` and takes every other chunk in the batch with it. A missing chunk degrades to `None` and the retrieval carries on; an oversized one kills the query. Worth picking one on purpose. I'd log and skip to match the missing-key path, though there's a real argument for failing loud rather than quietly serving thin context.

lexical-graph/src/graphrag_toolkit/lexical_graph/storage/chunk/s3_chunk_store.py
```diff
@@ -68,7 +72,8 @@ def __init__(self,
                  prefix: Optional[str] = None,
                  kms_key_arn: Optional[str] = None,
                  fallback: Optional[ChunkStore] = None,
-                 num_threads: Optional[int] = None):
+                 num_threads: Optional[int] = None,
+                 max_chunk_bytes: int = DEFAULT_MAX_CHUNK_BYTES):
```
noel-improv, line 82: Nit: `max_chunk_bytes=0` reads like "no cap" and rejects every non-empty chunk. A guard next to the `bucket_name` one at line 77 would catch it.

noel-improv (commented):
thanks for taking this on, somehow got missed while working on the parent PR. Left a few comments below

lexical-graph/src/graphrag_toolkit/lexical_graph/storage/chunk/s3_chunk_store.py
```diff
@@ -94,17 +103,51 @@ def _num_workers(self, num_items: int) -> int:
...
         except ClientError as e:
             if e.response.get('Error', {}).get('Code') in MISSING_KEY_CODES:
                 return None
             raise
-        return response['Body'].read().decode('UTF-8')
+        # Read one byte past the cap to size-check without pulling the whole
+        # object into memory; closing() returns the connection to the pool on
+        # every path, including the oversized one where the stream isn't drained.
+        with contextlib.closing(response['Body']) as body:
+            data = body.read(self.max_chunk_bytes + 1)
```
noel-improv: Also consider, `read(amt)` skips the content-length check that a bare `.read()` does, so a truncated transfer returns short text silently sizing off `response['ContentLength']` before reading keeps the cap and the verification.

noel-improv (approved):
Approved. Thanks for addressing items from last review, one optional note there

lexical-graph/src/graphrag_toolkit/lexical_graph/storage/chunk/s3_chunk_store.py
```diff
@@ -94,17 +103,51 @@ def _num_workers(self, num_items: int) -> int:
...
+        through `endpoint_url` can collapse `../` before the request gets to
+        S3, escaping the prefix with IAM none the wiser. Blocking separators
+        stops that; real ids are IdGenerator hashes and hold none.
+        """
+        if not chunk_id or not chunk_id.strip():
+            raise ValueError('chunk_id must be a non-empty string.')
+        if '/' in chunk_id or '\\' in chunk_id:
+            raise ValueError(f'chunk_id must not contain a path separator: {chunk_id!r}')
+        if any(ord(c) < 0x20 or ord(c) == 0x7f for c in chunk_id):
+            raise ValueError(f'chunk_id must not contain control characters: {chunk_id!r}')
```
andy-k-improving: Nit: Perhaps to check on s3 file key length, the max is 1024 bytes.

andy-k-improving (approved):

mykola-pereyma (approved):
LGTM

lexical-graph/src/graphrag_toolkit/lexical_graph/storage/chunk/s3_chunk_store.py
```diff
@@ -94,17 +108,79 @@ def _num_workers(self, num_items: int) -> int:
...
+
+        # Always bound the read: memory safety must not depend on a spoofable
+        # Content-Length. closing() returns the connection to the pool.
+        with contextlib.closing(response['Body']) as body:
+            data = body.read(self.max_chunk_bytes + 1)
+
+        if len(data) > self.max_chunk_bytes:
+            return self._skip_oversized(key)
+        # Content-Length is kept only as a secondary integrity signal.
+        if content_length is not None and len(data) != content_length:
```
noel-improv, line 187: Could the mismatch return `None` instead of logging, so it's treated as a miss?

lexical-graph/src/graphrag_toolkit/lexical_graph/storage/chunk/s3_chunk_store.py
```diff
@@ -94,17 +108,79 @@ def _num_workers(self, num_items: int) -> int:
...
             if e.response.get('Error', {}).get('Code') in MISSING_KEY_CODES:
                 return None
             raise
-        return response['Body'].read().decode('UTF-8')
+
+        # Enforce the cap on the declared length up front, so an oversized object
+        # is never pulled into memory.
+        content_length = response.get('ContentLength')
+        if content_length is not None and content_length > self.max_chunk_bytes:
+            return self._skip_oversized(key, content_length)
```
noel-improv: `test_declared_length_over_cap_is_skipped_without_reading` asserts `read` wasn't called but doesn't assert `close`, so it passes either way.

lexical-graph/src/graphrag_toolkit/lexical_graph/storage/chunk/s3_chunk_store.py
```diff
@@ -94,17 +108,79 @@ def _num_workers(self, num_items: int) -> int:
...
+        # fail the whole batch, and the fallback still gets a chance.
+        detail = f', content_length: {content_length}' if content_length is not None else ''
+        logger.warning(
+            f'Skipping oversized chunk [bucket: {self.bucket_name}, key: {key}{detail}, '
+            f'max_bytes: {self.max_chunk_bytes}]'
+        )
+        return None
 
     def _get_chunk(self, chunk_id: str, s3_client) -> Optional[str]:
+        key = self._key(chunk_id)
```
noel-improv: `_skip_oversized` takes the opposite line ("one bad chunk shouldn't fail the whole batch"). Worth applying the same rule to an unusable id on reads, and keeping the raise for `put`/`put_batch`?

noel-improv (approved):
lgtm, just a few comments raised

Issue #485, opened 2026-08-18
[FEATURE] Cut `lexical.long` integration-test runtime dominated by the batch-extract steps


acarbonetto:
### Package

lexical-graph

### Problem statement

The `lexical.long` integration-test suite takes **over 2 hours** end-to-end, and the
overwhelming majority of that time is spent in the two batch-extract steps:

- `batch_extract.BatchExtractToS3`
- `batch_extract.BatchExtractAutoTuneToS3`

Together these two steps take roughly **1 hour 20 minutes** (for ONE of the extract processes) for a **70-document** corpus
(`integration-tests/source-data/corpus-modified.json`). This is too slow for a
70-document test set and makes the long suite painful to iterate on.

### Why it is slow (root cause)

Both tests call `graph_index.extract(...)` over the **same** 70-doc corpus, and they run
**sequentially** in a single SageMaker notebook (`test_suite.py` runs the test list in a
`for` loop). So we pay the batch-extract cost **twice, back-to-back**.

Wall-clock for a batch-extract step is **not** driven by the number of documents. From the
lexical-graph batch-inference implementation:

- **Fixed per-job Bedrock overhead dominates.** Each Bedrock `create_model_invocation_job`
  carries a large fixed queue/startup cost (job scheduling + model provisioning) that is
  independent of record count.
  (`lexical-graph/.../indexing/utils/batch_inference_utils.py`)
- **Completion is discovered on a 60-second poll cadence, and the poll sleeps *before* the
  first check** — so even an instantly-completing job costs ≥60s of polling latency:
  ```python
  # batch_inference_utils.py — wait_for_job_completion
  while status not in ['Completed', 'Failed', 'Stopped', 'PartiallyCompleted', 'Expired']:
      time.sleep(60)
      response = bedrock_client.get_model_invocation_job(jobIdentifier=job_arn)
      status = response['status']
  ```
- **Bedrock has a 100-record-per-job minimum** (`BEDROCK_MIN_BATCH_SIZE = 100`) and the code
  merges any sub-100 tail into the previous job. So a small corpus does **not** produce
  smaller/faster jobs — it produces the *same single fully-overhead job*.
- Rounds are **sequential**: with `BatchExtractToS3`'s `extraction_batch_size=100` (documents)
  the 70-doc corpus is one round; auto-tune uses `round_capacity = num_workers * max_batch_size`
  (`2 * 250 = 500`), also one round. Within a round, jobs run concurrently, so total wall-clock
[...]

PR #486, opened 2026-08-18
[FEATURE] Split long integration extract tests

acarbonetto:
## Description

1. Update the `lexical.long` integration-test suite to run the BatchExtractAutoTuneToS3 extraction by default. 

This is safe because `BatchExtractAutoTuneToS3` is standalone: it writes
`params['auto_tune_batch_collection_id']`, which is **never consumed** by any downstream test
(only `BatchExtractToS3`'s `batch_collection_id` / `multihop_expected_num_batch_docs` feed
`batch_build.BuildFromS3` and the queries).

2. `BatchExtractAutoTuneToS3` now works with `batch_build.BuildFromS3`
  `lexical.long` pairs the auto-tune extract with `BuildFromS3`, but the auto-tune extract
  previously published its output under keys/prefix that `BuildFromS3` doesn't read, so the build
  step would `KeyError` before using it. Fixed on the extract side so the contract matches
  `BatchExtractToS3`:
  - Extracted docs now written to the `…/extracted` prefix (was `…/extracted-auto-tune`), which is
    what `BuildFromS3` reads.
  - Publishes `params['batch_collection_id']` and `params['multihop_expected_num_batch_docs']`
    (was the orphaned `auto_tune_batch_collection_id`), which `BuildFromS3` and its assertion consume.
  - Kept auto-tune–specific behavior: `auto_tune=True`, no fixed `batch_size`/`max_num_concurrent_batches`,
    and the `…/batch-inference-auto-tune` **working** prefix (irrelevant to the build; preserves the
    small-batch fill comparison).

3. Opt-in extraction doc cap for fast runs
  New `BENCHMARK_EXTRACT_DOC_LIMIT` caps how many source documents are extracted (first N only),
  to shrink the ~1h20m extract phase during development/smoke runs. Unset = all docs (unchanged
  behavior).
  - Set via env (`.env` / `.env.testing`) or the new `build-tests.sh --benchmark-extraction-doc-limit N`
    flag (mirrors the existing `BENCHMARK_QA_LIMIT` plumbing: CLI → `.env.testing` → runtime env).
  - Applied in `apply_extraction_doc_limit()` **before** `extract()`, so it's worker-count agnostic
    (single- and multi-threaded runs behave identically) and downstream assertions / `BuildFromS3`
    counts stay consistent (they read `len(docs)` after the cap).
  - Applies to all four `batch_extract.py` tests.

4. Removed tests `BatchExtractSmallBatchToS3 and 

## Changes

- `integration-tests/lexical.long` — uses `batch_extract.BatchExtractAutoTuneToS3` (replaces `BatchExtractToS3`).
- `integration-tests/README.md` — documented the new suite in the *Available Test Suites* table.

No test-script (`.py`) or `lexical-graph` source changes — this PR only re-partitions the suite files.

## Problem

`lexical.long` takes **over 2 hours**, with ~**1h20m** spent in the two batch-extract steps for
a 70-document corpus. The two extracts (`BatchExtractToS3` and `BatchExtractAutoTuneToS3`) run
sequentially over the same corpus, so the batch-extract cost is paid twice, back-to-back. Batch
wall-clock is dominated by fixed per-job Bedrock overhead and a 60s completion-poll cadence — not
the document count — so parallelizing the two independent extracts is the immediate win.

Related issue (if any): #485

## Testing

- [ ] Unit tests added/updated — n/a (suite-file re-partition only)
- [x] Integration tests added (as appropriate) — new `lexical.autotune.long` suite; `lexical.long` still covers the fixed-batch path + downstream build/query.
- [ ] Existing tests pass (`pytest`) — n/a

Manual validation to run before/after merge:

```bash
# Fixed-batch path + downstream build/query/progress
sh build-tests.sh --test-file lexical.long
```

image

## Checklist

- [x] Code follows existing style and conventions (suite files match the existing `<module>.<ClassName>` format)
- [x] License headers present on new files — n/a (suite files are plain test-name lists, consistent with `lexical.short` / `byokg.short`)
- [x] Documentation updated (README *Available Test Suites* table)
- [x] No breaking changes — `lexical.long` retains all previously-covered functionality except the auto-tune extract which replaces the existing extract. 
---


integration-tests/lexical.autotune.long
```diff
@@ -0,0 +1,2 @@
+batch_extract.BatchExtractAutoTuneToS3
+batch_build.BuildFromS3
```
noel-improv: Can we verify this suite populates what `BuildFromS3` reads? It starts with `params['batch_collection_id']` and asserts on `params['multihop_expected_num_batch_docs']`, both set by `BatchExtractToS3` (still in `lexical.long`); the auto-tune test sets `auto_tune_batch_collection_id`. Looks like a `KeyError` after the ~40min extract. Also the paths differ, `extracted-auto-tune` vs the hardcoded `extracted`.

noel-improv (commented):
Neat split for the lexical integration test, just one potentially blocking comment below

integration-tests/lexical.autotune.long
acarbonetto: @oussamahansal is the intention of batch_extract.BatchExtractAutoTuneToS3 to just throw away the extracted results?

oussamahansal: So the auto-tune test doesnt hand its collection to any build or query step, it is a standalone extraction test not part of the (extract,build,query chain). Its actual value is the "batch job" input files it leaves in `batch-inference-auto-tune/` which are what we compare against for the job count and fill

integration-tests/test-scripts/graphrag_toolkit_tests/batch_extract.py
```diff
@@ -28,11 +28,45 @@ def get_metadata(data):
...
+        limit = int(raw_limit)
+    except ValueError:
+        raise ValueError(
+            f"BENCHMARK_EXTRACT_DOC_LIMIT must be an integer, but got '{raw_limit}'"
+        )
+
+    if limit <= 0 or limit >= len(docs):
+        return docs
+
+    print(f'[benchmark] BENCHMARK_EXTRACT_DOC_LIMIT set: extracting first {limit} of {len(docs)} documents')
```
noel-improv: Can we confirm the multi-hop tests still have both hops under a cap? The BlueBell Toys evidence is planted in doc 0 (UK orders/demand) and doc 47 (BlueBell/XFast distribution) of 70, so any limit below 48 drops the second hop. `MultiHopQuery` and `ChunkBasedSemanticSearchMultiHopQuery` only assert `assertGreater(num_source_nodes, 0)`, which a top-k satisfies from any chunk — so a capped run would go green on a query that is no longer multi-hop. Same misleading-pass case the small-batch skip guards against; worth skipping those two when the limit is set, or keeping both docs in the slice?

integration-tests/test-scripts/graphrag_toolkit_tests/batch_extract.py
```diff
@@ -215,14 +257,32 @@ class BatchExtractSmallBatchToS3(IntegrationTestBase):
...
+
+        # This test's only unique value is as the under-fill baseline for the
+        # auto-tune comparison, which needs the small batch_size to produce more
+        # than one (under-filled) batch job. A benchmark cap at or below the
+        # batch_size yields a single partial job — indistinguishable from what
+        # auto-tuning would produce — so there is no comparison signal. Skip
+        # (rather than report a misleading pass); correctness of the small-batch
+        # path is already covered by BatchExtractToS3 over the same corpus.
+        doc_limit = os.environ.get('BENCHMARK_EXTRACT_DOC_LIMIT', '').strip()
+        if doc_limit.isdigit() and 0 < int(doc_limit) <= extraction_batch_size:
```
noel-improv: Can we check this parses the same way as `apply_extraction_doc_limit`? The helper uses `int(raw_limit)`, this uses `isdigit()`. `+5` fails `isdigit()` but is a valid int, so it slips past the guard and runs 5 docs against `batch_size=10` — the single under-filled job the guard exists to avoid, reported as a pass. Going the other way, `isdigit()` is true for things like `²` where `int()` raises, so `int(doc_limit)` on the next line can throw. Would one shared parse returning the int (or `None`) cover both?

benchmarks/env.template
```diff
@@ -21,6 +21,7 @@ export DB_PASSWORD=<OPTIONAL: Database password – used by environments that re
 export BENCHMARK_DATA_DIR=<OPTIONAL: Local directory containing benchmark data to upload with deployment, default: Empty>
 export BENCHMARK_DATA_S3_URI=<OPTIONAL: S3 URI for benchmark data synced at runtime, e.g. s3://graphrag-toolkit-benchmarking/, default: Empty>
 export BENCHMARK_QA_LIMIT=<OPTIONAL: Max number of QA pairs to evaluate for prototype runs, default: Empty (all pairs)>
+export BENCHMARK_EXTRACT_DOC_LIMIT=<OPTIONAL: Cap the number of source documents extracted by the batch_extract.py tests (only the first N docs are extracted). Speeds up benchmark runs. default: Empty (all docs)>
```
noel-improv: Can we confirm this has an effect on the benchmarks path? `benchmark_extract.py:151` loads its corpus with `SimpleDirectoryReader(...).load_data()` and never calls `apply_extraction_doc_limit`, and nothing under `benchmarks/` reads the var — so setting it for a benchmark run looks like a silent no-op on the step you were trying to shorten. Honoring it there wouldn't be pure plumbing either, since `expected_docs` is hardcoded at line 187. Scope the entry to the integration-tests suites instead?

integration-tests/build-tests.sh
```diff
@@ -184,6 +184,7 @@ while [[ "$#" -gt 0 ]]; do
         --benchmark-data-dir) BENCHMARK_DATA_DIR="$2"; shift ;;
         --benchmark-data-s3-uri) BENCHMARK_DATA_S3_URI="$2"; shift ;;
         --benchmark-qa-limit) BENCHMARK_QA_LIMIT="$2"; shift ;;
+        --benchmark-extract-doc-limit) BENCHMARK_EXTRACT_DOC_LIMIT="$2"; shift ;;
```
noel-improv: Should this also go in the `--help` usage block? Every sibling flag is listed there, `--benchmark-qa-limit` included, but this one isn't, so `build-tests.sh --help` won't mention it.

integration-tests/test-scripts/graphrag_toolkit_tests/batch_extract.py
```diff
@@ -28,11 +28,45 @@ def get_metadata(data):
...
+
+    print(f'[benchmark] BENCHMARK_EXTRACT_DOC_LIMIT set: extracting first {limit} of {len(docs)} documents')
+    return docs[:limit]
+
 class BatchExtractToS3(IntegrationTestBase):
     
     @property
     def description(self):
-        return 'Extract propositions and topics from Neptune documentation using Bedrock batch inference, and save to S3'
+        return ('Baseline batch extraction: extract propositions and topics from the local Neptune docs '
```
noel-improv: Nit — is "Neptune docs corpus" right for this file? `corpus-modified.json` is 70 news articles (categories `business|entertainment|health|sports|technology`, metadata `author`/`source`/`published_at`), which is why the tenant is `multihop` and the queries are about BlueBell Toys. `init_with_test_details` copies the description into every result JSON uploaded to S3, and the auto-tune and small-batch descriptions inherit it via "the same local corpus".

integration-tests/test-scripts/graphrag_toolkit_tests/batch_extract.py
```diff
@@ -28,11 +28,45 @@ def get_metadata(data):
...
+    except ValueError:
+        raise ValueError(
+            f"BENCHMARK_EXTRACT_DOC_LIMIT must be an integer, but got '{raw_limit}'"
+        )
+
+    if limit <= 0 or limit >= len(docs):
+        return docs
+
+    print(f'[benchmark] BENCHMARK_EXTRACT_DOC_LIMIT set: extracting first {limit} of {len(docs)} documents')
+    return docs[:limit]
```
noel-improv, line 61: Can we confirm the multi-hop tests survive a cap? BlueBell evidence is in docs 0 and 47 of 70, so a limit under 48 loses the second hop, but both queries only assert `num_source_nodes > 0`, so they'd still pass.

integration-tests/build-tests.sh
noel-improv, line 188: Add this to the `--help` block too? Every sibling flag is listed there, `--benchmark-qa-limit` included.

integration-tests/test-scripts/graphrag_toolkit_tests/batch_extract.py
noel-improv: Nit — "Neptune docs corpus"? `corpus-modified.json` is 70 news articles (`business|entertainment|sports|...`), which is why the queries are about BlueBell Toys.

integration-tests/lexical.long
```diff
@@ -1,4 +1,3 @@
-batch_extract.BatchExtractToS3
 batch_extract.BatchExtractAutoTuneToS3
```
noel-improv, line 1: With `BatchExtractToS3` dropped here, is any suite still running it? Fixed-batch extraction looks uncovered now.

noel-improv (commented):
thanks for addressing the comments just a few more that might need addressing

integration-tests/test-scripts/graphrag_toolkit_tests/batch_extract.py
acarbonetto: oh - they probably not be useful results.  the limit/cap feature is really only to dry-run the results.

integration-tests/lexical.long
acarbonetto: that's fine.  Oussama and I want the have hte option available for debugging, but it doesn't need to be added to a suite.

noel-improv (approved):

oussamahansal (approved):
LGTM

PR #490, opened 2026-08-19
[FIX] Make the extract benchmark's inference mode and worker count explicit

noel-improv:
## Description

`run_benchmark_extract` defaulted `use_batch` to `True`, and `WikihowBenchmarkExtract` and `PgaBenchmarkExtract` never passed it, so both ran Bedrock batch inference whether or not that was wanted. `ConcurrentQaBenchmarkExtract` does pass it, which is what makes this look like an omission rather than a decision.

## Changes

- Drop the `use_batch` default so every caller states its inference mode
- Wikihow and Pga read `BENCHMARK_USE_BATCH`, required rather than defaulted
- `extraction_batch_size` and `extraction_num_workers` read the environment, keeping 15000 and 2 as defaults
- `build-tests.sh` propagates all three into `.env.testing`, without which the change is inert
- New `benchmarks/utils/benchmark_env.py` with `env_bool` and `env_int`, both treating an empty string as absent
- Extract `apply_extraction_config()` and `build_indexing_config()` from `run_benchmark_extract` so the config logic is reachable without running a full benchmark

## Problem

Batch inference is queue-scheduled, so wall time reflects queue wait rather than pipeline throughput. Two runs on byte-identical config (5000 docs, 2 workers x 4 threads, `use_batch` true, s3 doc store) finished at 22.5h and 3.96h. The slow run overlapped two other stacks submitting jobs; the fast one had the queue to itself. Two on-demand runs of the same corpus landed within 0.2% of each other, so the instability is specific to batch. Any timing produced by a benchmark that silently chose batch is unusable.

Separately, lines 97-98 hardcoded `extraction_batch_size` and `extraction_num_workers`, overriding the environment variables `GraphRAGConfig` already reads. A worker sweep could set `EXTRACTION_NUM_WORKERS`, pass, and report `workers=2` at every point on the curve. With the hardcode removed, worker scaling measured 21.0, 26.9, 31.7 and 33.2 docs/min at 2, 4, 8 and 16 workers, so eight times the workers buys 1.58x and the curve flattens near 33.

Related issue (if any): #

## Testing

43 tests across two files. `test_benchmark_extract_config.py` parses `benchmark_extract.py` with `ast` rather than importing it, since `benchmarks/tests` runs without `graphrag_toolkit` installed. `test_benchmark_extract_wiring.py` calls the two config helpers against the real `GraphRAGConfig` and skips via `importorskip` unless lexical-graph is present.

Mutation-tested. Reverting the signature default, dropping `use_batch` at a call site, adding a literal or named default to the env read, re-hardcoding 15000, reverting to a positional argument, and mistyping the variable name each fail at least one test. The `ast` tests cannot see whether `use_batch` still does anything, which is what the wiring tests cover: deleting the `if not use_batch: return None` guard leaves all 37 structural tests passing and fails two wiring tests.

Writing the wiring tests surfaced an ordering problem. `apply_extraction_config` assigned `extraction_llm` before `aws_region`, and that assignment builds a `BedrockConverse` which fails outright if no region is configured. Region now goes first.

Run the wiring tests with `uv run --with pytest --with-editable ./lexical-graph python -m pytest benchmarks/tests/test_benchmark_extract_wiring.py`.

## Checklist

`BENCHMARK_USE_BATCH` is now required, so a path that reaches `_run_test` without it fails at startup. `build-tests.sh` writes it unconditionally as `${BENCHMARK_USE_BATCH:-true}`, so behaviour through the harness is unchanged and batch remains the standing default. What the change buys is a greppable line in `.env.testing` recording the choice, and a loud failure on any path that bypasses the harness. The implicit default is made visible, not removed.

One consequence worth calling out: a notebook deployed by an older `build-tests.sh` has an `.env.testing` without the variable, so re-running extract on an existing stack fails after the stack is already up.


benchmarks/scripts/benchmark_extract.py
```diff
@@ -66,7 +114,9 @@ def run_benchmark_extract(handler: IntegrationTestHandler,
         dataset_name: Dataset key (e.g. 'concurrentqa', 'wikihow', 'pga').
         data_dir: Root path to the benchmark data directory.
         expected_docs: Expected number of source documents (for assertion).
-        use_batch: Whether to use Bedrock batch inference (default: True).
+        use_batch: Whether to use Bedrock batch inference. No default -
```
acarbonetto: I don't think you need to do anything except remove the `(default: True)`.  This is an implementation note, and doesn't need to be defined in the API.

benchmarks/scripts/benchmark_extract.py
```diff
@@ -45,11 +46,58 @@ def _count_source_docs(extracted_docs) -> int:
     return sum(len(page.get('CommonPrefixes', [])) for page in pages)
 
 
+def apply_extraction_config():
+    """
+    Apply the benchmark's extraction settings to GraphRAGConfig from the
+    environment.
+
+    extraction_batch_size and extraction_num_workers used to be hardcoded here,
```
acarbonetto: this comment is not needed - we don't need to be this verbose

benchmarks/scripts/benchmark_extract.py
```diff
@@ -45,11 +46,58 @@ def _count_source_docs(extracted_docs) -> int:
...
+        GraphRAGConfig.aws_region = aws_region
+
+    GraphRAGConfig.extraction_llm = os.environ.get(
+        'TEST_EXTRACTION_LLM', 'us.anthropic.claude-sonnet-4-6'
+    )
+    GraphRAGConfig.extraction_batch_size = env_int('EXTRACTION_BATCH_SIZE', 15000)
+    GraphRAGConfig.extraction_num_workers = env_int('EXTRACTION_NUM_WORKERS', 2)
+
+
+def build_indexing_config(use_batch: bool, dataset_name: str) -> Optional[IndexingConfig]:
```
acarbonetto: remove `use_batch`

if the calling procedure knows that use_batch is false, it doens't need to call this method

acarbonetto: something like
```
indexing_config = if use_batch build_indexing_config(dataset_name) else None
```

integration-tests/build-tests.sh
```diff
@@ -414,6 +414,15 @@ if [[ "$BENCHMARK_S3_JSONL" ]]; then
 	echo "export BENCHMARK_S3_JSONL=$BENCHMARK_S3_JSONL" >> lexical-graph-examples/.env.testing
 fi
 
+# Unconditional: the benchmarks require this rather than defaulting it.
```
acarbonetto: huh?  this means we are defined the default (True) here?

integration-tests/env.template
```diff
@@ -26,6 +26,9 @@ export BENCHMARK_DATASET=<OPTIONAL: Dataset to evaluate. Supports split variants
 export BENCHMARK_ALL_RETRIEVERS=<OPTIONAL: When "true", run all retrievers in a single pass (extract+build once, then query+evaluate per retriever). Incompatible with DELETE_ON_PASS=True, default: Empty (false)>
 export BENCHMARK_DOC_STORE=<OPTIONAL: Where extracted documents are staged between extract and build, file or s3, default: Empty (file)>
 export BENCHMARK_S3_JSONL=<OPTIONAL: With BENCHMARK_DOC_STORE=s3, store one JSONL object per source document instead of one object per chunk, true or false, default: Empty (false)>
+export BENCHMARK_USE_BATCH=<OPTIONAL: Inference mode for the extract benchmarks, true (Bedrock batch) or false (on demand), default: true. Batch is queue-scheduled, so its wall time reflects queue wait rather than pipeline throughput - set false for any run measuring workers or threads>
+export EXTRACTION_NUM_WORKERS=<OPTIONAL: Extraction worker processes, default: Empty (2)>
```
acarbonetto:
```suggestion
export EXTRACTION_NUM_WORKERS=<OPTIONAL: Extraction worker processes, default: 2>
```

integration-tests/env.template
```diff
@@ -26,6 +26,9 @@ export BENCHMARK_DATASET=<OPTIONAL: Dataset to evaluate. Supports split variants
...
+export EXTRACTION_BATCH_SIZE=<OPTIONAL: Extraction batch size, default: Empty (15000)>
```
acarbonetto:
```suggestion
export EXTRACTION_BATCH_SIZE=<OPTIONAL: Extraction batch size, default: 15000>
```

integration-tests/env.template
```diff
@@ -26,6 +26,9 @@ export BENCHMARK_DATASET=<OPTIONAL: Dataset to evaluate. Supports split variants
 export BENCHMARK_ALL_RETRIEVERS=<OPTIONAL: When "true", run all retrievers in a single pass (extract+build once, then query+evaluate per retriever). Incompatible with DELETE_ON_PASS=True, default: Empty (false)>
 export BENCHMARK_DOC_STORE=<OPTIONAL: Where extracted documents are staged between extract and build, file or s3, default: Empty (file)>
 export BENCHMARK_S3_JSONL=<OPTIONAL: With BENCHMARK_DOC_STORE=s3, store one JSONL object per source document instead of one object per chunk, true or false, default: Empty (false)>
+export BENCHMARK_USE_BATCH=<OPTIONAL: Inference mode for the extract benchmarks, true (Bedrock batch) or false (on demand), default: true. Batch is queue-scheduled, so its wall time reflects queue wait rather than pipeline throughput - set false for any run measuring workers or threads>
```
acarbonetto:
```suggestion
export BENCHMARK_USE_BATCH=<OPTIONAL: Inference mode for the extract benchmarks, true (Bedrock batch) or false (on demand), default: true>
```
This should be defined in a readme

acarbonetto (commented):
Other places env_bool / env_int should be used: 
  - benchmarks/scripts/benchmark_query.py:276-277 — int(os.environ.get('AGENTIC_MAX_ITERATIONS','3')) and BYOKG_MAX_ITERATIONS, repeated at lines 312-313, 346-347, 381-382. Not a live crash today (those vars aren't written by build-tests.sh), but if either gets added to the allowlist the way EXTRACTION_* were, an unset var arrives as '' and crashes at startup.
  - benchmark_query.py:286 — qa_limit = int(limit_str) if limit_str else None hand-reimplements env_int's empty-means-absent logic; env_int('BENCHMARK_QA_LIMIT', None) reproduces it exactly.
  - benchmark_extract.py:125 (the file the PR itself edits) — BENCHMARK_S3_JSONL still read as ...lower()=='true' right next to a new env_bool call, leaving two idioms side by side.
  - benchmark_extract.py:186 — is_prototype == 'true' is case-sensitive; setting BENCHMARK_IS_PROTOTYPE=True silently takes the non-prototype branch (full 13501-doc batch run instead of the 2-doc prototype). Same idiom recurs in benchmark_build.py:134,152, benchmark_evaluate.py:125-126,147-148, benchmark_query.py:273-274,309-310.
  - integration-tests/.../query_gpu.py:34 — os.environ.get('USE_GPU','False') == 'False' — setting USE_GPU=false (lowercase) doesn't match, so the GPU path runs when the user meant to disable it.

Also consider adding `env_string` for TEST_EXTRACTION_LLM / TEST_RESPONSE_LLM / BENCHMARK_JUDGE_LLM. 

Also note: 
  The `apply_extraction_doc_limit` hand-rolls env_int's semantics — but it lives in the integration-tests package, not benchmarks. Reusing benchmark_env.py there would first require promoting the helper to a shared location both packages import.

benchmarks/scripts/benchmark_extract.py
noel-improv: removed

benchmarks/scripts/benchmark_extract.py
noel-improv: cut

benchmarks/scripts/benchmark_extract.py
noel-improv: Agreed, that reads better.

integration-tests/build-tests.sh
noel-improv: The Python side requires `BENCHMARK_USE_BATCH` rather than defaulting it, so something has to state the value. This line is that one place, and it is greppable in `.env.testing` afterwards. The behaviour through `build-tests.sh` is unchanged and batch is still what you get by default. the default is now visible in one line instead of invisible in a function signature and any path that bypasses the harness fails loudly.
I can drop the `:-true` and make it a hard failure everywhere if you would rather, but that breaks every existing caller.

integration-tests/env.template
noel-improv: done

integration-tests/env.template
noel-improv: done

integration-tests/env.template
noel-improv: done

integration-tests/build-tests.sh
```diff
@@ -419,6 +419,15 @@ if [[ "$BENCHMARK_S3_JSONL" ]]; then
...
```
acarbonetto:
```suggestion
# Benchmarks require this.  Defaulting to true.  
```

integration-tests/build-tests.sh
acarbonetto: No its fine.  I didn't realize this was a duplicate comment.  I was confused by the "unconditional" comment.

acarbonetto (approved):

PR #405, opened 2026-07-16
[FEATURE] Add RDF/SPARQL graph store with optional Neptune IAM authentication

KilianTrunk:
## Description

This PR adds a new `lexical-graph-contrib/sparql` package for storing and retrieving the lexical graph through SPARQL 1.1 query and update endpoints.

The package uses RDFLib and supports generic HTTP endpoints. It also provides optional IAM authentication for the Amazon Neptune SPARQL endpoint.

Lexical graph builders and retrievers now pass a `GraphOperation` value together with the existing graph query. The SPARQL graph store uses this value to select its native SPARQL implementation. Existing graph stores continue to execute their existing queries.

## Changes

- Add the `graphrag-toolkit-lexical-graph-sparql` contributor package.
- Add RDFLib and Requests as package dependencies.
- Add Botocore as an optional dependency for Neptune IAM authentication.
- Add a graph store factory for `sparql`, `sparql+http`, `sparql+https`, and `sparql+neptune` connection schemes.
- Support separate SPARQL query and update endpoints.
- Support HTTP Basic authentication, custom HTTP headers, and request timeouts for generic endpoints.
- Add native SPARQL implementations for all 32 graph operations introduced by this PR: 17 update operations and 15 query operations.
- Represent lexical graph nodes with deterministic RDF IRIs and configurable namespaces.
- Represent facts as RDF resources connected to their subject, predicate, object, and supporting statement.
- Store each tenant in a deterministic named graph.
- Sign Neptune SPARQL requests with AWS Signature Version 4 for the `neptune-db` service.
- Resolve Neptune credentials through Botocore for every request so temporary credentials can be refreshed.
- Pass `GraphOperation` through retries, batched writes, query trees, and multi-tenant graph stores.
- Keep the existing query execution as the default for graph stores that do not provide an operation-specific implementation.
- Add SPARQL unit tests and extend core tests for operation forwarding and graph store factory selection.
- Add a contributor README and an RDF/SPARQL page to the documentation site.
- Rename the local development Compose project from `local-dev` to `graphrag-toolkit-rdf-dev`.
- Run the Jupyter reader dependency installation as `root`, apply `fix-permissions` to the Conda and notebook-user directories, and then return to the `jovyan` user.

## Problem

The toolkit did not provide a `GraphStore` implementation for SPARQL query and update endpoints.

[...]

PR #491, opened 2026-08-20
[FEATURE] Add backend-neutral GraphQueryOperation support

KilianTrunk:
## Description

This PR adds backend-neutral, typed graph operation identifiers to `lexical-graph`.

Graph builders and retrievers continue to construct their existing native queries, but now also pass an optional `GraphQueryOperation` and structured parameters. This allows graph stores using another query language to select an equivalent native implementation without parsing openCypher.

This is the core-only split requested during the review of #405. It does not include the SPARQL contributor package or development-environment changes.

## Changes

- Add and publicly export the `GraphQueryOperation` enum with 32 lexical-graph read and write operations.
- Add an optional `operation` argument to `GraphStore.execute_query()` and `GraphStore.execute_query_with_retry()`.
- Add the `_execute_operation()` dispatch hook for graph stores that provide operation-specific implementations.
- Preserve existing graph-store behavior through the default `_execute_operation()` implementation, which delegates to `_execute_query()`.
- Propagate operation context through retries, `Query`, `QueryTree`, `Job`, `GraphBatchClient`, and `MultiTenantGraphStore`.
- Forward tenant context to operation-specific implementations while preserving the existing native-query calling convention.
- Assign semantic operations across graph builders, the in-graph chunk store, retrievers, and query-context providers.
- Forward backend-neutral parameters required to preserve existing behavior, including entity classifications, chunk relationship types, local-entity literal values, and `includeChunkMetadata`.
- Preserve the existing public `include_chunk_details` option while exposing its value to operation-specific backends as `includeChunkMetadata`.
- Simplify correlation-ID handling so an omitted or empty value produces only the generated ID, while a supplied value produces `<supplied>/<generated>`.
- Add and update unit tests for operation dispatch, fallback behavior, batching, query trees, tenant propagation, chunk-store forwarding, and correlation-ID generation.

## Problem

The existing `GraphStore` API only exposed a native query string and its parameters.

That works for graph stores using the same query language, but an alternative backend cannot reliably determine the semantic intent of generated openCypher without parsing it.

This change adds an optional semantic operation contract while preserving the existing execution path for current graph stores.

Related issue (if any): N/A

## Testing

- [ ] Integration tests added (existing integration tests and targeted E2E scenarios were used)
- [x] Tested manually as described below

Validation performed:

- Ran the complete test suite on Python 3.10, 3.11, and 3.12:
  - 2,002 passed
  - 15 skipped
  - 0 failed
  - 64.31% coverage against the required 56%
- Repeated the Python 3.12 suite in a clean, CI-equivalent environment:
  - 2,002 passed
  - 15 skipped
  - 0 failed
- Ran the existing integration suite against a live Neo4j 5 container:
  - 8 passed
  - 7 AWS-dependent tests skipped
  - 0 failed
- Ran an end-to-end lexical-graph build against live Neo4j, exercising all graph builders modified by this PR.
- Ran the same build against the parent commit and confirmed identical graph node and relationship counts.
- Ran end-to-end traversal-based retrieval against live Neo4j and pgvector and confirmed that results were returned.
- Ran a multi-tenant build and traversal retrieval and confirmed that tenant-scoped labels and queries worked correctly.
- Specifically exercised the modified `QueryTree`, `Job`, `GraphBatchClient`, and `MultiTenantGraphStore` forwarding paths.
- Verified that existing graph stores preserve their previous behavior through the default `_execute_operation()` fallback.
- Ran license-header validation: 0 files missing headers.
- `git diff --check upstream/main..HEAD`: no errors.
- Merge-tree simulation against the current `upstream/main`: no conflicts.

## Checklist

- [x] Documentation updated (not applicable to this core-only change; user-facing SPARQL documentation is included in the follow-up contributor PR)
- [x] No breaking changes

`GraphQueryOperation` is optional. Existing graph stores that do not implement operation-specific execution continue to execute their existing native queries through `_execute_query()`.

---


acarbonetto:
Thank you for this.  I will take a look at this asap.

lexical-graph/src/graphrag_toolkit/lexical_graph/storage/graph/graph_store.py
```diff
@@ -496,22 +500,19 @@ def property_assigment_fn(self, key:str, value:Any) -> Callable[[str], str]:
...
+            parameters=parameters,
+            max_attempts=1,
+            max_wait=0,
+            correlation_id=correlation_id,
+            operation=operation,
+        )
+
+    def _execute_operation(self, operation:GraphQueryOperation, query:str, parameters:Dict[str, Any], correlation_id=None, **kwargs) -> List[Any]:
+        """Execute the existing native query for stores without an override."""
+        return self._execute_query(query, parameters, correlation_id=correlation_id)
```
acarbonetto: `operation` not used

acarbonetto: was the intention to make this an `abstractmethod`?

lexical-graph/src/graphrag_toolkit/lexical_graph/storage/graph/graph_store.py
```diff
@@ -390,7 +391,7 @@ def __exit__(self, exception_type, exception_value, traceback):
     def unretriable_exception_types(self) -> Tuple:
         return ()
 
-    def execute_query_with_retry(self, query:str, parameters:Dict[str, Any], max_attempts=3, max_wait=5, **kwargs) -> Dict[str, Any]:
+    def execute_query_with_retry(self, query:str, parameters:Dict[str, Any], max_attempts=3, max_wait=5, operation:Optional[GraphQueryOperation]=None, **kwargs) -> Dict[str, Any]:
```
acarbonetto: ### Suggestion: replace the operation/hints kwargs with a per-store query_handlers() + _dispatch(), so GraphStore stays query-language-neutral

  The operation enum + RDF hint kwargs (_predicate/_subject_literal/_object_literal) are the right goal — giving a store that can't consume openCypher a neutral way to express intent — but as written they land on the shared base signature, which couples the neutral layer to RDF-shaped thinking and leaves the caller passing three overlapping representations (the openCypher string, the operation, and the hints) with no single source of truth. It also means the shipped Cypher store carries operation/hint params it ignores.

  I think we can keep this PR's SPARQL scope intact and leave the Cypher path completely undisturbed by folding all three into one type and letting each store declare what it accepts.

#### 1. Bundle the representations into one object (instead of parallel `kwargs`):

```
  @dataclass
  class GraphOperation:
      """A query as a backend-neutral operation plus its authoritative openCypher
      rendering. Property-graph stores execute `.cypher`; RDF/SPARQL stores render
      natively from `.operation` and `.hints`."""
      operation: GraphQueryOperation
      cypher: str
      parameters: Dict[str, Any] = field(default_factory=dict)
      hints: Dict[str, Any]      = field(default_factory=dict)   # predicate, subject_literal, ...
```

#### 2. Let each store own its accepted query types — replace the module-level `QUERY_TYPE` union (line 22, also referenced by `format_log_entry` at 142/171/205 and execute_query at 499) and the hardcoded `isinstance` ladder (~437) with a per-class handler map:

```
  def query_handlers(self) -> Dict[type, Callable[..., List[Any]]]:
      # Property-graph default. Override to declare a different accepted set.
      return {
          str:            lambda q, p, **k: self._execute_query(q, p, **k),
          QueryTree:      lambda q, p, **k: q.run(p, self.execute_query_with_retry),
          GraphOperation: lambda q, p, **k: self._execute_operation(q, p, **k),
      }

  def _dispatch(self, query, parameters, **kwargs):
      handlers = self.query_handlers()
      for qtype, handler in handlers.items():
          if isinstance(query, qtype):
              return handler(query, parameters, **kwargs)
      accepted = ', '.join(t.__name__ for t in handlers)
      raise ValueError(
          f'{type(self).__name__} does not accept queries of type '
          f'{type(query).__name__}; accepts: {accepted}'
      )
```

`execute_query_with_retry` just calls `self._dispatch(...)` inside the retry loop.

#### 3. The default _execute_operation reuses the authoritative string, so the Cypher store needs zero new per-operation code — it keeps making query-string calls exactly as today:

```
  def _execute_operation(self, op: GraphOperation, parameters={}, **kwargs):
      # Property-graph default: run the authoritative openCypher rendering.
      return self._execute_query(op.cypher, {**op.parameters, **parameters}, **kwargs)
```

#### 4. Move all SPARQL-specific handling into the RDF store — it's the only new implementer. It narrows the accepted set (dropping str, so a stray Cypher string gets a clear error) and consumes operation/hints:

```
  class SparqlGraphStore(GraphStore):
      def query_handlers(self):
          return {GraphOperation: lambda q, p, **k: self._execute_operation(q, p, **k)}

      def _execute_operation(self, op, parameters={}, **kwargs):
          sparql = self._render_sparql(op.operation, op.hints, {**op.parameters, **parameters})
          return self._run_sparql(sparql, **kwargs)
```

### Net effect on this PR:
  - The operation= kwarg and the RDF hint kwargs come off the shared base signature — they live inside GraphOperation and are consumed only by the RDF store.
  - The Cypher store is untouched: it runs op.cypher (or a bare string) via the default handler.
  - GraphQueryOperation and the hints move next to the SPARQL store that uses them.
  - QUERY_TYPE becomes set(self.query_handlers()), owned per store rather than a global union the base must know about. Keep a QUERY_TYPE = Union[str, QueryTree, GraphOperation] alias for the format_log_entry/execute_query annotations
    so this stays contained.
    
Same SPARQL capability, but the neutral layer never names a query language and the caller passes one self-documenting object instead of three loose parallel args.

mykola-pereyma (changes requested):
Really solid, well-tested feature — the abstraction is cleanly additive and backward compatibility holds (optional `operation`, default `_execute_operation()` delegating to `_execute_query()`, native queries/params unchanged for the default backend, and the E2E identical node/relationship counts vs the parent commit are convincing).

I did a per-assignment semantic audit of all 32 operations against the queries they tag. 31 are assigned and accurate — including the easy-to-swap `FIND_COMPLEMENTS` / `FIND_SUBJECTS` pair (verified correct against the query comments and the `get-complements-for-subject` / `get-real-subjects-for-complement` tree names), and `entity_vss_provider`'s topic/chunk branch (`FIND_ENTITIES_BY_TOPICS` / `FIND_ENTITIES_BY_CHUNKS`).

Requesting changes on one point before approval, since the enum is now a **public, exported API contract**:

### `SEARCH_BY_TOPIC` is defined but never assigned

`GraphQueryOperation.SEARCH_BY_TOPIC` has no call site anywhere in the PR. Every other value is wired to a real query and exercised; this one is inert. Shipping an unused member in a public contract is a maintenance hazard — a future backend implementer will assume every value has a well-defined, exercised meaning.

Please do one of the following:

1. **Assign it** if a topic-search site was missed, and add a unit test that dispatches through it (mirroring the operation-dispatch tests you already added). Candidate to double-check: `keyword_vss_provider` currently tags `get_statements_for_topic(topic_id)` with `GET_TOPIC`. That reads as a post-VSS fetch-by-id (returns statements for an already-resolved topic id), so `GET_TOPIC` is defensible — but if the intent was the topic *search* entry point, `SEARCH_BY_TOPIC` may belong there. Please confirm which is correct.
2. **Document it as an intentional placeholder** — a short comment on the enum member explaining the future/alternative-backend use case it's reserved for, so it's clearly deliberate rather than an oversight.
3. **Remove it** to keep the contract minimal until there's a consumer.

Everything else looks good; happy to approve once `SEARCH_BY_TOPIC` is resolved (assigned + tested, documented, or removed).

KilianTrunk:
> Really solid, well-tested feature — the abstraction is cleanly additive and backward compatibility holds (optional `operation`, default `_execute_operation()` delegating to `_execute_query()`, native queries/params unchanged for the default backend, and the E2E identical node/relationship counts vs the parent commit are convincing).
> 
> I did a per-assignment semantic audit of all 32 operations against the queries they tag. 31 are assigned and accurate — including the easy-to-swap `FIND_COMPLEMENTS` / `FIND_SUBJECTS` pair (verified correct against the query comments and the `get-complements-for-subject` / `get-real-subjects-for-complement` tree names), and `entity_vss_provider`'s topic/chunk branch (`FIND_ENTITIES_BY_TOPICS` / `FIND_ENTITIES_BY_CHUNKS`).
> 
> Requesting changes on one point before approval, since the enum is now a **public, exported API contract**:
> 
> ### `SEARCH_BY_TOPIC` is defined but never assigned
> `GraphQueryOperation.SEARCH_BY_TOPIC` has no call site anywhere in the PR. Every other value is wired to a real query and exercised; this one is inert. Shipping an unused member in a public contract is a maintenance hazard — a future backend implementer will assume every value has a well-defined, exercised meaning.
> 
> Please do one of the following:
> 
> 1. **Assign it** if a topic-search site was missed, and add a unit test that dispatches through it (mirroring the operation-dispatch tests you already added). Candidate to double-check: `keyword_vss_provider` currently tags `get_statements_for_topic(topic_id)` with `GET_TOPIC`. That reads as a post-VSS fetch-by-id (returns statements for an already-resolved topic id), so `GET_TOPIC` is defensible — but if the intent was the topic _search_ entry point, `SEARCH_BY_TOPIC` may belong there. Please confirm which is correct.
> 2. **Document it as an intentional placeholder** — a short comment on the enum member explaining the future/alternative-backend use case it's reserved for, so it's clearly deliberate rather than an oversight.
> 3. **Remove it** to keep the contract minimal until there's a consumer.
> 
> Everything else looks good; happy to approve once `SEARCH_BY_TOPIC` is resolved (assigned + tested, documented, or removed).

Hi @mykola-pereyma ! 

First of all, thanks for the review!

Secondly, I sadly haven't received any email notifications about the review, so apologies for the late response :/ - however, GraphQueryOperation.SEARCH_BY_TOPIC was actually used in the following file: lexical-graph/src/graphrag_toolkit/lexical_graph/retrieval/retrievers/entity_network_search.py

Either way, I had to rebase it as we had some merge conflicts due to not merging this for a while.

Let me know if anything still looks odd, and thank you again!

PR #494, opened 2026-08-21
[FIX] Size the LLM call pool and the Bedrock connection pool to the thread count

noel-improv:
One concern in two layers: the executor that runs the LLM calls, and the client those calls go through.

## The thread ceiling

Extraction runs one asyncio job per node, each awaiting a synchronous LLM call. `asyncio.to_thread` hands those to the default executor, which CPython sizes at `min(32, cpu_count + 4)`. That cap, not `extraction_num_threads_per_worker`, decided how many calls were in flight.

500 docs, one worker, on-demand Sonnet 4.6, `ml.m5.4xlarge`, model-call phase only:

| Threads | Before | After | Speed-up |
|---|---|---|---|
| 16 | 21.9 min | 21.5 min | flat |
| 32 | 17.5 min | 11.0 min | 1.6x |
| 64 | 17.8 min | 5.8 min | 3.1x |

16 is flat because both settings sit under the cap of 20 on that instance.

The pool is floored at CPython's size, so a low thread setting cannot make things worse than before, and capped at 256. Every spawned process gets its own pool and every thread in it holds a bedrock-runtime connection, so an unreasonably high thread count would otherwise multiply across processes into thousands of sockets.

It is sized once, by whichever caller reaches it first, and never resized. Growing it would mean replacing it, and a replaced pool's threads keep running until their work drains, so the cap would bound one pool rather than the process. A later request for more logs once and uses the existing size. In the pipeline the extractors take the same worker count, so they agree.

## The connection pool

`llm_cache.py` built its bedrock-runtime client with no `max_pool_connections`, so it took botocore's default of 10 while the change above drives up to 64 concurrent calls through it. Past the pool botocore discards and reopens connections, costing back the concurrency.

The pool is sized `max(botocore default, num_threads * 2, MIN_POOL_SIZE)`. The first two terms match how `config.py` already sizes its own client. The floor is there because this client sits behind the LLM call pool and `pool_size()` reads 0 until something creates it, so a caller arriving first would otherwise get a pool of 10 while the executor runs at its own floor. Reading the config alone would be a no-op on the path that needs it, because extraction runs in a spawned process and a thread count set on `GraphRAGConfig` in the parent reads back in the worker as the default:

```
parent sets 64  ->  parent reads 64 | child reads env None, config 4
```

Known limitation: a count set programmatically above the CPython floor stays invisible to the client, so the pool can still sit under peak in-flight calls. #506 closes this by propagating a `GraphRAGConfig` snapshot into spawn workers. That PR targets v3.19.x and this one targets main, so the limitation stands until they converge.

## Security review

A review of this branch found no High or Medium issues, and no attacker-facing surface in the diff: no new input parsing, auth, endpoints, query construction or deserialization. Two Low findings, both resource management, both fixed here.

A caller-supplied thread count reached the bedrock client unclamped, so a request far above the pool's maximum sized the socket pool from a number the executor would never reach. It is capped now, and a negative count is rejected rather than absorbed by the floors.

The 256 ceiling bounded a single pool rather than the process, which the sizing change above fixes.

## Also

`llm_concurrency` moves from `indexing/utils/` to `utils/`, next to the cache that reads it. Three duplicated client-construction blocks and the cross-region tests' repeated patch and assert blocks collapse into helpers.

2018 tests pass. The one error in `test_integ_dependency_compatibility.py` is pre-existing on main.

Six files and about 170 production lines, over the five-file guideline. Splitting would put the executor fix on main still throttled by a pool of 10.


mykola-pereyma (changes requested):
The fix itself is well-diagnosed and correct — the real thread ceiling was CPython's default executor cap (`min(32, cpu+4)`), not `extraction_num_threads_per_worker`, and routing blocking LLM calls through a pool sized to the actual worker count is the right fix. The connection-pool sizing and the collapse of the three duplicated client-construction blocks are good. Tests are strong (the barrier only releases at the requested concurrency, so a pool stuck at the floor would time out). Requesting changes only for docstring conciseness and a couple of small consistency items.

### Docstrings: trim stale-prone benchmark detail

**`llm_concurrency.py` module docstring** — please remove the embedded timing measurement:

> Measured on a 14-core machine, where the default cap is 18: raising the job semaphore to 32 or 64 left peak in-flight pinned at 18, while running the same 64 jobs against a pool sized to the semaphore reached 64 in-flight and cut wall time from 4.9s to 2.6s.

A concrete wall-time number tied to a specific machine will go stale and duplicates the benchmark already in the PR description. Docstrings should state the contract; keep the numbers in the PR description or a BENCHMARKS note. The other three paragraphs (executor-cap ceiling, pool contract, spawn rationale) are exactly right — keep them.

**`_pool_size_for` docstring** — reduce to the one-line contract and drop the worked example:

```python
"""
How many workers a caller asking for `num_threads` gets: the requested
count, floored at MIN_POOL_SIZE. Nothing here caps the request.
"""
```

The "16 vCPU machine, where the floor is 20" walkthrough is filler, and it also uses a different example machine than the module docstring's "14-core" — if any worked example is retained, use a single consistent machine so the two numbers don't contradict.

### Small consistency item (non-blocking)

`config.py` already sizes another client pool as `max(BOTOCORE_DEFAULT_MAX_POOL_CONNECTIONS, num_threads * 2)` (line ~178), while the new `_bedrock_client` uses `max(default, threads, cpython_floor)` without the `* 2`. Two different heuristics for the same concern — worth aligning, or a comment noting why they differ.

### Note for later (not this PR)

The documented limitation — a programmatically-set thread count above the CPython floor stays invisible to `llm_cache` in the spawned worker — is exactly what PR #506 (GraphRAGConfig snapshot propagation to spawn workers) resolves. Worth linking them so the limitation is closed when they converge (#506 is on v3.19.x, this is on main).

Everything else LGTM — happy to approve once the docstrings are trimmed.

oussamahansal (approved):
LGTM

mykola-pereyma (approved):
LGTM — thanks for the thorough revision. All the docstring feedback is resolved (the timing measurement is out of the module docstring, `_pool_size_for` is a one-line contract, and the machine-specific inconsistency is gone), and both earlier suggestions landed too: `max_pool_connections` now follows config.py's `num_threads * 2` heuristic, and `llm_cache` sizes its connection pool from the live `pool_size()` — which nicely closes the spawned-worker limitation the original version documented.

The extra hardening is a real improvement: the `MAX_POOL_SIZE` clamp bounds pool growth, the per-call `contextvars.copy_context()` in `run_blocking` prevents a ContextVar leaking across pooled-thread reuse (good catch, and nicely covered by a red/green test), and the submit-under-lock guard against a concurrent `shutdown` is tested with an explicit race test.

Two non-blocking notes:
- **Please squash before merge** — 11 commits for this change; a squash keeps the history tidy.
- `batch_inference_utils.py` (which in the first revision routed its retry path through `run_blocking`) is no longer in the diff after the rebase, so that batch-API path still uses the default-capped `asyncio.to_thread`. The hot extraction paths use the sized pool, so this is fine for the PR's goal — just confirm the omission is intentional rather than a lost call site worth a follow-up.

Approving.

PR #495, opened 2026-08-21
[FIX]: Widen secret scanning to release branches

oussamahansal:
## Description

- The gitleaks workflow only ran on main, so PRs/pushes to release branches were never scanned per-commit

## Changes

- Widen secret scanning to vX.Y.x release branches

## Problem

Related issue (if any): #

## Testing

- [ ] Unit tests added/updated
- [ ] Integration tests added (as appropriate)
- [x] Existing tests pass (`pytest`)
- [ ] Tested manually (describe below)

## Checklist

- [x] Code follows existing style and conventions
- [ ] License headers present on new files
- [ ] Documentation updated (if applicable)
- [x] No breaking changes (or clearly documented)

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #496, opened 2026-08-21
[FIX]: Read coverage values via env

oussamahansal:
## Description

- coverage-percentage.txt is produced by the test job (which runs PR code) and was interpolated as ${{ steps.coverage.outputs.* }} inside a JS template literal in a job with pull-requests: write. Pass the values via env: and reference them as process.env.COVERAGE / process.env.TARGET so they can't break out of the script into executable JS

## Changes

- Read coverage values via env in the PR-comment github-script step

## Problem

Related issue (if any): #

## Testing

## Checklist

---

### DIFF
--- a/.github/workflows/lexical-graph-tests.yml
+++ b/.github/workflows/lexical-graph-tests.yml
@@ -157,11 +157,16 @@ jobs:
 
       - name: Comment on PR with coverage report link
         uses: actions/github-script@v9
+        env:
+          # Pass via env, not inline ${{ }}, so the values can't break out of the
+          # script into executable JS.
+          COVERAGE: ${{ steps.coverage.outputs.percentage }}
+          TARGET: ${{ steps.coverage.outputs.target }}
         with:
           script: |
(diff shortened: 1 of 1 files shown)


noel-improv (commented):
Fix is right, but it's landing where it can't run: this workflow triggers on `branches: [main]` for both `push` and `pull_request`, so it never fires on v3.19.x, and `origin/main` still has the inline `${{ steps.coverage.outputs.* }}` interpolation. Same as @mykola-pereyma 's [point](https://github.com/awslabs/graphrag-toolkit/pull/495#pullrequestreview-4996301359) on #495, this needs to go to main to actually protect anything. Def worth doing before the pre-release.

> Optional, one step earlier: `echo "percentage=${COVERAGE_PCT}" >> "$GITHUB_OUTPUT"` lets a newline in the value add an extra step output. Probably not reachable, the value's a Python `round(float(...))` and the comment job is same-repo only, but it's the next hop under the same threat model.

mykola-pereyma (approved):

PR #499, opened 2026-08-22
fix: batch domain-label inserts under one stable query per label

### LINKED ISSUE #477: [BUG] Domain-label inserts defeat batching/dedup — each insert becomes its own single-row batch
### Package version

3.19.0

### Package

lexical-graph

### Python version

3.12.0

### Operating System

macOS

### Description

Discovered while reviewing: #476   

## Description
  
  `EntityGraphBuilder.insert_domain_entity` (`lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/build/entity_graph_builder.py:123-128`) builds its Cypher with a fresh `new_query_var()` UUID and an `// awsqid:{e_id}-{e_label}` comment interpolated directly into the query text:

  ```python
  e_var = new_query_var()
  e_comment = f'// awsqid:{e_id}-{e_label}'.replace('\r', ' ').replace('\n', ' ')
  query_e = f"UNWIND $params AS params MERGE ({e_var}:`__Entity__`{{...: params.entityId}}) SET {e_var} :`{e_label}` {e_comment}"
```

`GraphBatchClient.execute_query_with_retry` keys its batch dictionary on the full query string (`graph_batch_client.py`:139-141: `self.batches[query].extend(...)`). Because both the `e_var` UUID and the `e_id`-bearing comment vary on every call, every domain-label insert produces a unique batch key holding a single param row. Batching and dedup are completely defeated for domain labels:

  - `self.batches` grows O(fact-entity occurrences) distinct full query strings in memory before flush.
  - The same entity's MERGE is re-issued once per fact it appears in — never deduped.

This path was previously unreachable in batch mode (it crashed with KeyError: 'params', fixed in #476), so the cost is newly live. The sibling `insert_for_entity` (:82-101) uses a constant query string and batches correctly — this method should do the same.

**Expected**: domain-label inserts collapse into a small number of label-keyed batches (one per distinct label) and dedup across params, matching insert_for_entity.
  
**Fix direction**: Cypher can't parameterize a label, so the label must stay inlined — but the var name and comment don't need to. Use a constant var name (e.g. entity), drop the now-vestigial // awsqid: comment (it's only consumed by `_add_parameterless_query`, which never runs for a query that carries params), so the query string is constant per label and batches/dedups.

### Steps to reproduce

```python
from graphrag_toolkit.lexical_graph.indexing.build.entity_graph_builder import EntityGraphBuilder
  from graphrag_toolkit.lexical_graph.indexing.build.graph_batch_client import GraphBatchClient

  class _Store:
      def node_id(self, name): return name
  
  client = GraphBatchClient(graph_client=_Store(), batch_writes_enabled=True, batch_write_size=100)

  # Ingest N facts referencing the same entity/classification via EntityGraphBuilder().build(..., include_domain_labels=True)
  # Expected: 1 label-keyed batch with N deduped param rows.
  # Actual:   N distinct query-string keys, each holding a single param row.
  print(len(client.batches))  # grows ~linearly with fact-entity occurrences
```

```shell
(none — this is a memory/throughput regression, not a crash)
```

yzxcj797:
Fixes #477.

## Root cause

`EntityGraphBuilder.insert_domain_entity` builds its Cypher with a fresh `new_query_var()` UUID and an `// awsqid:{e_id}-{e_label}` comment interpolated directly into the query text. `GraphBatchClient.execute_query_with_retry` groups param rows by the **full query string** (`self.batches[query].extend(...)`), so both the per-call variable name and the per-entity id made every domain-label insert a unique batch key holding a single param row — batching and dedup defeated, `self.batches` growing one full query string per fact-entity occurrence, one round trip per insert.

## Fix

The query text is now **stable per label**:

- a fixed `e` variable — the statement has exactly one MERGE, so the UUID collision-avoidance the helper provides isn't needed here;
- a label-scoped comment (`// awsqid:domain-label-{label}`) instead of the per-entity id — the id already travels in the params row. The comment keeps its newline/carriage-return stripping and the label stays escaped via `escape_cypher_label`, preserving the injection hardening from the earlier fix this test file covers.

Inserts for distinct entities of the same classification now share one batch entry (`UNWIND $params` MERGEs them in one statement), and different classifications group under their own query.

## Testing

Added `test_domain_entity_inserts_batch_under_one_query_per_label` next to the existing regression tests for this builder: three facts with distinct entity ids of the same classification must queue under exactly **one** domain query with all three id rows, instead of three single-row batches.

- With the fix: the three builder test files pass (10 tests).
- Differential with the builder reverted (tests kept): the new test fails — three separate single-row batches, exactly the reported behavior.

### DIFF
--- a/lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/build/entity_graph_builder.py
+++ b/lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/build/entity_graph_builder.py
@@ -120,11 +120,16 @@ class EntityGraphBuilder(GraphBuilder):
                     if entity.classification and entity.classification == LOCAL_ENTITY_CLASSIFICATION:
                         return
 
-                    e_var = new_query_var()
                     e_id = entity.entityId
                     e_label = escape_cypher_label(label_from(entity.classification or DEFAULT_CLASSIFICATION))
-                    e_comment = f'// awsqid:{e_id}-{e_label}'.replace('\r', ' ').replace('\n', ' ')
-                    query_e = f"UNWIND $params AS params MERGE ({e_var}:`__Entity__`{{{graph_client.node_id('entityId')}: params.entityId}}) SET {e_var} :`{e_label}` {e_comment}"
+                    # The query text must be stable for a given label: the batch
+                    # client groups param rows by the full query string, so a
+                    # per-entity variable name or a per-entity id embedded in the
+                    # text makes every insert its own single-row batch and defeats
+                    # batching and dedup entirely (#477). The entity id already
+                    # travels in the params; the comment stays label-scoped.
+                    e_comment = f'// awsqid:domain-label-{e_label}'.replace('\r', ' ').replace('\n', ' ')
+                    query_e = f"UNWIND $params AS params MERGE (e:`__Entity__`{{{graph_client.node_id('entityId')}: params.entityId}}) SET e :`{e_label}` {e_comment}"
                     graph_client.execute_query_with_retry(query_e, self._to_params({'entityId': e_id}), max_attempts=5, max_wait=7)
 
                 insert_domain_entity(fact.subject)
--- a/lexical-graph/tests/unit/indexing/build/test_entity_graph_builder_domain_label_batch.py
+++ b/lexical-graph/tests/unit/indexing/build/test_entity_graph_builder_domain_label_batch.py
@@ -83,3 +83,57 @@ def test_domain_entity_insert_uses_batch_params_shape():
     assert 'params.entityId' in query
     assert '$entityId' not in query
     assert params == [{'entityId': ENTITY_ID}]
+
+
+def _fact_node_multi(entity_ids):
+    subjects = [
+        {
+            'entityId': eid,
+            'value': f'Acme {i}',
+            'classification': 'Company',
+        }
+        for i, eid in enumerate(entity_ids)
+    ]
+    return [
+        TextNode(
+            text='x',
+            metadata={
+                'fact': {
+                    'factId': f'fact-{i}',
+                    'subject': subject,
+                    'predicate': {'value': 'operates'},
+                    'object': None,
+                }
+            },
+        )
+        for i, subject in enumerate(subjects)
+    ]
+
+
+def test_domain_entity_inserts_batch_under_one_query_per_label():
+    """Batch writes group param rows by the full query string. The query used
+    to embed a fresh variable name and the per-entity id in its text, so every
+    domain-label insert landed in its own single-row batch (#477). Inserts for
+    distinct entities of the same classification must share one query entry."""
+    batch_client = GraphBatchClient(
+        graph_client=_MockStore(),
+        batch_writes_enabled=True,
+        batch_write_size=100,
+    )
+
+    builder = EntityGraphBuilder()
+    for node in _fact_node_multi(['ent-1', 'ent-2', 'ent-3']):
+        builder.build(
+            node,
+            batch_client,
+            include_domain_labels=True,
+            include_local_entities=False,
+        )
+
+    domain_batches = {q: p for q, p in batch_client.batches.items() if 'awsqid' in q}
+    assert len(domain_batches) == 1, (
+        f'expected one shared domain-label query, got {len(domain_batches)}: '
+        f'{list(domain_batches)}'
+    )
+    query, params = next(iter(domain_batches.items()))
+    assert sorted(p['entityId'] for p in params) == ['ent-1', 'ent-2', 'ent-3']


mykola-pereyma:
Thank you @yzxcj797 for your contribution we will review it and respond shortly.

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/build/entity_graph_builder.py
```diff
@@ -120,11 +120,16 @@ def insert_domain_entity(entity:Entity):
...
-                    e_comment = f'// awsqid:{e_id}-{e_label}'.replace('\r', ' ').replace('\n', ' ')
-                    query_e = f"UNWIND $params AS params MERGE ({e_var}:`__Entity__`{{{graph_client.node_id('entityId')}: params.entityId}}) SET {e_var} :`{e_label}` {e_comment}"
+                    # The query text must be stable for a given label: the batch
+                    # client groups param rows by the full query string, so a
+                    # per-entity variable name or a per-entity id embedded in the
+                    # text makes every insert its own single-row batch and defeats
+                    # batching and dedup entirely (#477). The entity id already
+                    # travels in the params; the comment stays label-scoped.
+                    e_comment = f'// awsqid:domain-label-{e_label}'.replace('\r', ' ').replace('\n', ' ')
+                    query_e = f"UNWIND $params AS params MERGE (e:`__Entity__`{{{graph_client.node_id('entityId')}: params.entityId}}) SET e :`{e_label}` {e_comment}"
```
noel-improv, line 132: Confirmed the differential locally rather than reading it off the diff. Against current main's builder the new test fails with three separate query keys; on this branch both pass:

```
AssertionError: expected one shared domain-label query, got 3
```

Build test dir is 27 failed / 15 errors on both this branch and `origin/main`, so nothing here regressed.

`escape_cypher_label` and the `\r`/`\n` stripping both survive, and pulling `e_id` out of the query text leaves less inlined than before. No objection from the injection-hardening side.

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/build/entity_graph_builder.py
```diff
@@ -120,11 +120,16 @@ def insert_domain_entity(entity:Entity):
...
                     e_label = escape_cypher_label(label_from(entity.classification or DEFAULT_CLASSIFICATION))
-                    e_comment = f'// awsqid:{e_id}-{e_label}'.replace('\r', ' ').replace('\n', ' ')
-                    query_e = f"UNWIND $params AS params MERGE ({e_var}:`__Entity__`{{{graph_client.node_id('entityId')}: params.entityId}}) SET {e_var} :`{e_label}` {e_comment}"
+                    # The query text must be stable for a given label: the batch
+                    # client groups param rows by the full query string, so a
+                    # per-entity variable name or a per-entity id embedded in the
+                    # text makes every insert its own single-row batch and defeats
+                    # batching and dedup entirely (#477). The entity id already
+                    # travels in the params; the comment stays label-scoped.
+                    e_comment = f'// awsqid:domain-label-{e_label}'.replace('\r', ' ').replace('\n', ' ')
```
noel-improv, line 131: Dropping the per-entity id is safe. The only consumer of `awsqid` is `_add_parameterless_query` (`graph_batch_client.py:111`), reached only when `properties` is falsy — and since #476 this query always passes `_to_params(...)`, so it was already dead on this path.

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/build/entity_graph_builder.py
```diff
@@ -120,11 +120,16 @@ def insert_domain_entity(entity:Entity):
                     if entity.classification and entity.classification == LOCAL_ENTITY_CLASSIFICATION:
                         return
 
-                    e_var = new_query_var()
                     e_id = entity.entityId
```
noel-improv, line 123: Nit: `new_query_var` is now unused in this file (import on line 9). Worth dropping with it.

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/build/entity_graph_builder.py
noel-improv, line 1: Rebase note: this predates `dc6a096f` (downmerge #502), which already landed #476's `UNWIND $params` / `_to_params` conversion on main, hence the conflict. After a rebase the change reduces to the fixed `e` var plus the label-scoped comment. Main also carries a comment above this block calling the batching "a future optimization"; that should go with the fix.

noel-improv (changes requested):
thanks for your contribution @yzxcj797, please review these requested changes.

PR #507, opened 2026-08-25
[FIX] Overlap chunk uploads across documents and carry the thread count into workers

noel-improv:
## Description

`S3ChunkUploader` drained one source document's chunk uploads before submitting the next, so in-flight uploads were bounded by a single document's chunk count, not by the thread pool. The WikiHow corpus averages 3.31 chunks per document, so most of the pool sat idle at any size. A thread sweep measured the write phase at ~13 minutes for 16, 32 and 64 threads alike, while extraction either side of it halved with each doubling.

The pool was also sized wrong. `S3BasedDocs` read `GraphRAGConfig.extraction_num_threads_per_worker` at upload time, inside a worker spawned by `run_pipeline` that inherits no parent memory, so a run asking for 64 uploaded with the default of 4.

End to end this is 1.15x on the same LLM work, with byte-identical output.

## Changes

- `S3ChunkUploader.upload` keeps a bounded window of documents in flight. Documents are still yielded in order, and still only once their own chunk uploads have been attempted.
- Thread count is resolved in `S3BasedDocs.__init__`, which runs in the process that configured `GraphRAGConfig`, and carried as a field so it pickles with the handler, the same trip an extractor's `num_workers` already makes.
- A `ConfiguredThreadCount` mixin holds the count-or-config fallback; both uploaders and both downloaders use it, so read back is sized by whatever configured the writer.

## Problem

Uploads to the S3 doc store did not scale with `extraction_num_threads_per_worker`: the uploader never queued enough work to use the pool, and the pool was sized from a config read in the wrong process.

Related issue (if any): #

## Testing

`pytest tests/unit`: 1981 passed. The one error, `test_integ_dependency_compatibility::test_boto3_botocore_version_alignment`, is a pre-existing pip-resolution failure unrelated to this change.

New tests cover cross-document overlap (a `threading.Barrier` that only releases on genuinely simultaneous uploads, so it fails against the serial version), yield order, each chunk uploaded exactly once, the `INDEX_KEY` skip, a failed chunk not stopping the stream, and the thread count surviving into an uploader and a downloader whose process sees the default.

Measured against the WikiHow benchmark corpus on real S3, output byte-identical to the previous implementation.

## Checklist

`num_threads` is now resolved when `S3BasedDocs` is constructed, so setting `GraphRAGConfig.extraction_num_threads_per_worker` afterwards no longer affects it. The call sites in `benchmarks/scripts/benchmark_extract.py` and the integration-test scripts read the value from the environment, which the config picks up lazily, so none are affected.

A failed chunk upload is logged, not raised, and its document is still yielded. That is unchanged behaviour, but it means a yielded document doesn't guarantee every chunk reached S3. The docstring now says so.

---

### DIFF
--- a/lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/load/s3_based_docs.py
+++ b/lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/load/s3_based_docs.py
@@ -378,30 +378,63 @@ class S3ChunkUploader(BaseComponent):
                 ServerSideEncryption='AES256'
             )
 
-    def upload(self, source_documents: List[SourceDocument]):
+    def _drain(self, futures):
+        for future in futures:
+            try:
+                future.result()
+            except Exception as e:
+                logger.error(f'Error uploading chunk: {str(e)}')
 
+    def upload(self, source_documents: List[SourceDocument]):
+        """
+        Upload each source document's chunks, yielding a document once its own
+        chunks are written.
+
+        Uploads for several documents are in flight at once. Waiting for one
+        document before submitting the next held in-flight uploads to that
+        document's chunk count - 3.31 on the WikiHow benchmark corpus - so the
+        pool sat idle however many threads it was given, and the write phase
+        took the same 13 minutes at 16, 32 and 64 threads.
+
+        Documents are still yielded in order, and still only after their own
+        chunks are durable, so a consumer sees what it saw before.
+        """
         s3_client = GraphRAGConfig.s3
-        
-        with concurrent.futures.ThreadPoolExecutor(max_workers=GraphRAGConfig.extraction_num_threads_per_worker) as executor:
+        num_threads = GraphRAGConfig.extraction_num_threads_per_worker
+
+        # Enough queued chunks to keep every thread busy while the oldest
+        # document drains, without reading the whole corpus into memory.
+        max_inflight = num_threads * 2
+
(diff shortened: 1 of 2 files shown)


mykola-pereyma (approved):
LGTM — correct, well-diagnosed fix for the write-phase bottleneck. Verified the concurrency: ordering is preserved (deque + popleft), documents are drained before being yielded, no deadlock risk (submit never blocks, result() only waits on that doc's work), memory is bounded, and error handling matches the prior fail-soft behavior. The thread-count-carried-as-a-field fix correctly addresses the spawn-no-inherited-memory problem. Tests are thorough.

A few non-blocking suggestions:

1. **Mixin field collection is pydantic-version-sensitive.** `ConfiguredThreadCount` declares `num_threads: Optional[int] = None` as a class annotation and relies on `BaseComponent`'s field system picking it up across the MRO. It works (the propagation tests assert `handler._uploader.num_threads == 64`), but this behavior can differ between pydantic v1/v2. Consider a brief comment noting the dependency, or declaring the field explicitly on each component for robustness.

2. **`self.num_threads or config` treats explicit `0` as unset.** `num_threads=0` silently falls back to the config value. `0` is nonsensical here so it's not a real bug, but `if self.num_threads is None` would be more precise.

3. **`max_inflight = num_threads * 2` is a magic number.** Reasonable heuristic; worth a one-line note that this is the tuning knob if deeper queueing is ever needed.

oussamahansal (approved):
LGTM

PR #509, opened 2026-08-27
[FIX] Validate ids used to build FileSystemTap output paths

noel-improv:
## Description

`FileSystemTap` builds its output paths by joining an id onto a directory it prepared, with no validation of the id. An id carrying a path separator opens a new path segment, so the write lands outside the directory the tap owns. This adds an id check at the three places the tap does that join.

The tap is opt-in. `ExtractionPipeline` defaults its decorator to `PassThroughDecorator`, and nothing in the toolkit constructs a `FileSystemTap`, so a default pipeline never reaches this code. In a pipeline that does use it, ids have normally been rewritten to `IdGenerator` hashes first, and those carry no separator.

## Changes

- `FileSystemTap._validate_id` rejects an id that is empty or whitespace only, contains a path separator, contains a control character, or names a directory. It is a blocklist, not the allowlist regex `S3ChunkStore._validate_chunk_id` uses. Filenames legitimately carry characters an S3 key allowlist has no reason to permit, and control characters and `.`/`..`, which are valid in an S3 key, have to be refused for a filesystem write.
- Called before the two source writes in `handle_input_docs`, and before the chunk write in `handle_output_doc`. That third site builds its path the same way and needed the same check.
- Rejecting separators is what makes a relative segment inert, so there is no separate parent-reference case.

## Problem

Ids reaching the tap are not guaranteed to have been rewritten, and the tap trusted them to be usable as filenames. Ids the pipeline produces are unaffected, since they hold no separators.

Related issue (if any): #

## Testing

New tests cover each rejected form for both `doc_id` and `node_id`, a containment check that snapshots every file under the temp root and asserts nothing new appears outside the tap's own directories, and a rewritten id in the usual hash form still writing both of its files. They were written and run red against the unfixed code first. The existing tests in that file already write to real temp directories, so the check is exercised against a real filesystem rather than a mock.

`file_system_tap.py` is at 100% statement coverage. Full suite: 2073 passed, 15 skipped, 65.23% total against the 56% gate. The one error, `test_integ_dependency_compatibility::test_boto3_botocore_version_alignment`, is a pre-existing pip-resolution failure unrelated to this change.

## Checklist

One behaviour change: an id that would previously have written outside the tap's directories now raises `ValueError`. No id the pipeline produces is affected, and no default pipeline reaches this code.

The end of the file also lost some trailing whitespace and gained a final newline. Both were pre-existing.

Not addressed here: `IdRewriter._new_id` returns an id unchanged when it already starts with `aws:`, so a caller can supply one that is never hashed. That is a separate concern and needs its own change.

---

### DIFF
--- a/lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/extract/file_system_tap.py
+++ b/lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/extract/file_system_tap.py
@@ -53,6 +53,24 @@ class FileSystemTap(PipelineDecorator):
         self.chunks_dir = chunks_dir
         self.sources_dir = sources_dir
 
+    @staticmethod
+    def _validate_id(value:str, name:str) -> None:
+        """
+        Reject an id that would write outside the directories the tap prepared.
+
+        Output paths join an id onto a prepared directory, so a separator in the
+        id opens a new path segment and a climbing id lands anywhere. Rewritten
+        ids are IdGenerator hashes and carry no separator.
+        """
+        if not value or not value.strip():
+            raise ValueError(f'{name} must be a non-empty string.')
+        if '/' in value or '\\' in value:
+            raise ValueError(f'{name} must not contain a path separator: {value!r}')
+        if any(ord(c) < 0x20 or ord(c) == 0x7f for c in value):
+            raise ValueError(f'{name} must not contain control characters: {value!r}')
+        if value.strip() in ('.', '..'):
+            raise ValueError(f'{name} must not name a directory: {value!r}')
+
     def handle_input_docs(self, docs:Iterable[SourceDocument]) -> Iterable[SourceDocument]:
         """
         Processes a collection of source documents, saving their raw textual content and JSON representation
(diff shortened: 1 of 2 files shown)


lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/extract/file_system_tap.py
```diff
@@ -53,6 +53,24 @@ def __init__(self, subdirectory_name, clean=True, output_dir=None):
         self.chunks_dir = chunks_dir
         self.sources_dir = sources_dir
 
+    @staticmethod
+    def _validate_id(value:str, name:str) -> None:
```
oussamahansal, line 57: in the PR description you mentioned "same shape as S3ChunkStore._validate_chunk_id" but that one is an allowlist regex and this is a blocklist to me it is different mechanism (and S3 actually permits, which you correctly reject here). maybe reword the description

lexical-graph/tests/unit/indexing/extract/test_file_system_tap.py
```diff
@@ -346,3 +346,114 @@ def test_full_pipeline_flow(self):
...
+            tap = FileSystemTap(
+                subdirectory_name="test_run",
+                clean=True,
+                output_dir=temp_dir
+            )
+
+            with pytest.raises(ValueError, match='non-empty'):
+                tap.handle_input_docs([_doc(doc_id)])
+
+    @pytest.mark.parametrize('doc_id', ['.', '..'])
```
oussamahansal: The ./.. test only feeds exact values, but the code does value.strip() in ('.', '..')  add ' . ' / ' .. ' to pin the .strip(), else dropping it would slip through untested

oussamahansal (approved):

mykola-pereyma (approved):
LGTM

PR #514, opened 2026-08-31
[FIX] Harden SageMaker notebook defaults in CFN templates

oussamahansal:
## Description

Harden the NeptuneNotebookInstance defaults across all 9 lexical-graph CloudFormation templates

## Changes

- `RootAccess: Disabled`
- New optional `NotebookKmsKeyId` parameter + `HasNotebookKmsKey`condition; `KmsKeyId` set to the CMK when provided (empty default)

## Problem

Related issue (if any): #

## Testing

## Checklist

---

### DIFF
--- a/examples/lexical-graph/cloudformation-templates/graphrag-toolkit-neptune-db-aurora-postgres-existing-vpc.json
+++ b/examples/lexical-graph/cloudformation-templates/graphrag-toolkit-neptune-db-aurora-postgres-existing-vpc.json
@@ -32,9 +32,9 @@
       "ConstraintDescription": "Must be a valid subnet ID in the specified VPC or empty"
     },
     "NotebookSubnetId": {
-      "Description": "ID of the subnet for the SageMaker notebook (must have internet access)",
+      "Description": "ID of the subnet for the SageMaker notebook. Must be a private subnet with outbound internet access via a NAT gateway (the notebook is launched with direct internet access disabled).",
       "Type": "AWS::EC2::Subnet::Id",
-      "ConstraintDescription": "Must be a valid subnet ID in the specified VPC with internet access"
+      "ConstraintDescription": "Must be a valid private subnet ID in the specified VPC with outbound internet access via a NAT gateway"
     },
     "NeptuneDbInstanceType": {
       "Description": "Neptune DB instance type (use db.serverless to provision a Neptune Serverless cluster)",
@@ -106,6 +106,11 @@
     "IamPolicyArn": {
       "Type": "String",
       "Description": "ARN of an additional IAM policy to be attached to the GraphRAG client IAM role (optional)",
+      "Default": ""
+    },
+    "NotebookKmsKeyId": {
+      "Type": "String",
+      "Description": "Optional KMS key ARN or id used to encrypt the SageMaker notebook ML storage volume with a customer managed key (CMK). Leave empty to use the default AWS managed encryption.",
       "Default": ""
     },
     "NotebookInstanceType": {
@@ -211,6 +216,18 @@
     }
   },
   "Conditions": {
+    "HasNotebookKmsKey": {
+      "Fn::Not": [
+        {
+          "Fn::Equals": [
+            {
+              "Ref": "NotebookKmsKeyId"
+            },
+            ""
+          ]
+        }
+      ]
+    },
     "ExampleNotebooksURLIsNotBlank": {
       "Fn::Not": [
         {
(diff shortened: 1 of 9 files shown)


noel-improv (approved):
lgtm just a couple of comments:
- The body lists only RootAccess and the KMS parameter, but the two changes with real deployment impact are DirectInternetAccess: Disabled and the subnet move. Both force a notebook replacement on a stack update and drop its public IP.
- **Question**: With egress now forced through NAT, S3 and Bedrock traffic bills at the NAT data-processing rate. An S3 gateway endpoint is free, might be worth adding while the templates are open?
- **Nit**: NotebookKmsKeyId takes any string, so a typo fails mid-create instead of at parameter validation. An AllowedPattern would catch it.

oussamahansal:
> * **Question**: With egress now forced through NAT, S3 and Bedrock traffic bills at the NAT data-processing rate. An S3 gateway endpoint is free, might be worth adding while the templates are open?

Keeping this PR scoped to the fix rather than expanding the VPC resources here we can do that in another pr

examples/lexical-graph/cloudformation-templates/graphrag-toolkit-neptune-db-aurora-postgres-existing-vpc.json
```diff
@@ -1335,6 +1354,19 @@
           "Ref": "NotebookInstanceType"
         },
         "PlatformIdentifier": "notebook-al2023-v1",
+        "RootAccess": "Disabled",
+        "DirectInternetAccess": "Disabled",
```
mykola-pereyma, line 1358: **Egress risk (this template specifically).** `DirectInternetAccess` was absent on `main` (SageMaker default = **Enabled**), so this line flips the notebook from having direct internet to none. Unlike the self-contained VPC templates, this one provisions **no NAT gateway** and the notebook runs on a user-supplied subnet (`NotebookSubnetId`). The OnStart lifecycle script runs `sudo pip install packaging==24.2` and sets up graph-notebook — all需 internet — so with no egress the failure happens at **startup** (the instance may never reach InService), not just later.

Please either keep `DirectInternetAccess: Enabled` for the existing-VPC case, or document that `NotebookSubnetId` **must** have outbound internet (NAT gateway or the required VPC endpoints for SageMaker + Bedrock + S3 + PyPI). Can you confirm this is covered?

examples/lexical-graph/cloudformation-templates/graphrag-toolkit-neptune-db-aurora-postgres-existing-vpc.json
```diff
@@ -1335,6 +1354,19 @@
           "Ref": "NotebookInstanceType"
         },
         "PlatformIdentifier": "notebook-al2023-v1",
+        "RootAccess": "Disabled",
```
mykola-pereyma, line 1357: **RootAccess vs the lifecycle script (applies to all 9 templates).** `RootAccess` was default-Enabled before; disabling it removes ec2-user root for notebook users. Every template's OnStart lifecycle script does `sudo -u ec2-user -i` then `sudo pip install …` (2 sudo uses). AWS runs lifecycle configs as root even when RootAccess is disabled, so this *may* be fine — but the inner `sudo pip install` executes **as ec2-user**, whose sudo is exactly what `RootAccess: Disabled` restricts.

Could you confirm via an actual deployment that the notebook still reaches **InService** and the lifecycle `sudo pip install` succeeds with `RootAccess: Disabled`? That specific hop is the thing to prove.

examples/lexical-graph/cloudformation-templates/graphrag-toolkit-neptune-db-aurora-postgres-existing-vpc.json
```diff
@@ -1335,6 +1354,19 @@
...
+        "DirectInternetAccess": "Disabled",
+        "KmsKeyId": {
```
mykola-pereyma, line 1359: **CMK key-policy requirement (opt-in path).** Default empty → `AWS::NoValue` → AWS-managed key, so the default path is unchanged. But if a user supplies `NotebookKmsKeyId`, standard AWS behavior requires the SageMaker service to be granted use of that key (`kms:CreateGrant`, `Decrypt`, `GenerateDataKey`) or `CreateNotebookInstance` fails. Please document that the supplied CMK's key policy must allow SageMaker. Is this covered in the docs?

mykola-pereyma (changes requested):
The hardening itself is correct and consistent where applicable (RootAccess + KMS on all 9; DirectInternetAccess + public→private subnet on the self-contained VPC templates). Requesting changes to confirm three functional points are covered — inline. Separately, the PR description still only lists RootAccess + KMS; please also add the `DirectInternetAccess: Disabled` and public→private subnet changes (they change the notebook's network reachability).

mykola-pereyma (approved):
LGTM

PR #515, opened 2026-09-01
[FEATURE] Add a source-id key collision generator to benchmarks

noel-improv:
## Description

Adds a tool that measures how often `IdGenerator` produces the same source id for different documents, and a test suite that locks it to the real implementation.

`create_source_id` returns `aws::{md5(text)[:8]}:{md5(metadata_str)[:4]}` (`id_generator.py:84`), so a source id discriminates on 48 bits. `IdRewriter` passes `_get_properties_str(node.metadata, '')` (`id_rewriter.py:74`), so a corpus loaded without metadata gives every document the same second component and 32 bits do the work. Both S3 storage prefixes are built from the bare source id (`s3_based_docs.py:209` and `:434`), so two documents sharing an id share a prefix.

Measured, against a birthday expectation of 116.4 at 1M on 32 bits:

| Documents | Metadata absent (32 bit) | Metadata present (48 bit) |
|---|---|---|
| 100,000 | 1 colliding pair | 0 |
| 1,000,000 | 127 colliding pairs | 0, and a 0.2% chance of any |
| 10,000,000 | 11,762 colliding pairs | 0, and a 16% chance of any |

Nothing in the toolkit changes. This is measurement only, and it feeds the storage-key decision in the restart design.

## Changes

- `benchmarks/utils/key_collisions.py`. Generates ids synthetically and counts collisions against the closed form. Documents sharing text share a key by design, so those pairs are counted apart from hash collisions. `--corpus` measures the duplicate rate on real text. `create_chunk_id` models both `use_chunk_id_delimiter` settings, not only the default.
- `benchmarks/tests/test_key_collisions.py`. 16 tests. `test_source_id_matches_the_id_generator` and `test_chunk_id_matches_the_id_generator` compare the reimplementation against a real `IdGenerator` rather than against hardcoded id literals, so changing the hash slices in `id_generator.py` fails the suite. `IdGenerator()` needs no AWS config or credentials to construct and hash, so the comparison is cheap.
- `benchmarks/requirements.txt`. Adds `numpy>=1.24.0`. The module is the first use of numpy under `benchmarks/`, and it resolved only transitively through `llama-index-core`.

Source keys fit in `uint64` and go through numpy, which is what keeps 10M documents tractable. Composite source-plus-chunk keys are 80 bits and count through a `Counter` instead.

## Problem

Storage prefixes inherit the source id's width. Once completion markers land, two documents sharing a prefix share one marker, which then matches neither and re-stages both on every restart. Sizing that risk needed a number rather than an estimate.

Related issue (if any): #325. The document-level manifest needs storage identity to be unique: two documents sharing a prefix share one completion marker, which then matches neither and re-stages both on every restart.

## Testing

- [x] Existing tests pass (`pytest`) — full `benchmarks/tests` suite, 146 passed

Ran against WikiHow-2000 as a real-corpus check: 2,000 distinct texts, no duplicates, no collisions, which is what 32 bits predicts at that size.

## Checklist

One deviation worth naming rather than leaving to review: this is the only module in `benchmarks/` with `__main__` and argparse. The tool is meant to be run once by a person, so a CLI seemed right, but say the word and it becomes import-only.

CI now runs these tests. No workflow had a `benchmarks/**` path filter, so nothing was running them and the fidelity check against `IdGenerator` caught drift only when someone ran the suite locally. `benchmarks-tests.yml` closes that, folded in here rather than split out. It runs tests only: two files under `benchmarks/scripts/` predate the license-header convention and a header job would fail on arrival, so that stays a separate change.

`benchmarks/tests/conftest.py` registers the Hypothesis `ci` profile the workflow selects, mirroring `lexical-graph/tests/conftest.py`. Without it the env var is inert and the 49 property tests run at 100 examples against a 200 ms deadline, which is where a slow runner turns them flaky.

One fix on top of the original spike: the low-16-bits assertion masked a numpy `uint64` with a Python int. Before numpy 2.0 that raises `TypeError` — the int becomes `int64` and `uint64 & int64` has no safe common type. `requirements.txt` allows `numpy>=1.24.0`, so the test passed or failed on whatever pip resolved. Casting with `int()` first matches what the neighbouring metadata assertion already did.

### DIFF
--- a/.github/workflows/benchmarks-tests.yml
+++ b/.github/workflows/benchmarks-tests.yml
@@ -1,0 +1,59 @@
+# Runs the benchmarks unit tests on push/PR to main. The lexical-graph
+# workflow is scoped to lexical-graph/**, so nothing was running these.
+name: Benchmarks Unit Tests
+
+permissions:
+  contents: read
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - "benchmarks/**"
+      - ".github/workflows/benchmarks-tests.yml"
+  pull_request:
+    branches: [main]
+    paths:
+      - "benchmarks/**"
+      - ".github/workflows/benchmarks-tests.yml"
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+
+    steps:
+      - uses: actions/checkout@v7
+
+      - name: Set up Python
+        uses: actions/setup-python@v7
+        with:
+          python-version: "3.12"
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@20cfd1bf945f4377ade1205e4dbc17946fc9a30d # v10.0.1
+
+      - name: Create virtual environment
+        run: uv venv --seed .venv
+
+      - name: Install dependencies
+        # The benchmarks import graphrag_toolkit, so lexical-graph and its
+        # requirements come first. Separate commands: requirements.txt opens
+        # with `--only-binary :all:`, which would block building the local
+        # package if combined.
+        env:
+          # Copy instead of hardlink into the venv so bundled data files have
+          # st_nlink=1; nltk's pathsec check refuses to open multiply-linked files.
+          UV_LINK_MODE: copy
+        run: |
+          uv pip install --python .venv/bin/python \
+            -r lexical-graph/src/graphrag_toolkit/lexical_graph/requirements.txt
+          uv pip install --python .venv/bin/python -e ./lexical-graph
+          uv pip install --python .venv/bin/python -r benchmarks/requirements.txt
+
+      - name: Run tests
+        env:
+          HYPOTHESIS_PROFILE: ci
+          PYTHONPATH: benchmarks
+        run: |
+          .venv/bin/python -m pytest -v -l --tb=short benchmarks/tests/
--- a/.github/workflows/benchmarks-tests.yml
+++ b/.github/workflows/benchmarks-tests.yml
@@ -54,6 +54,5 @@ jobs:
       - name: Run tests
         env:
           HYPOTHESIS_PROFILE: ci
-          PYTHONPATH: benchmarks
         run: |
           .venv/bin/python -m pytest -v -l --tb=short benchmarks/tests/
--- a/benchmarks/requirements.txt
+++ b/benchmarks/requirements.txt
@@ -9,6 +9,7 @@
 # llama-index-llms-bedrock-converse, boto3, botocore) is assumed to be
 # installed separately via `pip install -e ./lexical-graph`.
 
+numpy>=1.24.0
 python-dotenv>=1.0.0
 boto3>=1.28.0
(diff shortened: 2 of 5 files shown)


mykola-pereyma (approved):
LGTM. This is a clean, low-risk, measurement-only addition, isolated to `benchmarks/` with no toolkit changes. The engineering is careful:

- **Fidelity lock** — `test_source_id_matches_the_id_generator` / `test_chunk_id_matches_the_id_generator` pin the reimplemented `create_source_id`/`create_chunk_id` to the real `IdGenerator` (both delimiter variants), so the numbers can't silently drift from the toolkit.
- **Sound counting** — combinatorial pairs (not excess docs), max-group, >uint64 composite path via `Counter`, and empty/edge cases, all tested (14 tests). The measured figures line up with the birthday closed form.
- I verified the cited facts against source: 48-bit shape at `id_generator.py:84`, and the 32-bit-when-metadata-absent path at `id_rewriter.py:74-75`.
- Dependency hygiene: adds only `numpy>=1.24.0`.

One thing to record for whoever acts on this measurement (the restart/storage-key work in #325): **widening the source id to the full 48 bits is a backward-incompatible change to persisted storage identity** if applied globally or in place, and it should be treated with care:

- **Why it's breaking:** source nodes MERGE on `sourceId` (`source_graph_builder.py:79`), chunk ids embed the source id (`create_chunk_id` → `{source_id}:{hash}`), and S3 prefixes derive from the bare source id. Changing the derivation makes re-ingested documents mint new ids that no longer match existing nodes/prefixes → duplicate source/chunk/statement nodes, orphaned S3 artifacts, and (once completion markers land) markers that match neither copy.
- **Feature flag / id-version:** gate the wider derivation behind an opt-in flag defaulting to current behavior, mirroring the existing `use_chunk_id_delimiter` precedent in `id_generator.py` ("Old behavior … preserves existing chunk IDs"). Apply 48-bit only to new graphs/tenants; leave existing deployments untouched.
- **Migration:** because ids cascade (source → chunk, and nodes MERGE on them), there is no cheap in-place migration — a retrofit means a full rebuild or a re-key of the whole source/chunk/statement hierarchy plus S3 re-prefixing. If a migration path is offered, document it explicitly and test it.
- **Documentation & operator messaging:** update the docs to describe the id scheme, the flag, and the compatibility boundary; and emit a clear, actionable log/error when the toolkit encounters an id/version mismatch (old data under a new scheme) rather than silently duplicating — so operators know to rebuild or migrate.

None of this blocks #515 — it's just the compatibility contract the follow-up change will need to honor. Approving the benchmark.

oussamahansal (changes requested):
Can you please check the unit tests ?
test_absent_metadata_leaves_32_bits_discriminating FAILED benchmarks/tests/test_key_collisions.py::TestKeyGeneration::test_absent_metadata_leaves_32_bits_discriminating
1 failed, 15 passed

oussamahansal (approved):

PR #516, opened 2026-09-02
[FIX] Notebook AWS_REGION in .env and NextGen OpenSearch VPC endpoint

oussamahansal:
## Description

Two fixes for the lexical-graph CloudFormation templates
## Changes

- Write AWS_REGION into the notebook .env
- Use a standard EC2 interface VPC endpoint for the NextGen collection

## Problem

Related issue (if any): #

## Testing

## Checklist

- [ ] Code follows existing style and conventions

---

### DIFF
--- a/examples/lexical-graph/cloudformation-templates/graphrag-toolkit-neptune-db-aurora-postgres-existing-vpc.json
+++ b/examples/lexical-graph/cloudformation-templates/graphrag-toolkit-neptune-db-aurora-postgres-existing-vpc.json
@@ -1657,6 +1657,11 @@
                       "Ref": "AWS::Region"
                     },
                     "\" >> .env\n",
+                    "  echo \"export AWS_REGION=",
+                    {
+                      "Ref": "AWS::Region"
+                    },
+                    "\" >> .env\n",
                     "  chmod -R 700 .\n",
                     "  chmod 600 .env\n",
                     "  if [ -f \"./run_test_suite.sh\" ]\n",
--- a/examples/lexical-graph/cloudformation-templates/graphrag-toolkit-neptune-db-aurora-postgres-existing-vpc.json
+++ b/examples/lexical-graph/cloudformation-templates/graphrag-toolkit-neptune-db-aurora-postgres-existing-vpc.json
@@ -1657,11 +1657,6 @@
                       "Ref": "AWS::Region"
                     },
                     "\" >> .env\n",
-                    "  echo \"export AWS_REGION=",
-                    {
-                      "Ref": "AWS::Region"
-                    },
-                    "\" >> .env\n",
                     "  chmod -R 700 .\n",
                     "  chmod 600 .env\n",
                     "  if [ -f \"./run_test_suite.sh\" ]\n",
--- a/examples/lexical-graph/cloudformation-templates/graphrag-toolkit-neptune-db-opensearch-nextgen-serverless.json
+++ b/examples/lexical-graph/cloudformation-templates/graphrag-toolkit-neptune-db-opensearch-nextgen-serverless.json
@@ -1227,12 +1227,22 @@
       }
     },
     "OpenSearchServerlessVpcEndpoint": {
-      "Type": "AWS::OpenSearchServerless::VpcEndpoint",
+      "Type": "AWS::EC2::VPCEndpoint",
       "Condition": "EnableVpcAccess",
       "Properties": {
-        "Name": {
-          "Fn::Sub": "${ApplicationId}-vpce"
+        "ServiceName": {
+          "Fn::Sub": "com.amazonaws.${AWS::Region}.aoss-data"
         },
+        "VpcEndpointType": "Interface",
+        "PrivateDnsEnabled": true,
+        "Tags": [
+          {
+            "Key": "Name",
+            "Value": {
+              "Fn::Sub": "${ApplicationId}-vpce"
+            }
+          }
+        ],
         "VpcId": {
           "Fn::If": [
             "CreateVPC",
@@ -1734,6 +1744,11 @@
                     "  echo \"export EVALUATION_MODEL=",
                     {
                       "Ref": "ModelId"
+                    },
+                    "\" >> .env\n",
+                    "  echo \"export AWS_REGION=",
+                    {
+                      "Ref": "AWS::Region"
                     },
                     "\" >> .env\n",
                     "  chmod -R 700 .\n",
--- a/examples/lexical-graph/cloudformation-templates/graphrag-toolkit-neptune-db-opensearch-nextgen-serverless.json
+++ b/examples/lexical-graph/cloudformation-templates/graphrag-toolkit-neptune-db-opensearch-nextgen-serverless.json
@@ -92,7 +92,7 @@
       "ConstraintDescription": "Must contain only letters, numbers, and the characters . _ : / -"
     },
     "ExistingVpcId": {
-      "Description": "Optional existing VPC ID to reuse. If provided, ExistingSubnetIds must also be provided. Leave empty to create a new VPC.",
+      "Description": "Optional existing VPC ID to reuse. If provided, ExistingSubnetIds must also be provided. The VPC must have DNS support and DNS hostnames enabled so the OpenSearch Serverless collection endpoint resolves privately over the VPC endpoint. Leave empty to create a new VPC.",
       "Type": "String",
       "Default": ""
     },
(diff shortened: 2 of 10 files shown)


examples/lexical-graph/cloudformation-templates/graphrag-toolkit-neptune-db-aurora-postgres-existing-vpc.json
```diff
@@ -1657,6 +1657,11 @@
                       "Ref": "AWS::Region"
                     },
                     "\" >> .env\n",
+                    "  echo \"export AWS_REGION=",
```
mykola-pereyma: This `export AWS_REGION=` is a **duplicate** — the same statement already exists a few lines up (L1655 here), so this block writes AWS_REGION to `.env` twice with the identical `Ref AWS::Region` value.

It's not just this template: **every one of the 9 templates already exports AWS_REGION on main**, so this PR adds a redundant duplicate to each (existing-vpc goes from 2 → 3). Counts (main → PR): the 8 others 1 → 2, existing-vpc 2 → 3.

Proof — already present on main before this PR:
- https://github.com/awslabs/graphrag-toolkit/blob/a6132c850ef5b302bfe8d65b59d3270f39765210/examples/lexical-graph/cloudformation-templates/graphrag-toolkit-neptune-db-aurora-postgres-existing-vpc.json#L1655
- (representative) https://github.com/awslabs/graphrag-toolkit/blob/a6132c850ef5b302bfe8d65b59d3270f39765210/examples/lexical-graph/cloudformation-templates/graphrag-toolkit-neptune-db-aurora-postgres.json#L1426

It's harmless at runtime (same value re-exported), but since the region is already written, the AWS_REGION part of this PR looks unnecessary. Could you drop these added lines? If AWS_REGION is genuinely missing on some path, please point to the specific template/script — I don't see one; all 9 already write it.

mykola-pereyma (commented):
On the AWS_REGION change: every template already exports AWS_REGION into the notebook .env on `main`, so these additions are duplicates (one per template; existing-vpc ends up with three). Details inline. Harmless at runtime but redundant — suggest dropping the AWS_REGION additions unless there's a path where it's actually missing.

(Separately, the NextGen VPC-endpoint type swap is the substantive change and would benefit from confirmation it was deployed end-to-end — the AOSS network policy still references this endpoint in `SourceVPCEs`, and swapping the AOSS-managed endpoint for a generic EC2 interface endpoint needs to be verified against a live collection.)

mykola-pereyma (commented):
For the NextGen VPC-endpoint change (Change 2): it's a structurally sound CloudFormation swap, but its actual functioning depends on two AOSS control-plane behaviors that aren't verifiable from the template. Could you provide deploy confirmation? Specifics inline on the endpoint's `ServiceName`.

examples/lexical-graph/cloudformation-templates/graphrag-toolkit-neptune-db-opensearch-nextgen-serverless.json
```diff
@@ -1227,12 +1227,22 @@
...
     },
     "OpenSearchServerlessVpcEndpoint": {
-      "Type": "AWS::OpenSearchServerless::VpcEndpoint",
+      "Type": "AWS::EC2::VPCEndpoint",
       "Condition": "EnableVpcAccess",
       "Properties": {
-        "Name": {
-          "Fn::Sub": "${ApplicationId}-vpce"
+        "ServiceName": {
+          "Fn::Sub": "com.amazonaws.${AWS::Region}.aoss-data"
```
mykola-pereyma, line 1234: Swapping the AOSS-managed `AWS::OpenSearchServerless::VpcEndpoint` for a standard `AWS::EC2::VPCEndpoint` here is a well-formed CloudFormation change (SubnetIds/SecurityGroupIds/VpcId present, `Name` correctly moved to a tag, and the network policy's `SourceVPCEs` still renders since `Ref` returns a `vpce-` id). But two things depend on AOSS control-plane behavior that can't be confirmed from the template — could you confirm them with a deployment?

1. **`com.amazonaws.${AWS::Region}.aoss-data` is a valid PrivateLink service name and supports `PrivateDnsEnabled: true`.** The long-standing service is `...aoss`; if `aoss-data` (or private DNS) is wrong, the endpoint `CREATE_FAIL`s.
2. **The AOSS network policy accepts an EC2 interface endpoint id in `SourceVPCEs`.** Historically `SourceVPCEs` took OpenSearch-Serverless-*managed* endpoint ids (from `aoss:CreateVpcEndpoint` / `AWS::OpenSearchServerless::VpcEndpoint`), not arbitrary EC2 interface endpoints. If AOSS still validates that, `OpenSearchServerlessNetworkPolicy` `CREATE_FAIL`s or the collection is unreachable from the VPC.

Specifically, could you deploy this template with `EnableVpcAccess=true` / `EnablePublicAccess=false` and confirm:
- the `OpenSearchServerlessVpcEndpoint` reaches CREATE_COMPLETE,
- the `OpenSearchServerlessNetworkPolicy` reaches CREATE_COMPLETE (this is the one that validates SourceVPCEs accepts the EC2 endpoint id), and
- the notebook can actually query the collection endpoint.

With that evidence I'm happy to sign off; without it I can't verify the VPC path works end-to-end.

examples/lexical-graph/cloudformation-templates/graphrag-toolkit-neptune-db-aurora-postgres-existing-vpc.json
oussamahansal: You're right for existing-vpc — it already writes AWS_REGION to .env on main, so I've dropped my duplicate there.
 For the other 8, though, main writes AWS_REGION only to ~/.bashrc ([e.g. aurora-postgres L1426 → >> ~/.bashrc](https://github.com/awslabs/graphrag-toolkit/blob/main/examples/lexical-graph/cloudformation-templates/graphrag-toolkit-neptune-db-aurora-postgres.json#L1426)), not to .env. The notebooks load config via load_dotenv('.env'), and the Jupyter kernel doesn't inherit ~/.bashrc, so the kernel had no region — which is the empty-region SigV4 401 against OpenSearch we hit in testing. So the .env line on those 8 is the actual fix, not a duplicate. Keeping those, dropped the existing-vpc one.

examples/lexical-graph/cloudformation-templates/graphrag-toolkit-neptune-db-opensearch-nextgen-serverless.json
oussamahansal: Confirmed by deployment...the stack reached CREATE_COMPLETE, so both the AWS::EC2::VPCEndpoint (validates com.amazonaws.<region>.aoss-data + PrivateDnsEnabled: true) and OpenSearchServerlessNetworkPolicy (validates SourceVPCEs accepts the EC2 interface-endpoint id) created cleanly. From the notebook the collection endpoint resolved to private VPC IPs — vs a public IP with the previous managed endpoint.

examples/lexical-graph/cloudformation-templates/graphrag-toolkit-neptune-db-opensearch-nextgen-serverless.json
```diff
@@ -1227,12 +1227,22 @@
       }
     },
     "OpenSearchServerlessVpcEndpoint": {
-      "Type": "AWS::OpenSearchServerless::VpcEndpoint",
+      "Type": "AWS::EC2::VPCEndpoint",
```
noel-improv, line 1230: Blocking: this breaks `test_shared_resources_are_identical_except_the_known_delta`, which wants this resource identical across the Classic and NextGen templates (fails here, passes on base).

I dont think CI will show it as the test workflow is path-filtered to `lexical-graph/**` and this PR is `examples/**` only, so it'd surface on the next unrelated PR instead. Could you either make the same change in the Classic template, or add this resource to the test's intentional-delta list?

examples/lexical-graph/cloudformation-templates/graphrag-toolkit-neptune-db-opensearch-nextgen-serverless.json
```diff
@@ -1227,12 +1227,22 @@
...
       "Condition": "EnableVpcAccess",
       "Properties": {
-        "Name": {
-          "Fn::Sub": "${ApplicationId}-vpce"
+        "ServiceName": {
+          "Fn::Sub": "com.amazonaws.${AWS::Region}.aoss-data"
         },
+        "VpcEndpointType": "Interface",
+        "PrivateDnsEnabled": true,
```
noel-improv, line 1237: Non-blocking: the deploy covered the `CreateVPC` path, where the template sets the DNS attributes itself. On the `ExistingVpcId` path private DNS fails if the user's VPC has DNS hostnames or support off, which worked before with the managed endpoint. Worth a line on the `ExistingVpcId` description?

examples/lexical-graph/cloudformation-templates/graphrag-toolkit-neptune-db-opensearch-nextgen-serverless.json
```diff
@@ -1227,12 +1227,22 @@
...
+        "Tags": [
```
noel-improv, line 1238: Nit: other resources here carry `StackId`/`Stack`/`Application` alongside `Name` maybe worth matching?

noel-improv (changes requested):
The change looks good, just one blocking item and suggestions referenced below

mykola-pereyma (approved):
LGTM

noel-improv (approved):
lgtm, thanks for addressing the raised items

PR #517, opened 2026-09-03
[FEATURE] Make the source id text digest width configurable

noel-improv:
## Description

A source id is `md5(text)[:8]` plus `md5(metadata_str)[:4]` (`id_generator.py:84`), so 48 bits. `IdRewriter` passes `_get_properties_str(node.metadata, '')`, so a corpus loaded without metadata gives every document the same second component and 32 bits discriminate. Measured collision counts on synthetic ids:

| Documents | Metadata absent (32 bit) | Metadata present (48 bit) |
|---|---|---|
| 100,000 | 1 colliding pair | 0 |
| 1,000,000 | 127 colliding pairs | 0, and a 0.2% chance of any |
| 10,000,000 | 11,762 colliding pairs | 0, and a 16% chance of any |

Colliding documents share a `__Source__` node, so chunks from both attach to one document, and they share an S3 storage prefix.

This makes the text component's width a setting rather than a literal. It changes nothing by default.

## Changes

- `IdGenerator` takes `source_id_hash_length`, defaulting to 8, and rejects a length outside the 32-character digest.
- `GraphRAGConfig.source_id_hash_length`, read from `SOURCE_ID_HASH_LENGTH`, following the pattern of the other extraction settings.
- Both `IdGenerator` call sites pass it, so the setting reaches the ids a run writes rather than sitting unreachable on a constructor.

Widening changes every source id and chunk id, since a chunk id embeds the source id, so a graph written at one width cannot be read at another. Chunk text is unaffected: a re-ingest reproduces the same chunks under new ids rather than re-chunking the corpus.

## Problem

Related issue (if any): #325. Restart needs storage identity to be unique, and the same collision merges distinct documents in the graph.

## Testing

- [x] Existing tests pass (`pytest`) — 2,087 in `tests/unit`

Verified end to end through `IdRewriter` and `SentenceSplitter` that widening changes ids while leaving chunk text byte-identical, and that the default produces `aws::5eb63bbb:d41d` for `('hello world', '')` exactly as before.

## Checklist

The width to use at scale is still open and is being taken to the team, which is why this ships the mechanism at today's default rather than picking a number.

### DIFF
--- a/docs-site/src/content/docs/lexical-graph/configuration.mdx
+++ b/docs-site/src/content/docs/lexical-graph/configuration.mdx
@@ -44,6 +44,7 @@ The configuration includes the following parameters:
 | `embed_dimensions` | Number of dimensions in each vector | `1024` | `EMBEDDINGS_DIMENSIONS` |
 | `extraction_num_workers` | The number of parallel processes to use when running the extract stage | `2` | `EXTRACTION_NUM_WORKERS` |
 | `extraction_num_threads_per_worker` | The number of threads used by each process in the extract stage | `4` | `EXTRACTION_NUM_THREADS_PER_WORKER` |
+| `source_id_hash_length` | Characters of the text digest used in a source id. Widening it separates documents that would otherwise share an id, and changes every source and chunk id, so a graph written at one width cannot be read at another | `8` | `SOURCE_ID_HASH_LENGTH` |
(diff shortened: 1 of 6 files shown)


oussamahansal (approved):
LGTM

mykola-pereyma (approved):
LGTM

PR #518, opened 2026-09-03
[FIX] Derive the S3 document key suffix from the document so a retry overwrites

noel-improv:
## Description

**Stacked on #517.** That PR is the first commit here and is not part of this review. Until it merges, the diff below shows both; the change under review is `feat(lexical-graph): derive the S3 document key suffix from the document`.

`S3DocUploader._upload_doc` builds the key as `{source_id}-{uuid4:5}.jsonl`, so the key is different on every attempt. A retry writes a second object beside the first instead of replacing it, and a restart cannot tell a re-stage from a duplicate.

Behind a flag the suffix becomes a digest of the document's own node ids. A retry then produces the same key and overwrites.

## Changes

- `deterministic_document_key` on `S3DocUploader`, `S3ChunkUploader` and `S3BasedDocs`, defaulting to `False`. With it off, keys match today's layout exactly.
- The suffix is `get_hash(sorted node ids)[:5]` when the flag is on, `uuid4().hex[:5]` when it is off.

Keying on `source_id` alone would not work. `_extract_auto_tuned` emits one source as several `SourceDocument`s when a document exceeds the round capacity, and those would overwrite each other. Node ids differ per round, so the rounds stay in separate objects while a retry of one round still overwrites. Both properties come from the same expression and both are covered.

## Problem

The random suffix is also what currently keeps two colliding documents in separate objects, so dropping it needs the wider key from #517. Sizing that collision is what #515 measured: 127 colliding pairs at 1M documents on the 32 bits that discriminate when metadata is absent.

Related issue (if any): #325. A restart cannot resume against keys that change on every attempt.

## Testing

- [x] Existing tests pass (`pytest`) — 2,094 in `tests/unit`

Nine tests covering both flag states, retry stability, round separation, and independence from the order chunks arrive in.

## Checklist

The flag is not yet reachable from configuration, so today it is set by constructing `S3BasedDocs` directly. Wiring it to a setting belongs with the rest of the restart work rather than here.

### DIFF
--- a/docs-site/src/content/docs/lexical-graph/configuration.mdx
+++ b/docs-site/src/content/docs/lexical-graph/configuration.mdx
@@ -44,6 +44,7 @@ The configuration includes the following parameters:
 | `embed_dimensions` | Number of dimensions in each vector | `1024` | `EMBEDDINGS_DIMENSIONS` |
 | `extraction_num_workers` | The number of parallel processes to use when running the extract stage | `2` | `EXTRACTION_NUM_WORKERS` |
 | `extraction_num_threads_per_worker` | The number of threads used by each process in the extract stage | `4` | `EXTRACTION_NUM_THREADS_PER_WORKER` |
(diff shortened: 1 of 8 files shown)


mykola-pereyma (approved):
LGTM

oussamahansal (approved):

PR #520, opened 2026-09-04
[CHORE] Update LlamaIndex deps

oussamahansal:
## Description

Keep LlamaIndex deps current
## Changes

- Dependency floors bumped to current latest
- Monthly dependency-freshness workflow

## Problem

Related issue (if any): #

## Testing

## Checklist

---

### DIFF
--- a/.github/scripts/check_llama_index_updates.py
+++ b/.github/scripts/check_llama_index_updates.py
@@ -1,0 +1,160 @@
+#!/usr/bin/env python3
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""Check PyPI for llama-index-* releases newer than our pinned floors.
+
+Reads the requirements.txt files listed in the REQUIREMENTS_FILES env var
+(whitespace-separated), extracts every ``llama-index*`` requirement and its
+``>=`` floor, queries PyPI for the latest release of each, and writes a markdown
+report of the packages that have a newer version available.
+
+Outputs (written to $GITHUB_OUTPUT when running under Actions):
+    has_updates : "true" if at least one package has a newer release, else "false"
+
+The markdown report is written to $GITHUB_WORKSPACE/llama_index_report.md.
+
+This is a best-effort nudge: a network or parse error for a single package is
+logged and skipped rather than failing the job.
+"""
+
+import json
+import os
+import re
+import sys
+import urllib.request
+import urllib.error
+
+try:
+    from packaging.version import Version, InvalidVersion
+except ImportError:  # pragma: no cover - packaging ships with pip on the runner
+    print("packaging not available; install it in the workflow step", file=sys.stderr)
+    raise
+
+# Matches e.g. "llama-index-llms-bedrock-converse>=0.15.0". Only >= floors are
+# tracked because that is the pinning style used across the toolkit.
+REQ_LINE = re.compile(r'^(llama-index[a-z0-9-]*)\s*>=\s*([0-9][0-9A-Za-z.\-]*)\s*$')
+
+PYPI_URL = 'https://pypi.org/pypi/{pkg}/json'
+
+
+def latest_version(pkg: str) -> str:
+    """Return the latest (info.version) release of a package on PyPI."""
+    with urllib.request.urlopen(PYPI_URL.format(pkg=pkg), timeout=30) as resp:
+        data = json.load(resp)
+    return data['info']['version']
+
+
+def collect_requirements(files):
+    """Map each llama-index package to its lowest pinned floor and the files it appears in."""
+    found = {}
+    for path in files:
+        if not path or not os.path.exists(path):
+            print(f'skipping missing requirements file: {path}', file=sys.stderr)
+            continue
+        with open(path, encoding='utf-8') as fh:
+            for raw in fh:
+                match = REQ_LINE.match(raw.strip())
+                if not match:
+                    continue
+                pkg, floor = match.group(1), match.group(2)
+                entry = found.setdefault(pkg, {'floor': floor, 'files': set()})
+                entry['files'].add(path)
+                # Keep the lowest floor seen across files (most conservative).
+                try:
+                    if Version(floor) < Version(entry['floor']):
+                        entry['floor'] = floor
+                except InvalidVersion:
+                    pass
+    return found
+
+
+TEST_COMMANDS = """### Suggested tests to run before merging a bump
+
+```bash
+# lexical-graph unit tests
+cd lexical-graph && PYTHONPATH=src python -m pytest tests/
+
+# byokg-rag unit tests
+cd byokg-rag && PYTHONPATH=src python -m pytest tests/
+```
+
+Pay particular attention to the Bedrock / LLM paths, which are most sensitive to
+llama-index-llms-bedrock-converse and llama-index-llms-anthropic changes:
+
+```bash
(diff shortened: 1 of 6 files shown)


mykola-pereyma (approved):
LGTM

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/utils/batch_inference_utils.py
```diff
@@ -65,47 +66,126 @@ def split_nodes(nodes: List[Any], batch_size: int) -> List[List[Any]]:
...
+def _parse_output_text(json_data: dict, output_path: tuple, output_mode: str) -> str:
+    """Extract generated text from a batch output record per a provider's output spec."""
+    node = json_data
+    for key in output_path:
+        if not isinstance(node, dict):
+            node = None
+            break
+        node = node.get(key)
+    if output_mode == 'text':
+        return node if isinstance(node, str) else ''
```
noel-improv: Question, possibly blocking: main did `json_data['generation']`, which raises; walking the path with `.get()` means a missing key now returns `''`, so every node gets an empty extraction and the batch still reports success.

It also stands out that nova and claude read under `modelOutput` while llama reads at the record root. Bedrock wraps provider payloads in `modelOutput`, so should the llama spec be `('modelOutput', 'generation')`? The fixture puts it at the root too, so the test passes either way.

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/utils/batch_inference_utils.py
```diff
@@ -65,47 +66,126 @@ def split_nodes(nodes: List[Any], batch_size: int) -> List[List[Any]]:
...
-        converse_messages, system_prompt = messages_to_converse_messages(messages)
-        request_body = {
-            'messages': converse_messages,
-            'parameters': {
-                'max_new_tokens': inference_parameters['max_tokens'],
-                'temperature': inference_parameters['temperature'],
-            }
+    }
+    if system_prompt:
+        request_body['system'] = [{'text': system_prompt}]
```
noel-improv: Non-blocking: `messages_to_converse_messages` already returns `{'text': ...}` dicts, so this serialises as `"system": [{"text": [{"text": ...}]}]`. Pre-existing and moved verbatim, but the claude builder just below handles its return correctly.

lexical-graph/tests/unit/indexing/utils/test_batch_inference_utils.py
```diff
@@ -393,6 +396,43 @@ def test_parse_output_empty_content(self):
...
         assert result == ''
+
+
+class TestBatchModelProviderRegistry:
+    """Tests for the extensible provider registry backing batch dispatch."""
+
+    def test_registry_contains_expected_families(self):
+        """The registry holds exactly the supported families; adding one is a single entry."""
+        names = {provider.name for provider in BATCH_MODEL_PROVIDERS}
+        assert names == {'amazon.nova', 'anthropic.claude', 'meta.llama'}
```
noel-improv: Suggestion: equality means every new provider needs a test edit, which cuts against the docstring right above it. Containment keeps the check without the coupling.

.github/workflows/llama-index-deps-check.yml
```diff
@@ -0,0 +1,81 @@
...
+        with:
+          python-version: "3.12"
+
+      - name: Install packaging
+        run: python -m pip install --disable-pip-version-check packaging
+
+      - name: Check PyPI for newer llama-index releases
+        id: check
+        env:
+          REQUIREMENTS_FILES: >-
```
noel-improv, line 37: Nit: `integration-tests/requirements-integ-test.txt` pins four llama-index packages and isn't listed here, so those floors go unchecked.

noel-improv (commented):
Nice work on this, just a couple of clarifying questions and a few nits

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/utils/batch_inference_utils.py
oussamahansal: it's pre-existing not introduced here but I'll fix it part of this PR

lexical-graph/tests/unit/indexing/utils/test_batch_inference_utils.py
oussamahansal: I'd like to keep equality here, It's a deliberate exact-set guard: it fails if a family is accidentally dropped or an unexpected one appears exactly the regression this test exists to catch. `issubset` would silently pass when a family is removed, which defeats the point.

byokg-rag/src/graphrag_toolkit/byokg_rag/requirements.txt
```diff
@@ -6,7 +6,7 @@ colorama==0.4.6
 faiss-cpu==1.9.0
 langchain_huggingface==0.3.1
 langchain_aws==0.2.33
-llama-index-embeddings-bedrock>=0.8.2
+llama-index-embeddings-bedrock>=0.9.0
```
acarbonetto, line 9: we should make this a separate issue

lexical-graph/src/graphrag_toolkit/lexical_graph/requirements.txt
```diff
@@ -4,10 +4,10 @@ anthropic-bedrock==0.8.0
 boto3>=1.40.61
 botocore>=1.40.61
 json2xml==5.2.0
-llama-index-core>=0.14.23
-llama-index-embeddings-bedrock>=0.8.2
-llama-index-llms-anthropic>=0.11.8
-llama-index-llms-bedrock-converse>=0.14.16
+llama-index-core>=0.14.24
```
acarbonetto, line 7: these should be done in a separate PR

.github/scripts/check_llama_index_updates.py
```diff
@@ -0,0 +1,160 @@
...
+REQ_LINE = re.compile(r'^(llama-index[a-z0-9-]*)\s*>=\s*([0-9][0-9A-Za-z.\-]*)\s*$')
+
+PYPI_URL = 'https://pypi.org/pypi/{pkg}/json'
+
+
+def latest_version(pkg: str) -> str:
+    """Return the latest (info.version) release of a package on PyPI."""
+    with urllib.request.urlopen(PYPI_URL.format(pkg=pkg), timeout=30) as resp:
+        data = json.load(resp)
+    return data['info']['version']
```
acarbonetto: confirm that this release isn't a pre-release, or yanked.

lexical-graph/tests/unit/indexing/utils/test_batch_inference_utils.py
```diff
@@ -367,14 +372,24 @@ def test_parse_output_claude_model(self):
     def test_parse_output_llama_model(self):
         """Verify parsing function works for Llama model output."""
         parse_fn = get_parse_output_text_fn('meta.llama3-70b-instruct-v1:0')
-        
+
+        # Bedrock wraps the provider payload under 'modelOutput', same as nova/claude.
```
acarbonetto: - [ ] confirm this change works with real data. 

We should manually test this (or integration test this) before adopting this change.

noel-improv (approved):
LGTM. Just one non-blocking note: `dependabot.yml` has no pip entry for `/integration-tests` the report's "_Dependabot opens the actual bump PRs_" line won't hold for the four integ-test packages, and they're already stale enough to trip the first run. Do we have this planned as a follow-up at least?

PR #521, opened 2026-09-04
[FEATURE] Split release changelog by project (lexical-graph | byokg | both)

oussamahansal:
## Description

Add a path-based changelog generator that classifies each commit by the project folders it touched, so per-project releases no longer list the other project's changes.

## Changes

- classify commits by changed top-level folders
- `release_message.sh` now delegates to the generator for local previews (`release_message.sh lexical-graph|byokg|both`)
- Unit tests

## Problem

Related issue (if any): #

## Testing

## Checklist

---

### DIFF
--- a/.github/scripts/generate_changelog.py
+++ b/.github/scripts/generate_changelog.py
@@ -1,0 +1,238 @@
+#!/usr/bin/env python3
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""Generate a release changelog grouped/filtered by project.
+
+The toolkit is a monorepo that releases ``lexical-graph`` and ``byokg`` one at a
+time, but the changelog historically listed every commit since the previous tag
+regardless of project. This script classifies each commit in a git range into
+``lexical-graph``, ``byokg``, or ``both`` by the top-level folders it changed,
+then emits markdown release notes for the requested project.
+
+Classification (by changed path):
+    * ``lexical-graph/``, ``lexical-graph-contrib/``, ``examples/lexical-graph*``
+      -> lexical-graph
+    * ``byokg-rag/``, ``examples/byokg*`` -> byokg
+    * anything else (docs-site, benchmarks, integration-tests, .github, root, ...)
+      is "shared" and reported under ``both`` since it can affect either package
+
+A commit that touches both packages is ``both``; a commit that touches only
+shared paths is also ``both``. A single-project changelog includes that
+project's commits plus the ``both`` commits (those changes ship in the release
+too).
+
+Range: pass ``--from``/``--to`` explicitly, or let the script pick the previous
+released tag of the same project (``.dev`` prereleases are ignored as baselines)
+as the start and ``--to`` (default ``HEAD``) as the end.
+
+Examples:
+    generate_changelog.py --project lexical-graph
+    generate_changelog.py --project byokg --to graphrag-byokg/v3.19.1
+    generate_changelog.py --project both --from graphrag-lexical-graph/v3.19.0
+"""
+
+import argparse
+import subprocess
+import sys
+from typing import Dict, List, NamedTuple, Optional, Set
+
+LEXICAL = 'lexical-graph'
+BYOKG = 'byokg'
+BOTH = 'both'
+PROJECTS = (LEXICAL, BYOKG, BOTH)
+
+# Top-level path prefixes that map a changed file to a package. Order matters
+# only in that the lexical-graph-contrib prefix is listed explicitly (it is not
+# under 'lexical-graph/'). str.startswith accepts a tuple of prefixes.
+LEXICAL_PREFIXES = ('lexical-graph/', 'lexical-graph-contrib/', 'examples/lexical-graph')
+BYOKG_PREFIXES = ('byokg-rag/', 'examples/byokg')
+
+# Human-readable section headings used when rendering the 'both' report.
+SECTION_TITLES = {
+    LEXICAL: 'lexical-graph',
+    BYOKG: 'byokg',
+    BOTH: 'Shared / both projects',
+}
+
+# Field/record separators for the machine-readable git log format. Using control
+# characters avoids collisions with anything in a commit subject.
+_REC = '\x1e'   # record separator: one per commit
+_FLD = '\x1f'   # field separator: between sha / short-sha / subject
+
+
+class Commit(NamedTuple):
+    sha: str
+    short_sha: str
+    subject: str
+    files: List[str]
+
+
+def bucket_for_path(path: str) -> str:
+    """Map one changed file path to 'lexical', 'byokg', or 'shared'."""
+    if path.startswith(LEXICAL_PREFIXES):
+        return 'lexical'
+    if path.startswith(BYOKG_PREFIXES):
+        return 'byokg'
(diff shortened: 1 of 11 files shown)


mykola-pereyma (approved):
LGTM — nice, well-tested per-project changelog tooling, and thanks for resolving the RELEASE.md TODOs. No shell-injection concerns (git via `subprocess` list form; the ref passed as a quoted env var rather than a `${{ }}` expansion), least-privilege on the new test workflow, and strong unit coverage of the classifier.

One non-blocking thing worth a follow-up: **the changelog append isn't idempotent.**

`softprops/action-gh-release` with `append_body: true` appends `RELEASE_CHANGELOG.md` to the release body on **every** execution. These workflows run on `release: [released|prereleased]` events, which can be re-run from the Actions UI (or a `released` event can follow a `prereleased` one for the same release on the lexical-graph workflow). Each execution would append another copy of the generated changelog to the same release notes, so a retried/re-published release ends up with duplicated changelog blocks.

A couple of ways to make it safe under re-runs:
- Have the step compute the full release body (author notes + a single delimited changelog section) and **set** the body via the API rather than appending, replacing any prior auto-generated block (e.g. delimit with an HTML-comment marker like `<!-- auto-changelog:start -->…<!-- auto-changelog:end -->` and regenerate between the markers); or
- Guard the append so it only runs once per release (skip if the marker is already present in the release body).

Not blocking — approving as-is — but worth hardening before this is relied on for a release that gets re-run.

(Also minor, for later: `previous_release_tag` is the one logic-heavy piece without a unit test since it needs git — a small git-backed test over a temp repo with a couple of tags, including a `.dev` one, would lock in the baseline-selection behavior.)

.github/scripts/generate_changelog.py
```diff
@@ -0,0 +1,238 @@
...
+    for pattern in patterns:
+        tags.extend(
+            t for t in _git('tag', '--list', pattern, '--sort=-creatordate').splitlines() if t
+        )
+    # Newest first; drop the current ref and .dev prereleases (a final release
+    # should diff against the previous real release, not its own prerelease).
+    for tag in tags:
+        if tag == to_ref or '.dev' in tag:
+            continue
+        if _is_ancestor(tag, to_ref):
```
noel-improv: Potentially blocking: this only accepts a baseline that's an ancestor of `to_ref`, but we tag on release branches and squash-downmerge, so only `v3.18.5` qualifies — 3.18.6, 3.18.7, 3.19.0 and 3.19.1 all fail it. Tagging from main (RELEASE.md step 4) then picks v3.18.5 and emits 137 entries instead of 21. Does `git log A..B` need the ancestor filter at all?

.github/scripts/generate_changelog.py
```diff
@@ -0,0 +1,238 @@
...
+
+    tags: List[str] = []
+    for pattern in patterns:
+        tags.extend(
+            t for t in _git('tag', '--list', pattern, '--sort=-creatordate').splitlines() if t
+        )
+    # Newest first; drop the current ref and .dev prereleases (a final release
+    # should diff against the previous real release, not its own prerelease).
+    for tag in tags:
+        if tag == to_ref or '.dev' in tag:
```
noel-improv: Potentially blocking: this compares tag names, but `graphrag-byokg/v3.19.1` and `graphrag-lexical-graph/v3.19.1` are the same commit, so `--project both` picks its own sibling as the baseline and every section comes back `_No changes._`. Since `both` is `release_message.sh`'s default, could we compare `git rev-parse <tag>^{}` against `to_ref^{}` instead?

.github/workflows/lexical-graph-release.yml
```diff
@@ -73,6 +78,14 @@ jobs:
           echo "version=${VERSION}" >> $GITHUB_OUTPUT
           echo "pypi_version=${VERSION#v}" >> $GITHUB_OUTPUT
 
+      - name: Generate per-project changelog
```
noel-improv, line 81: Non-blocking: this runs early in `release-build` and `_git` uses `check=True`, and `pypi-publish` needs that job — so a release-notes hiccup would block the publish. Worth moving it next to `Update release notes`, or `continue-on-error: true`? Same placement in both byokg workflows.

.github/scripts/generate_changelog.py
```diff
@@ -0,0 +1,238 @@
...
+            continue
+        if _is_ancestor(tag, to_ref):
+            return tag
+    return None
+
+
+def read_commits(from_ref: Optional[str], to_ref: str) -> List[Commit]:
+    range_spec = f'{from_ref}..{to_ref}' if from_ref else to_ref
+    fmt = f'{_REC}%H{_FLD}%h{_FLD}%s'
+    out = _git('log', '--no-merges', f'--pretty=format:{fmt}', '--name-only', range_spec)
```
noel-improv: Nit: with default `core.quotepath`, `--name-only` quotes non-ASCII paths, so the leading `"` defeats the `startswith` checks and the commit buckets as `shared`. `-c core.quotepath=false` covers it.

.github/scripts/generate_changelog.py
```diff
@@ -0,0 +1,238 @@
...
+            continue
+        header, _, rest = record.partition('\n')
+        sha, short_sha, subject = header.split(_FLD)
+        files = [line for line in rest.split('\n') if line.strip()]
+        commits.append(Commit(sha=sha, short_sha=short_sha, subject=subject, files=files))
+    return commits
+
+
+def _git(*args: str) -> str:
+    result = subprocess.run(
```
noel-improv: Nit: stderr is captured and never surfaced, so a failure in a release job shows only `exit status 128` with no reason.

noel-improv (commented):
The changes for the path-based change-log generator looks cleanly implemented and workflows look decent just a few questions I added below the main ones revolve tag names.

.github/scripts/test_generate_changelog.py
```diff
@@ -0,0 +1,268 @@
...
+        _commit(repo, 'c1', '2020-01-01T00:00:00')
+        _tag(repo, 'graphrag-lexical-graph/v1.0.0', '2020-01-01T00:00:00')
+        _tag(repo, 'graphrag-byokg/v1.0.0', '2020-01-01T00:00:01')
+        _commit(repo, 'c2', '2020-01-02T00:00:00')
+        # The release commit carries both projects' tags (identical commit).
+        _tag(repo, 'graphrag-lexical-graph/v1.1.0', '2020-01-02T00:00:00')
+        _tag(repo, 'graphrag-byokg/v1.1.0', '2020-01-02T00:00:01')
+        monkeypatch.chdir(repo)
+
+        baseline = gc.previous_release_tag('both', 'graphrag-lexical-graph/v1.1.0')
```
noel-improv, line 267: Non-blocking: this passes against the pre-fix script too, so it doesn't yet guard the bug it names. The old code globbed each pattern separately and concatenated, lexical first, so with `to_ref` as the lexical tag the correct answer came up before the sibling.

Using the byokg tag here makes it fail on the old code and pass on the new one:

```
pre-fix:  AssertionError: 'graphrag-lexical-graph/v1.1.0' in (...v1.0.0, ...v1.0.0)
fixed:    1 passed
```

`test_non_ancestor_release_is_still_the_baseline` already fails correctly against the old code.

noel-improv (approved):
looks good, just one non-blocking comment raised

PR #522, opened 2026-09-04
[FIX] Surface a failed document upload instead of hanging S3 staging

### LINKED ISSUE #519: [BUG] S3BasedDocs staging hangs, or reports a failed upload as written
### Package version

3.19.1

### Package

lexical-graph

### Python version

3.12.13

### Operating System

macOS

### Description

A document that fails to upload during S3 staging either hangs the run or is reported as staged. Neither raises.

`_upload_batch` polls until `count == target_count` and treats `queue.Empty` as "keep waiting", without checking whether the producer thread is alive. Three `except Exception: log` blocks can put that count out of reach:

- `_get_callback_fn` puts nothing on the queue when `future.result()` raises
- `_submit_proxy` returns without releasing the semaphore, while `_doc_publisher` still runs `count += 1` for a document it never submitted
- `_doc_publisher` keeps `queue.put(count)` inside the `try`, so a failure above it skips the put

What keeps this from firing today is `_upload_doc` swallowing its own S3 exception and returning `None`. The callback puts the `None`, the count advances, and `accept()` yields it and counts it in the "Finished writing N source documents" total. So the failed document is reported as written, and a retry has no idea it needs to be redone.

That makes the two a pair: stop swallowing in `_upload_doc` on its own and the silent success becomes a hang.

Expected: an upload failure raises, and a document that failed to write is not counted as staged.

This is not #418. Same symptom and same path, but that one was a forked worker inheriting a held `BufferedWriter` lock, and it is fixed. This came in with `4b8743c0`.

### Steps to reproduce

```python
# 1. Producer dies -> the consumer never returns.
with patch.object(S3DocUploader, '_doc_publisher', side_effect=RuntimeError('boom')):
    list(uploader.upload([doc]))          # blocks indefinitely, no exception

# 2. The coupling. Against a working S3 stub, patch _upload_doc two ways:
#      returns None (today)  -> finished=True   yielded=[None]   <- failure reported as success
#      raises                -> finished=False  yielded=[]       <- hangs
```

### Other information

A fix would be one invariant: every submitted document puts exactly one item on the queue.

- `_get_callback_fn` always puts, a failure marker on exception
- `_submit_proxy` releases the semaphore in a `finally` and reports whether it submitted
- `_doc_publisher` puts the count in a `finally`, catching `BaseException`
- `_upload_batch` breaks on `queue.Empty` when the producer is dead, and re-raises after `join()`
- `_upload_doc` stops swallowing
- `accept()` excludes failures from its staged count

No test fails a document mid-batch, which is the gap that hid this.

noel-improv:
## Description

A failed document upload during S3 staging either hangs the run with no exception, or is reported as successfully staged. Both are silent, so a failed upload looks like a slow run or a clean one.

`_upload_batch` polls until it has seen one item per submitted document and treats `queue.Empty` as "keep waiting", never checking whether its producer is still alive. Three `except Exception: log` blocks could each leave that count unreachable.

What stops it firing today is `_upload_doc` swallowing its own S3 exception and returning `None`. The callback puts that `None`, the count advances, and the document is yielded and counted in the "Finished writing N source documents" total. The two defects are a pair: removing the swallow alone turns a silent success into a hang.

## Changes

One invariant — every submitted document puts exactly one item on the queue.

- `_get_callback_fn` always puts, a failure marker when the upload raised, and releases the semaphore in a `finally`.
- `_submit_proxy` reports whether it submitted, so the publisher counts only what the consumer will see.
- `_doc_publisher` puts its count in a `finally` and catches `BaseException`, so a producer that dies still releases the consumer.
- `_upload_batch` breaks out when the queue is empty and the producer is gone, and raises the first failure once the batch drains.
- `_upload_doc` stops swallowing.

The queue poll drops from 60s to 1s. It paces the liveness check rather than the work, so a dead producer is noticed in about a second instead of a minute.

## Problem

Related issue (if any): #

Introduced in `4b8743c0` ("Improved batch extract and S3BasedDocs for large ingests") rather than by a targeted change. #418 is the same symptom in the same path from a different cause and is already fixed. Also fixes #519.

## Testing

- [x] Existing tests pass (`pytest`) — 2,112 in `tests/unit`

Five tests, red before green, covering a failing document mid-batch, a dead producer, and the all-succeed path. No test previously failed a document mid-batch, which is the gap that hid this.

Checked against a real boto3 S3 client rather than a mock: `put_object` returns a genuine `ClientError`, the error propagates to the caller, no document is reported as staged, and the call returns instead of hanging.

## Checklist

A caller that relied on a failed upload being silently skipped will now see the exception. That is the intended change.

### DIFF
--- a/lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/load/s3_based_docs.py
+++ b/lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/load/s3_based_docs.py
@@ -225,7 +225,9 @@ class S3DocUploader(ConfiguredThreadCount, BaseComponent):
             # count it can never reach.
             try:
                 queue.put(future.result(timeout=1.0))
-            except Exception as e:
+            except BaseException as e:
+                # BaseException, matching _doc_publisher: an interrupt in a worker
+                # strands the consumer as surely as a ClientError does.
                 logger.error(f'Error uploading source document: {str(e)}')
                 queue.put(_UploadFailed(e))
             finally:
@@ -283,7 +285,7 @@ class S3DocUploader(ConfiguredThreadCount, BaseComponent):
 
         logger.debug(f'About to start polling queue [count: {count}, target_count: {target_count}]')
 
-        failure = None
+        failures:List[BaseException] = []
 
         while target_count is None or count < target_count:
             try:
@@ -293,22 +295,33 @@ class S3DocUploader(ConfiguredThreadCount, BaseComponent):
                 else:
                     count += 1
                     if isinstance(item, _UploadFailed):
-                        failure = failure or item.cause
+                        failures.append(item.cause)
                     else:
                         yield item
                 self._queue.task_done()
             except queue.Empty:
                 # A producer that died before putting its count would otherwise
-                # keep the consumer here for the rest of the run.
+                # keep the consumer here for the rest of the run. Reaching here
+                # means the batch is short, so it is a failure rather than an end.
                 if not thread.is_alive():
+                    failures.append(RuntimeError(
+                        f'Upload producer stopped before reporting its document count '
+                        f'[count: {count}, target_count: {target_count}]'
+                    ))
                     break
             
         logger.debug(f'Waiting on queue to empty [count: {count}, target_count: {target_count}]')
 
         thread.join()
 
-        if failure is not None:
-            raise failure
+        if failures:
+            # The first is raised rather than aggregated: 3.10 has no ExceptionGroup.
+            if len(failures) > 1:
+                logger.error(
+                    f'{len(failures)} source documents failed to upload, raising the first '
+                    f'[batch_size: {len(source_docs_batch)}]'
+                )
+            raise failures[0]
 
     def upload(self, source_documents: List[SourceDocument]):
 
--- a/lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/load/s3_based_docs.py
+++ b/lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/load/s3_based_docs.py
@@ -31,11 +31,26 @@ from llama_index.core.bridge.pydantic import PrivateAttr
 QUEUE_SIZE = 1000
 BATCH_SIZE = 100
 
+# How often the staging consumer wakes to check whether its producer is still
+# alive. Uploads are slow, so this paces the liveness check, not the work.
+QUEUE_POLL_SECONDS = 1.0
+
 logger = logging.getLogger(__name__)
 
 # Joins node ids before hashing them. Without a separator ['ab', 'c'] and
 # ['a', 'bc'] hash alike.
 _DOC_SUFFIX_DELIMITER = '\x00'
+
+class _UploadFailed:
+    """
+    Stands in for a document that did not upload.
+
+    The consumer counts one item per submitted document, so a failure has to
+    occupy a slot rather than being dropped.
+    """
+
+    def __init__(self, cause:BaseException):
+        self.cause = cause
 
 class ConfiguredThreadCount:
     """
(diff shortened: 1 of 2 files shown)


lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/load/s3_based_docs.py
```diff
@@ -31,12 +31,27 @@
 QUEUE_SIZE = 1000
 BATCH_SIZE = 100
 
+# How often the staging consumer wakes to check whether its producer is still
+# alive. Uploads are slow, so this paces the liveness check, not the work.
+QUEUE_POLL_SECONDS = 1.0
```
mykola-pereyma: why not MILLISECONDS in integer??

mykola-pereyma (approved):
LGTM

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/load/s3_based_docs.py
```diff
@@ -201,54 +216,63 @@ def _upload_doc(self, root_path:str, doc:SourceDocument, s3_client):
...
     def _get_callback_fn(self, queue:queue.Queue):
         def _task_complete_callback(future):
+            # The consumer counts one item per submitted document, so a failure
+            # has to put something too. Putting nothing leaves it polling for a
+            # count it can never reach.
             try:
-                doc = future.result(timeout=1.0)
-                queue.put(doc)
+                queue.put(future.result(timeout=1.0))
             except Exception as e:
```
acarbonetto: should this catch BaseException like we do in [_doc_publisher](https://github.com/noel-improv/graphrag-toolkit/blob/ce141cf2626c11c453e2fb4ab1e7ad461e934e64/lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/load/s3_based_docs.py#L270)?

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/load/s3_based_docs.py
```diff
@@ -259,22 +283,33 @@ def _upload_batch(self, source_docs_batch:List[SourceDocument]):
...
             try:
-                item = self._queue.get(timeout=60.0)
+                item = self._queue.get(timeout=QUEUE_POLL_SECONDS)
                 if isinstance(item, int):
                     target_count = item
                 else:
                     count += 1
-                    yield item
+                    if isinstance(item, _UploadFailed):
+                        failure = failure or item.cause
```
acarbonetto: Only the first upload failure in a batch is surfaced, can we store and report all document failures?

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/load/s3_based_docs.py
```diff
@@ -259,22 +283,33 @@ def _upload_batch(self, source_docs_batch:List[SourceDocument]):
...
+                    else:
+                        yield item
                 self._queue.task_done()
-            except queue.Empty as e:
-                continue
+            except queue.Empty:
+                # A producer that died before putting its count would otherwise
+                # keep the consumer here for the rest of the run.
+                if not thread.is_alive():
```
acarbonetto, line 312: breaking out here leaves failure as None and count < target_count, so the generator just ends normally. The if failure is not None: raise failure below is skipped, and upload() goes on to report "Finished writing N source documents" for a batch that was actually truncated.

That's the exact silent partial-success mode this PR set out to eliminate — it's just narrowed to the abnormal path where the producer dies before publishing its count (e.g. hard-killed thread, or put(count) itself never completing).

Suggest raising instead of returning silently when we break on a dead producer:

```
except queue.Empty:
    # A producer that died before putting its count would otherwise
    # keep the consumer here for the rest of the run.
    if not thread.is_alive():
        if failure is None:
            failure = RuntimeError(
                f'Upload producer died before completing '
                f'[count: {count}, target_count: {target_count}]'
            )
        break
```

That way a truncated batch surfaces as an error rather than a success, keeping the guarantee consistent with the normal-path _UploadFailed handling.

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/load/s3_based_docs.py
```diff
@@ -201,54 +216,65 @@ def _upload_doc(self, root_path:str, doc:SourceDocument, s3_client):
...
             logger.exception(f'Error in submit proxy: {str(e)}')
+            self._semaphore.release()
+            return False
+
+        future.add_done_callback(self._get_callback_fn(queue))
+        return True
     
     def _doc_publisher(self, queue:queue.Queue, source_documents:List[SourceDocument]=[]):
 
         s3_client = GraphRAGConfig.s3
```
oussamahansal: The comment at line 275 says the count is enqueued "whatever happens above, including a KeyboardInterrupt," but `s3_client = GraphRAGConfig.s3` sits above the `try`, so a client-init failure (bad creds/region) skips the `finally` and surfaces only as the generic "producer stopped before reporting its document count" `RuntimeError` — with the real cause dropped, since it isn't chained (`from e`) and lives in the daemon thread. Chain the producer's exception onto that `RuntimeError` so the cause isn't lost.

oussamahansal (commented):
Few comments:
- The fail-loud fix only covers S3DocUploader (the for_jsonl=True path). The default for_jsonl=False path uses S3ChunkUploader, whose _drain still does except Exception: logger.error(...) and swallows the failure — so the silent partial-success this PR set out to kill still happens on the default upload path, and test_a_failed_chunk_does_not_stop_the_stream even asserts the failing doc is still yielded. Intentional, or should the chunk path get the same treatment?

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/load/s3_based_docs.py
noel-improv: Good catch and it went further than the init line. Fixed: the producer re-raises after publishing its count, and the consumer raises that error rather than a generic one. There are tests for all three cases.

On the chunk path: this is intentional. This PR is narrow so it covers the jsonl uploader, where the hang primarily was. The chunk path has no queue to strand, and its silent-failure side is tracked on #532: the document still yields, but it gets no completion marker, _Restart_ can tell it's incomplete.

oussamahansal (approved):

PR #529, opened 2026-09-09
[FIX]: make entity linking deterministic across runs

oussamahansal:
## Description

Entity linking is nondeterministic. The same query on the same graph with the same `topk` returns different seed sets between runs — and sometimes a different *number* of seeds. On the demo KG, 5 identical runs produced 5 distinct seed sets.

## Changes

- `fuzzy_string.py` — `sorted(set(...))` for a hash-independent vocab order
- `byokg_query_engine.py` — `list(dict.fromkeys(...))` for the seed union (order-preserving dedup) and `sorted(explored_entities)` for the path retriever input.

## Problem

Related issue (if any): #

## Testing

## Checklist

---

### DIFF
--- a/byokg-rag/src/graphrag_toolkit/byokg_rag/indexing/fuzzy_string.py
+++ b/byokg-rag/src/graphrag_toolkit/byokg_rag/indexing/fuzzy_string.py
@@ -78,7 +78,10 @@ class FuzzyStringIndex(Index):
         :param vocab_list: list of vocab instances to add
 
         """
-        self.vocab = list(set(self.vocab) | set(vocab_list))
+        # sorted() gives a hash-independent vocab order; process.extract breaks
+        # equal-score ties by vocab position, so a stable order is what makes
+        # matching reproducible.
+        self.vocab = sorted(set(self.vocab) | set(vocab_list))
 
     def add_with_ids(self, ids, vocab_list):
         raise NotImplementedError(f"add_with_ids not implemented for {self.__class__.__name__}")
--- a/byokg-rag/src/graphrag_toolkit/byokg_rag/byokg_query_engine.py
+++ b/byokg-rag/src/graphrag_toolkit/byokg_rag/byokg_query_engine.py
@@ -225,8 +225,10 @@ class ByoKGQueryEngine:
             if "draft-answer-generation" in artifacts and artifacts["draft-answer-generation"]:
                 linked_answers = self.entity_linker.link(artifacts["draft-answer-generation"], return_dict=False)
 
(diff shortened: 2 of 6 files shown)


byokg-rag/src/graphrag_toolkit/byokg_rag/indexing/fuzzy_string.py
```diff
@@ -78,7 +78,10 @@ def add(self, vocab_list):
         :param vocab_list: list of vocab instances to add
 
         """
-        self.vocab = list(set(self.vocab) | set(vocab_list))
+        # sorted() gives a hash-independent vocab order; process.extract breaks
+        # equal-score ties by vocab position, so a stable order is what makes
+        # matching reproducible.
+        self.vocab = sorted(set(self.vocab) | set(vocab_list))
```
noel-improv, line 84: Confirmed the mechanism while reading this: `process.extract("Apple", ["Apple Inc", "Apple Ltd"])` returns `Apple Inc` at 90 and the reversed vocab returns `Apple Ltd` at 90, so position really was deciding the winner. On main the vocab order changes with every hash seed. Good fix.

Non-blocking, and outside this diff so I can't anchor it there: `graphstore/neptune.py:149` and `:202` have the same pattern, `set(self.node_type_to_property_mapping.values())` iterated straight into the Cypher, so the query text varies run to run. Results are unaffected since `OR` commutes, but it makes logs noisy and would defeat anything keyed on the query string later. Worth `sorted()` here too, or a follow-up?

noel-improv (approved):
Nice on this, just one non-blocker

mykola-pereyma (approved):
LGTM — correct root-cause fix for the entity-linking nondeterminism, and the tests actually prove it: `test_matching_is_deterministic_across_hash_seeds` runs the match in subprocesses across 5 `PYTHONHASHSEED` values and asserts identical hits, which is the right way to prove hash-independence rather than a run-once. Nice touch keeping the relevance order via `dict.fromkeys` for the seed union while using `sorted()` only where order is meaningless (explored_entities, the Neptune OR-predicates). Also folds in the `neptune.py:149/202` `sorted()` suggestion from the earlier review.

One coordination note (non-blocking): PR #540 (single_best_match) also edits `byokg_query_engine.py` in this same seed-union / entity-linking block, so whichever of #529/#540 merges second will likely need a rebase. The interaction is positive — #540's `link_grouped` reads the fuzzy index whose equal-score tie-break becomes deterministic precisely because of this PR's `sorted()` vocab — so it's just a matter of sequencing the two, not a conflict in intent.

Issue #505, opened 2026-08-24
[BUG] Semantic-guided read paths read chunk.value directly and break under an external ChunkStore


noel-improv:
**Package version:** 3.18.5
**Package:** lexical-graph
**Python version:** 3.12
**OS:** Linux

## Description

#437 moved the traversal-based read paths onto `ChunkStore`. The semantic-guided stack still reads `chunk.value` off the graph, so it breaks once a deployment configures an external chunk store and `ChunkGraphBuilder` stops writing that property (`indexing/build/chunk_graph_builder.py:80-88`).

Two queries fetch chunk nodes with no property list, which defaults to `['*']` (`storage/graph/graph_utils.py:130`) and drags the full text back on every call:

- `retrieval/utils/chunk_utils.py:60` — `get_chunks_query`
- `retrieval/utils/statement_utils.py:124` — `get_statements_query`

Three consumers index `['value']` on the result and raise `KeyError` once the property is gone:

- `retrieval/retrievers/deprecated/semantic_guided_chunk_retriever.py:173`
- `retrieval/retrievers/deprecated/semantic_guided_retriever.py:218`
- `retrieval/post_processors/statement_enhancement.py:103`, which feeds that text to the LLM as synthesis context

`StatementEnhancementPostProcessor` is not deprecated, and it ships in the docs, an example notebook, and the integration tests.

#437 skipped `get_chunks_query` on purpose: "left as-is since its only caller is deprecated." That held while the property was still written.

## Steps to reproduce

Needs the S3 backend from #465. `InGraphChunkStore` is the only store on `main` and it still writes `chunk.value`.

Set `GraphRAGConfig.s3_chunk_store`, index a document set, then query through `SemanticGuidedRetriever` — with `StatementEnhancementPostProcessor` enabled, or with a chunk retriever. Both hit `KeyError: 'value'`.

## Expected

The semantic-guided paths resolve chunk text through `ChunkStore`, the way `keyword_vss_provider.py` and `traversal_based_base_retriever.py` do.

## Proposed fix

- Wire `statement_enhancement.py` to `ChunkStore`.
- Wire the semantic-guided retrievers, or delete them along with `get_chunks_query`, which has no other caller.
- Name the properties in whichever `node_result` calls survive, so these queries stop pulling text they may not use.

This should land before an external chunk store becomes the default.

PR #531, opened 2026-09-09
[FIX] Resolve statement enhancement context through the chunk store

noel-improv:
## Description

`StatementEnhancementPostProcessor` read chunk text straight off the node:

```python
context=node.node.metadata['chunk']['value'],
```

Moving chunk text out of the graph is the point of the ChunkStore work, so this path stops working once the text leaves. The traversal retrievers and `KeywordVSSProvider` already resolve through `ChunkStore.get_batch()`; this one did not.

The failure was also not reported. Reading the key directly raises `KeyError`, which the broad `except` around the call caught before returning the original node, so a failed enhancement could not be distinguished from a successful one.

Related issue: #505.

## Changes

- `StatementEnhancementPostProcessor` takes an optional `graph_store` and resolves a `ChunkStore` from it, following `traversal_based_base_retriever.py:125` and `keyword_vss_provider.py:53`.
- Chunk text a node does not carry is fetched with one `get_batch` per call rather than one fetch per node.
- A node with no statement, or no chunk text from either the node or the store, is returned unchanged and the model is not called.
- Two comments corrected. `chunk_cosine_search.py` and `deprecated/statement_cosine_seach.py` both stated that their placeholder text would be populated by a `StatementGraphRetriever`, which does not exist in the repository. Those nodes are not populated because they do not need to be: the caller reads the id from the metadata and discards the node.

Behaviour is unchanged when no `graph_store` is passed, and the class is only ever constructed by callers, so there are no internal construction sites to update.

## Testing

- [ ] Tested manually

27 new tests in `tests/unit/retrieval/post_processors/test_statement_enhancement.py`, covering text carried on the node, text resolved from the store, a store miss, absent chunk metadata, absent statement, and a response with no match. 820 pass across `tests/unit/retrieval` and `tests/unit/storage`.

## Checklist

Draft: the second read path in the related issue, `get_chunks_query` in `chunk_utils.py` feeding the deprecated semantic-guided retriever, is not covered here. Whether that path is connected to the ChunkStore or removed is a team decision, so this PR covers only the first path.

### DIFF
--- a/lexical-graph/src/graphrag_toolkit/lexical_graph/retrieval/post_processors/statement_enhancement.py
+++ b/lexical-graph/src/graphrag_toolkit/lexical_graph/retrieval/post_processors/statement_enhancement.py
@@ -47,6 +47,26 @@ class StatementEnhancementPostProcessor(BaseNodePostprocessor):
     user_prompt: str = Field(default=ENHANCE_STATEMENT_USER_PROMPT)
     enhance_template: ChatPromptTemplate = Field(default=None)
 
+    @staticmethod
+    def _resolve_chunk_store(graph_store) -> Optional[ChunkStore]:
+        """Open the chunk store this post-processor reads statement context from.
+
+        An external store configured through `S3_CHUNK_STORE` is opened whether or
+        not a graph store is supplied; a graph store only adds the in-graph
+        fallback for chunks written before the migration. Without either, chunk
+        text has to be carried on the node.
+
+        Raises ValueError if `S3_CHUNK_STORE` holds a URI no registered factory
+        recognises, so a misconfigured store surfaces at construction rather than
+        as unenhanced statements at query time.
+        """
+        chunk_store_info = GraphRAGConfig.s3_chunk_store
+
+        if not chunk_store_info and graph_store is None:
+            return None
+
+        return ChunkStoreFactory.for_chunk_store(chunk_store_info, graph_store=graph_store)
+
     def __init__(
         self,
         llm:LLMCacheType=None,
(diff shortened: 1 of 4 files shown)


mykola-pereyma (approved):
LGTM

lexical-graph/src/graphrag_toolkit/lexical_graph/retrieval/post_processors/statement_enhancement.py
```diff
@@ -81,40 +111,106 @@ def __init__(
...
+
+    @classmethod
+    def _chunk_id(cls, node: NodeWithScore):
+        return cls._chunk(node).get('chunkId')
+
+    @classmethod
+    def _chunk_text(cls, node: NodeWithScore):
+        return cls._chunk(node).get('value')
+
+    def _chunk_text_by_id(self, nodes: List[NodeWithScore]) -> dict:
```
acarbonetto, line 123: In _chunk_text_by_id's comprehension, the filter evaluates not self._chunk_text(node) and self._chunk_id(node), then the value expression calls self._chunk_id(node) again — each rebuilds the metadata.get('chunk') or {} dict. Binding chunk = self._chunk(node) once per node avoids the repeated lookups.

acarbonetto (approved):

lexical-graph/src/graphrag_toolkit/lexical_graph/retrieval/post_processors/statement_enhancement.py
```diff
@@ -81,40 +111,106 @@ def __init__(
...
+        """
+        Chunk text for the nodes that do not carry it, fetched in one call.
+
+        A node whose chunk text is a graph property arrives with it already in
+        place. The rest name a chunk id and nothing more, which is what an
+        external chunk store exists to resolve.
```
acarbonetto, line 129: Statement extraction lacks the None-guard the chunk path has

metadata.get('statement', {}).get('value') raises AttributeError when 'statement' is present but set to None. The new _chunk helper explicitly guards this exact None-vs-absent case for chunks (... or {}), but statements don't get the same treatment. The result is caught by the broad except and logged as an error ("Error enhancing statement") instead of taking the clean debug "no statement" skip path — so the two cases are reported inconsistently. The author acknowledged the None-vs-absent distinction for chunks but not statements.

lexical-graph/src/graphrag_toolkit/lexical_graph/retrieval/post_processors/statement_enhancement.py
```diff
@@ -81,40 +111,103 @@ def __init__(
...
+    def _section(node: NodeWithScore, key: str) -> dict:
+        """A node's metadata section, or an empty dict.
+
+        A retriever can leave a section set to None rather than absent, which a
+        `get(key, {})` default does not catch.
+        """
+        return node.node.metadata.get(key) or {}
+
+    @classmethod
+    def _chunk_id(cls, node: NodeWithScore):
```
oussamahansal: _chunk_id has no caller anywhere in src or tests — _chunk_text_by_id and enhance_statement both read chunk.get('chunkId') inline off _section. if you don't think you would use it you can drop it

oussamahansal (approved):
LGTM

lexical-graph/src/graphrag_toolkit/lexical_graph/retrieval/post_processors/statement_enhancement.py
noel-improv: Good catch. It lost its last caller when `_section` came in. Removed

lexical-graph/src/graphrag_toolkit/lexical_graph/retrieval/post_processors/statement_enhancement.py
noel-improv: Done. The comprehension now reads each node's chunk section once, through `_section`.

lexical-graph/src/graphrag_toolkit/lexical_graph/retrieval/post_processors/statement_enhancement.py
noel-improv: The statement read now goes through the same `_section` guard as the chunk read, so a None statement takes the skip path. `test_a_none_statement_skips_rather_than_erroring` covers it.

Issue #344, opened 2026-06-23
Support cross-region Bedrock inference profiles in ProcessPoolExecutor extraction pipeline


shkim4u:
## Problem

When the deployment region (e.g., `ap-northeast-2` for Neptune DB) differs from the Bedrock model region (e.g., `us-west-2` for `us.anthropic.claude-*` inference profiles), the extraction pipeline fails with:

```
botocore.errorfactory.ValidationException: An error occurred (ValidationException) 
when calling the Converse operation: The provided model identifier is invalid.
```

## Root Cause

`pipeline_utils.py` uses `ProcessPoolExecutor` which pickles LLM objects to send them to worker processes:

```python
# graphrag_toolkit/lexical_graph/indexing/utils/pipeline_utils.py
with ProcessPoolExecutor(max_workers=num_workers) as p:
    processed_node_batches = p.map(transform, node_batches)
```

During pickle:
1. `_client` (boto3 client) is removed ("Removing unpickleable private attribute _client")
2. On unpickle in the worker process, a new boto3 client is created using the default `AWS_REGION` environment variable
3. The configured `region_name` from `BedrockConverse(region_name="us-west-2")` is lost
4. The worker calls Bedrock in the wrong region → model ID is invalid in that region

## Current Workaround

Use `global.*` inference profiles (e.g., `global.anthropic.claude-sonnet-4-6`) which work in all regions, eliminating the need for cross-region calls. However, this limits model selection to globally available profiles only.

## Suggested Solutions

1. **Add a `use_threads` option** to `GraphRAGConfig` that switches `ProcessPoolExecutor` → `ThreadPoolExecutor` in `pipeline_utils.py`. Thread pools share memory (no pickle), so `BedrockConverse` objects retain their `region_name` and `_client`. Since LLM extraction is I/O-bound (network calls), GIL is not a bottleneck.

2. **Preserve `region_name` across pickle** by implementing `__getstate__`/`__setstate__` on the LLM cache wrapper, or by storing `region_name` as a picklable attribute and recreating the client with the correct region on unpickle.

3. **Pass region via environment variable** that workers respect — e.g., a dedicated `BEDROCK_REGION` env var that `GraphRAGConfig.to_llm()` reads when recreating the LLM in worker processes.

## Environment

- `graphrag-lexical-graph==3.18.5`
- `llama-index-llms-bedrock-converse==0.14.6`
- Neptune DB in `ap-northeast-2`, Bedrock models in `us-west-2`
[...]

PR #535, opened 2026-09-10
[FIX] Preserve configured LLM in non-batch fallback extractors

### LINKED ISSUE #534: Non-batch fallback in BatchLLMPropositionExtractorSync/BatchTopicExtractorSync ignores configured LLM
## Description

When a node set passed to `BatchLLMPropositionExtractorSync` or `BatchTopicExtractorSync` falls below Bedrock's minimum batch size (100 records), `_process_nodes` in `batch_extractor_base.py` falls back to a non-batch extractor via `_run_non_batch_extractor`. Both implementations construct their fallback extractor without passing the configured `llm`, even though `self.llm` holds it correctly:

```python
# batch_llm_proposition_extractor_sync.py
def _run_non_batch_extractor(self, nodes):
    all_nodes = [node for node in nodes]
    extractor = LLMPropositionExtractor(
        prompt_template=self.prompt_template,
        source_metadata_field=self.source_metadata_field,
    )  # llm= omitted
```

```python
# batch_topic_extractor_sync.py
def _run_non_batch_extractor(self, nodes):
    all_nodes = [node for node in nodes]
    extractor = TopicExtractor(
        prompt_template=self.prompt_template,
        source_metadata_field=self.source_metadata_field,
        entity_classification_provider=self.entity_classification_provider,
        topic_provider=self.topic_provider,
    )  # llm= omitted
```

Both `LLMPropositionExtractor` and `TopicExtractor` accept `llm` and, when omitted, default to `GraphRAGConfig.extraction_llm`, which resolves to `DEFAULT_EXTRACTION_MODEL = 'us.anthropic.claude-sonnet-4-6'`.

## Impact

- **Hard failure outside a region where the `us.*` profile resolves.** We hit this in `eu-west-2`, with a `BatchConfig`-configured `eu.anthropic.claude-sonnet-4-6` LLM passed explicitly to the batch extractor. A worker whose chunk count landed under 100 fell back to the non-batch path, which then called Bedrock with `us.anthropic.claude-sonnet-4-6`:

  ```
  ValidationException: The provided model identifier is invalid. [Model config: {..., "model": "us.anthropic.claude-sonnet-4-6", ...}]
  ```

  Because the toolkit uses a plain `asyncio.gather` internally with no per-document fault isolation, this aborted our entire extraction run (`processed=0, failed=50`) despite the batch path having already completed successfully for the other 3 worker buckets.

- **Silent behaviour change even where it doesn't error** — any deployment with a non-default `extraction_llm` (different model, `max_tokens`, `temperature`, region, etc.) gets inconsistent behavior between the batch and non-batch paths for the same extraction run, with nothing logged to indicate the switch.

## Relationship to #344 / #347

This produces the same `ValidationException: The provided model identifier is invalid.` message as #344, but the mechanism is different:

- **#344** — `BedrockConverse` is pickled to a `ProcessPoolExecutor` worker; the recreated client loses the LLM's configured `region_name`. Fixed by #347 (pass `region_name=self.llm.region_name` when `LLMCache` recreates the client).
- **This issue** — the configured LLM is never passed to the fallback extractor at all, so `self.llm.region_name` is never even in play; the fallback ignores the caller's LLM outright and falls through to `GraphRAGConfig.extraction_llm`.

The `global.*` inference profile workaround suggested for #344 does not help here, since the fallback discards whatever LLM/profile was configured regardless.

## Reproduction

1. Configure `BatchLLMPropositionExtractorSync`/`BatchTopicExtractorSync` with an explicit `llm` pointing at a Bedrock inference profile only valid outside `us-*` (e.g. an `eu.*` cross-region profile), in a non-`us-*` region.
2. Run extraction with a node set whose per-worker chunk count for at least one `ProcessPoolExecutor` worker lands under `BEDROCK_MIN_BATCH_SIZE` (100).
3. That worker's `_process_nodes` takes the `_run_non_batch_extractor` branch, which calls Bedrock with `GraphRAGConfig.extraction_llm` (`us.anthropic.claude-sonnet-4-6`) instead of the configured LLM.

## Suggested fix

Pass `llm=self.llm` through in both `_run_non_batch_extractor` implementations. PR incoming: fixes this with a one-line change to each file plus regression tests asserting the fallback extractor is constructed with the configured LLM.

## Environment

- `graphrag-lexical-graph==3.19.1` (also present on `main`)
- Region: `eu-west-2`
- Model: `eu.anthropic.claude-sonnet-4-6` (cross-region inference profile) configured via `BatchConfig`

gillespied:
## Description

`_run_non_batch_extractor` in `BatchLLMPropositionExtractorSync` and `BatchTopicExtractorSync` builds its fallback extractor without passing the configured `llm`, so it silently defaults to `GraphRAGConfig.extraction_llm` (`us.anthropic.claude-sonnet-4-6`) instead of whatever LLM was explicitly configured.

Fixes #534.

## Changes

- `batch_llm_proposition_extractor_sync.py`: pass `llm=self.llm` to the fallback `LLMPropositionExtractor`
- `batch_topic_extractor_sync.py`: pass `llm=self.llm` to the fallback `TopicExtractor`
- Regression tests for both classes asserting the fallback extractor is constructed with the extractor's configured `llm`

## Problem

Related issue: #534

When a node set falls below Bedrock's minimum batch size (100 records), both batch extractors fall back to a non-batch extractor. That fallback ignores `self.llm` entirely, so it uses `GraphRAGConfig.extraction_llm` (`DEFAULT_EXTRACTION_MODEL = 'us.anthropic.claude-sonnet-4-6'`) regardless of what LLM was configured. Outside a region where that `us.*` profile resolves, this raises `ValidationException: The provided model identifier is invalid.` We hit this in `eu-west-2` with an explicitly configured `eu.anthropic.claude-sonnet-4-6` LLM.

This is a different mechanism from #344/#347 (pickled `BedrockConverse` losing `region_name` in a spawned worker) — here the configured LLM is never passed to the fallback at all, so the `global.*` inference profile workaround for #344 doesn't apply.

## Testing

- [x] Unit tests added/updated — one regression test per class, patching the target extractor class and asserting it's constructed with `llm=<the configured llm>`
- [ ] Integration tests added (as appropriate) — not applicable, this only affects extractor construction, no AWS call needed to reproduce
- [x] Existing tests pass (`pytest`) — ran `tests/unit/indexing/extract/test_batch_llm_proposition_extractor_sync.py` and `test_batch_topic_extractor_sync.py` (11 passed), and the full `tests/unit` suite (2137 passed; one pre-existing, unrelated error in `test_integ_dependency_compatibility.py` caused by no `pip` binary in my venv, reproduced identically on `main` without this change)

Manually reproduced against a real `eu-west-2` account: a `BatchConfig`-configured `eu.anthropic.claude-sonnet-4-6` extraction, where one `ProcessPoolExecutor` worker's chunk count landed under 100 and hit this exact fallback path, raising the `ValidationException` quoted in #534.

## Checklist

- [x] License headers present on new files (no new files; existing headers preserved)
- [ ] Documentation updated (if applicable) — not applicable, internal fallback behavior only
- [x] No breaking changes (or clearly documented) — additive `llm=` kwarg passthrough only

### DIFF
--- a/lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/extract/batch_llm_proposition_extractor_sync.py
+++ b/lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/extract/batch_llm_proposition_extractor_sync.py
@@ -67,6 +67,7 @@ class BatchLLMPropositionExtractorSync(BatchExtractorBase):
         all_nodes = [node for node in nodes]
 
         extractor = LLMPropositionExtractor(
+            llm=self.llm,
             prompt_template=self.prompt_template, 
             source_metadata_field=self.source_metadata_field
         )
--- a/lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/extract/batch_topic_extractor_sync.py
+++ b/lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/extract/batch_topic_extractor_sync.py
@@ -80,6 +80,7 @@ class BatchTopicExtractorSync(BatchExtractorBase):
         all_nodes = [node for node in nodes]
 
         extractor = TopicExtractor( 
+            llm=self.llm,
             prompt_template=self.prompt_template, 
             source_metadata_field=self.source_metadata_field,
             entity_classification_provider=self.entity_classification_provider,
--- a/lexical-graph/tests/unit/indexing/extract/test_batch_llm_proposition_extractor_sync.py
+++ b/lexical-graph/tests/unit/indexing/extract/test_batch_llm_proposition_extractor_sync.py
@@ -70,3 +70,44 @@ class TestBatchLLMPropositionExtractorSyncUpdateNode:
         result = extractor._update_node(node, node_metadata_map)
 
         assert result.metadata[PROPOSITIONS_KEY] == []
+
+
+class TestBatchLLMPropositionExtractorSyncRunNonBatchExtractor:
+    """Tests for _run_non_batch_extractor method.
+
+    Regression tests: when a node set falls below Bedrock's minimum batch
+    size, BatchLLMPropositionExtractorSync falls back to the non-batch
+    LLMPropositionExtractor. That fallback must reuse the configured LLM
+    (self.llm) rather than silently defaulting to
+    GraphRAGConfig.extraction_llm, which is a us.* inference profile and
+    is invalid in any other Bedrock region.
+    """
+
+    def _make_extractor(self, llm, prompt_template="prompt", source_metadata_field=None):
+        """Create a BatchLLMPropositionExtractorSync instance with fields
+        populated via model_construct, bypassing validation/__init__ so no
+        real BatchConfig or AWS setup is needed for this unit test."""
+        return BatchLLMPropositionExtractorSync.model_construct(
+            llm=llm,
+            prompt_template=prompt_template,
+            source_metadata_field=source_metadata_field,
+        )
+
+    @patch("graphrag_toolkit.lexical_graph.indexing.extract.batch_llm_proposition_extractor_sync.LLMPropositionExtractor")
+    def test_run_non_batch_extractor_passes_configured_llm(self, mock_extractor_cls):
+        """Verify the non-batch fallback is constructed with the extractor's
+        own configured llm, not left to default to GraphRAGConfig.extraction_llm."""
+        configured_llm = Mock(name="configured-llm")
+        extractor = self._make_extractor(llm=configured_llm)
+
+        mock_instance = mock_extractor_cls.return_value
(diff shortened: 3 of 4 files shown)


noel-improv (approved):
@gillespied  thanks for your contribution so far these looks good. Both extractors take `LLMCacheType`, so `self.llm` is the right thing to pass here.

mykola-pereyma (approved):
Thank you for the contribution! Clean, well-scoped fix — passing `llm=self.llm` into both non-batch fallbacks correctly preserves the configured inference profile, and the per-class regression tests pin the behavior nicely. Verified these are the only two fallback sinks (no non-sync siblings share the omission). LGTM.

oussamahansal (approved):
LGTM

Issue #533, opened 2026-09-10
[BUG] Two documents sharing a truncated source id merge into one __Source__ node


noel-improv:
### Package version

3.19.1

### Package

lexical-graph

### Python version

3.12.13

### Operating System

macOS

### Description

`create_source_id` truncates the text digest to 8 hex characters (`id_generator.py:82`):

```python
return f"aws::{self._get_hash(text)[:8]}:{self._get_hash(metadata_str)[:4]}"
```

`IdRewriter` passes `''` for the metadata component when a node carries no metadata (`id_rewriter.py:74`), which makes the second component constant. A corpus loaded without metadata therefore discriminates on 32 bits, and two documents collide sooner than the width suggests.

Every downstream identity derives from the source id: the S3 storage prefix, the `__Source__` node the graph MERGEs on, and every chunk, topic, statement and fact id. Nothing detects or reports a collision.

### Reproduction

Hashing sequentially numbered documents until two share a digest takes 30,059 of them:

```python
from graphrag_toolkit.lexical_graph.indexing.id_generator import IdGenerator

A = 'document 27347 body text'
B = 'document 30059 body text'

g = IdGenerator()
assert A != B
assert g.create_source_id(A, '') == g.create_source_id(B, '') == 'aws::a4439cdb:d41d'
```

Both documents then land under one S3 prefix. `S3DocDownloader._download_doc` merges every object under a prefix into one `SourceDocument`, so the pair reads back as a single document holding both their chunks. In the graph, `source_graph_builder.py:79` is `MERGE (source:__Source__ {sourceId: params.sourceId})`, which attaches both to one node.

### Impact

Retrieval returns chunks belonging to a different source, and the two documents cannot be distinguished. Nothing is logged and no error is raised, so a merged document cannot be told apart from a correct one.

Measured with a generator that reproduces `IdGenerator` exactly, metadata absent:

| Documents | Colliding pairs |
|---|---|
| 100,000 | 1 |
| 1,000,000 | 127 |
| 10,000,000 | 11,762 |

With metadata present nothing collided at 10M, but 48 bits is that path's ceiling and 10M carries a 16% chance of at least one collision.

### Why the configurable width does not close this

#517 made the digest width a configuration setting and left the default at 8, so a default deployment behaves as it did before.

Truncating to a larger width lowers the probability but does not remove the case. Two changes are under consideration:

[...]

PR #536, opened 2026-09-11
[TEST] Reproduce the source id collision with a real colliding pair

noel-improv:
## Status

The fix is now up as #541, and it carries these two commits, so the reproducer can land either way. Review this PR for the tests on their own, or review #541 for the tests and the fix together. Merging #541 merges these commits and closes this PR with them.

Nothing here has changed since it was opened. One thing does change in #541: the fix removes the `xfail(strict=True)` marker below and pins the graph characterisation tests to the legacy width, since a graph written before the fix still merges the colliding pair.

## Description

A test that reproduces the source id collision described in #533, so the fix has something to turn green.

`create_source_id` truncates the text digest to 8 hex characters, and `IdRewriter` passes an empty string for the metadata component when a node carries no metadata, so a corpus loaded without metadata discriminates on 32 bits.

The pair below was found by hashing sequentially numbered documents until two shared a digest. It took 30,059 of them:

```
A = 'document 27347 body text'
B = 'document 30059 body text'
both -> aws::a4439cdb:d41d
```

No production code changes.

## Changes

Three groups of tests, each doing a different job.

`TestCollidingPair` pins the arithmetic at width 8. These are facts about md5 rather than statements about configuration, so they hold whatever a run is set to. One covers the metadata component, which discriminates only when metadata actually differs between documents.

`TestCollisionConsequences` records what the collision costs. One prefix reads back as a single SourceDocument holding both documents' chunks, and nothing logs or raises, so a merged document cannot be told apart from a correct one.

`TestSourceIdUniqueness` is the assertion a fix has to satisfy, marked `xfail(strict=True)`. It reads the configured width rather than a pinned one, so it asserts that the default is wide enough rather than anything about a particular width.

The download helper and the chunk node factory move into `tests/unit/conftest.py`. `test_s3_based_docs.py` already had a copy of both, so this removes one rather than adding a second.

## Why xfail(strict=True) rather than a failing test

A plain failing test cannot be committed. `xfail(strict=True)` records the red result while leaving CI green, and fails the run once the assertions start passing, so the marker cannot outlive the defect.

```
default                       6 passed, 2 xfailed
--runxfail                    2 failed, with the real assertion output
SOURCE_ID_HASH_LENGTH=16      XPASS(strict) on both
```

## Testing

- [ ] Tested manually

2141 pass, 2 xfailed across `tests/unit`. One pre-existing failure is unrelated and reproduces on a clean tree: `test_integ_dependency_compatibility.py` errors with `No module named pip` in this venv.

## Checklist

Related issue: #533.

### DIFF
--- a/lexical-graph/tests/unit/conftest.py
+++ b/lexical-graph/tests/unit/conftest.py
@@ -3,8 +3,10 @@
 
 import pytest
 from unittest.mock import Mock, patch, MagicMock
+from llama_index.core.schema import NodeRelationship, RelatedNodeInfo, TextNode
 from graphrag_toolkit.lexical_graph.tenant_id import TenantId
 from graphrag_toolkit.lexical_graph.indexing.id_generator import IdGenerator
+from graphrag_toolkit.lexical_graph.indexing.load.s3_based_docs import S3DocDownloader
 
 
 @pytest.fixture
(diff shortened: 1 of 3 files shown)


noel-improv:
Closing in favour of #541, which carries these two commits plus the fix, so the reproducer and the change that turns it green land together as one PR.

PR #538, opened 2026-09-11
[FEATURE] Make Bedrock batch model dispatch extensible

oussamahansal:
## Description

Makes Bedrock batch-inference model dispatch extensible. Batch inference uses the provider-specific InvokeModel JSONL schema (not Converse), so each model family needs its own request body and output parsing. This replaces the two hardcoded `if/elif` chains with a single `BATCH_MODEL_PROVIDERS` registry, so qualifying a new family is a small, well-tested change.

## Changes

- Registry (`BATCH_MODEL_PROVIDERS`): each family is one entry

## Problem

Related issue (if any): #

## Testing

## Checklist

- [ ] Code follows existing style and conventions

---

### DIFF
--- a/lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/utils/batch_inference_utils.py
+++ b/lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/utils/batch_inference_utils.py
@@ -6,7 +6,8 @@ import asyncio
 import time
 import os
 import json
-from typing import Any, List, Dict
+from typing import Any, Callable, List, Dict
+from dataclasses import dataclass
 from os import stat, listdir
 from os.path import isfile, join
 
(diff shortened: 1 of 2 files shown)


lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/utils/batch_inference_utils.py
```diff
@@ -65,47 +66,141 @@ def split_nodes(nodes: List[Any], batch_size: int) -> List[List[Any]]:
...
-            'parameters': {
-                'max_new_tokens': inference_parameters['max_tokens'],
-                'temperature': inference_parameters['temperature'],
-            }
+    }
+    if system_prompt:
+        # messages_to_converse_messages already returns the system prompt as a
+        # list of {'text': ...} blocks, so assign it directly (wrapping it again
+        # would nest as [{'text': [{'text': ...}]}]).
+        request_body['system'] = system_prompt
```
mykola-pereyma, line 104: Correctness check needed here. The original wrapped the prompt (`[{'text': system_prompt}]`); this assigns it directly, on the premise that `messages_to_converse_messages` already returns a list of `{'text': ...}` blocks (so the old code double-wrapped to `[{'text': [{'text': ...}]}]`).

That matches the Bedrock Converse `system` contract, but the Nova test **mocks** `messages_to_converse_messages`, so it only encodes this assumption — it wouldn't catch a mismatch with the real helper. If `llama-index-llms-bedrock-converse` (>=0.15.0) actually returns the system prompt as a plain **string**, this line regresses Nova (the original wrapping would then have been correct).

Please either confirm the real return type of `messages_to_converse_messages` for this pinned version, or add a test that exercises the actual helper (not a mock) so this can't silently regress.

mykola-pereyma (changes requested):
Really nice change — the `BATCH_MODEL_PROVIDERS` registry is a clean, extensible replacement for the two `if/elif` chains, and it also **fixes two real bugs** the original had (llama reading top-level `generation` instead of `modelOutput.generation`; the fail-loud text-mode parse), backed by good full-record tests. Requesting changes on three points before approving — one is a genuine correctness question, one is cross-PR coordination, one is hygiene.

**1. (blocking) Confirm the Nova `system` return type — this change isn't actually proven by the test.**
See the inline comment: switching `system` from `[{'text': system_prompt}]` to `system_prompt` is correct only if `messages_to_converse_messages` returns the system prompt as a list of `{'text': ...}` blocks. The added Nova test *mocks* that helper, so it encodes the assumption rather than verifying it. If the installed `llama-index-llms-bedrock-converse` (>=0.15.0) returns a plain string, this regresses Nova. Please confirm the real return type, or add a test that exercises the actual `messages_to_converse_messages` rather than a mock.

**2. (blocking) De-conflict with #520.**
#520 still contains an older, bug-preserving copy of this same dispatch refactor (alongside the dep bumps + the deps-check workflow) and overlaps on this file. If #520 merges after #538, it will silently re-introduce the llama/nova bugs this PR fixes. Please land the corrected refactor here and drop it from #520 (leaving #520 as deps + workflow), or otherwise sequence/rebase so this version is the one that wins.

**3. (non-blocking) Document the behavioral changes + complete the checklist.**
The description reads as a pure refactor, but this PR changes output-parsing (llama path, text-mode raise) and request-body (nova system) behavior. Please call those out as fixes, and complete the PR checklist — tests were added and behavior changed, but every box is currently unticked (including "No breaking changes").

The llama-path fix shows a correct understanding of the batch `modelOutput` envelope, so I lean toward the Nova change also being right — happy to flip to approve once (1) is confirmed and (2) is coordinated.

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/utils/batch_inference_utils.py
```diff
@@ -65,47 +66,141 @@ def split_nodes(nodes: List[Any], batch_size: int) -> List[List[Any]]:
...
+        'messages': anthropic_messages,
+        'max_tokens': params['max_tokens'],
+        'temperature': params['temperature']
+    }
+    if system_prompt:
+        request_body['system'] = system_prompt
+    return request_body
+
+
+def _build_llama_request(messages: List[ChatMessage], params: dict) -> dict:
```
noel-improv, line 139: Question, not a change request: this emits the SageMaker JumpStart shape (`messages` / `parameters.max_new_tokens`), but Bedrock's Meta InvokeModel takes `prompt` / `max_gen_len`. Inherited from main, so not something this PR broke — except the new `test_llama_full_record` fixture is the response half of the `prompt`/`max_gen_len` schema, so the entry now parses a response its own request can't produce.

Has a Llama batch job run end to end? If not, dropping the entry until it's validated may beat shipping a row nothing satisfies. Separately, `system_prompt` is unpacked here and never used, where nova and claude both attach it.

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/utils/batch_inference_utils.py
noel-improv, line 104: **[praise]**: main wrapped this as `[{'text': system_prompt}]` while `messages_to_converse_messages` already returns `{'text': ...}` blocks, so it was emitting `[{'text': [{'text': ...}]}]`.

Worth naming in the description though: this and the Llama output path both change the wire format, and the body reads as a pure registry refactor.

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/utils/batch_inference_utils.py
```diff
@@ -65,47 +66,141 @@ def split_nodes(nodes: List[Any], batch_size: int) -> List[List[Any]]:
...
-    else:
-        raise ValueError(f'Unrecognized model_id: batch extraction for {model_id} is not supported')
+    }
+
+
+@dataclass(frozen=True)
+class BatchModelProvider:
+    """A model family's batch (InvokeModel JSONL) request builder and output spec."""
+    name: str
+    match_prefixes: tuple
```
noel-improv: Nit: a bare `tuple` doesn't stop `match_prefixes=('cohere.command')` being a `str`, and `any(p in model_id for p in 'cohere.command')` then matches single characters — which would capture Nova and Claude ids. `Tuple[str, ...]` plus a check in `__post_init__` would catch the missing comma.

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/utils/batch_inference_utils.py
```diff
@@ -65,47 +66,141 @@ def split_nodes(nodes: List[Any], batch_size: int) -> List[List[Any]]:
...
+    build_request: Callable[[List[ChatMessage], dict], dict]
+    output_path: tuple
+    output_mode: str = 'blocks'
```
noel-improv, line 164: Nit: a typo here (`'scalar'`, `'string'`) falls through to the lenient `blocks` branch, which then iterates a string character by character and raises `AttributeError` at parse time — after the batch job has been paid for. Validating in `__post_init__` would move that to import time.

noel-improv (approved):

oussamahansal:
image

acarbonetto (approved):

PR #540, opened 2026-09-11
[FEATURE]  Reduce entity-linking noise by keeping only the best-matching node per mention

oussamahansal:
## Description

The entity linker unions every top-k candidate for every extracted mention into the retrieval seed set, pulling in low-relevance nodes: three exact-match mentions produce eight seeds on the demo KG.

Add an opt-in `single_best_match` flag (default off) on ByoKGQueryEngine that keeps only the best-matching node per extracted mention. Off by default, so the seed set is unchanged; draft answers are never pruned.

This needs per-mention candidate grouping, which the matcher path destroys by flattening and globally sorting hits across all mentions. Add EntityLinker.link_grouped, which queries the index once per mention so attribution is preserved; link() is unchanged

## Changes

- **`ByoKGQueryEngine(single_best_match=False)`** — new opt-in flag. When on, each extracted mention contributes only its best-matching node.
- **`EntityLinker.link_grouped(...)`** — new method returning candidates grouped per mention.

## Problem

Related issue (if any): #

## Testing

## Checklist

---


byokg-rag/src/graphrag_toolkit/byokg_rag/graph_retrievers/entity_linker.py
```diff
@@ -111,4 +110,44 @@ def link(self, query_extracted_entities, retriever=None, topk=None, id_selector=
...
+        # mentions) so attribution is preserved.
+        index = getattr(retriever, "index", None)
+        if index is None:
+            raise AttributeError(
+                "link_grouped requires an index-backed retriever (retriever.index is None)"
+            )
+
+        grouped = []
+        for mention in query_extracted_entities:
+            hits = index.query(mention, topk)["hits"]
```
mykola-pereyma: `link_grouped` queries `index.query` per mention, but the default `link()` path on an `EntityMatcher` uses `index.match`. These agree for the embedding indexes (dense/graph_store — `match` is just a batched loop of the same per-input search), but `FuzzyStringIndex.match` applies a `max_len_difference=4` short-string filter and a global re-sort that `index.query` does NOT.

So when `single_best_match` is used with a fuzzy matcher, the "best per mention" here can be a candidate that the normal union path would have filtered out — the refinement is ranked/filtered by a different function than the path it refines. Consider reusing the matcher's per-query logic (or applying the same length filter) so the two paths agree regardless of index type, or documenting that `single_best_match` assumes an embedding-backed matcher.

byokg-rag/src/graphrag_toolkit/byokg_rag/graph_retrievers/entity_linker.py
```diff
@@ -73,7 +73,7 @@ def __init__(self, retriever=None, topk=3, **kwargs):
         self.retriever = retriever
         self.topk = topk
 
-    def link(self, query_extracted_entities, retriever=None, topk=None, id_selector=None, return_dict=True):
+    def link(self, query_extracted_entities, retriever=None, topk=None, return_dict=True):
```
mykola-pereyma: The PR description says "link() is unchanged," but this signature drops the `id_selector` parameter. It was effectively dead here (never forwarded — the `EntityMatcher` path routes through `match()`, which has no `id_selector`), so runtime behavior is unchanged — but this is still a public API signature change: an external caller passing `link(..., id_selector=[...])` would now raise `TypeError`. Please correct the "unchanged" wording and call out the removed kwarg in the description/changelog.

mykola-pereyma (approved):
Nice, well-contained change — the `single_best_match` flag is opt-in and off by default, the existing seed set is provably unchanged when off (the test asserts `link_grouped` is never called and seeds are identical), and draft answers are never pruned. I also verified the parts the tests mock against the real index source: `index.query(input, topk=1, id_selector=None)` matches the positional call, and `document_id` is a scalar string in dense/graph_store/fuzzy, so the seed-set shape is correct. Approving.

Three recommendations (none blocking, but worth addressing — see inline comments for two of them):

**1. `link_grouped` (index.query) and `link()` (index.match) aren't equivalent for the fuzzy index.** For the embedding indexes (DenseFaiss, GraphStore), `match()` is a batched loop of the same per-input search as `query()`, so per-mention results agree. But `FuzzyStringIndex.match()` applies a `max_len_difference=4` short-string filter (and a global re-sort) that `query()` does not — so with a fuzzy matcher, `single_best_match` can select a "best" candidate the normal union path would have filtered out. Consider routing `link_grouped` through the matcher's own per-query logic (or applying the same length filter), or documenting that `single_best_match` assumes an embedding-backed matcher. A quick check: on a FuzzyStringIndex, compare `link()` vs `link_grouped` per mention and confirm the top candidate agrees.

**2. The description says "link() is unchanged," but `link()`'s signature dropped the `id_selector` kwarg.** That kwarg was dead (it was never forwarded — the EntityMatcher path uses `match()`, which has no `id_selector`), so behavior is unchanged, but it's still a public signature change: a caller passing `link(..., id_selector=[...])` would now `TypeError`. Please correct the wording and note the removed kwarg.

**3. Please complete the PR checklist** — unit tests were added, but every box is currently unticked.

Thanks for keeping this opt-in and leaving the default path untouched — that's the right call for a seed-set change.

byokg-rag/src/graphrag_toolkit/byokg_rag/graph_retrievers/entity_linker.py
noel-improv: Was dropping `id_selector` from `link()` intentional? It wasn't used, but removing it breaks callers that pass it by keyword, and a positional `link(entities, retriever, topk, False)` now sends `False` into `return_dict`. The docs in `graph-retrievers.mdx` still list it too. Could we keep the signature as is and leave that cleanup for a separate PR?

byokg-rag/src/graphrag_toolkit/byokg_rag/byokg_query_engine.py
```diff
@@ -217,7 +222,13 @@ def query(self, query: str, iterations: int = 2, cypher_iterations: int = 2, use
             # Process extracted entities
             linked_entities = []
             if "entity-extraction" in artifacts and artifacts["entity-extraction"] and "FINISH" not in artifacts["entity-extraction"][0]:
-                linked_entities = self.entity_linker.link(artifacts["entity-extraction"], return_dict=False)
+                if self.single_best_match:
+                    # Keep only the top candidate per mention (grouped preserves
+                    # attribution; link() would flatten across mentions).
+                    grouped = self.entity_linker.link_grouped(artifacts["entity-extraction"])
```
noel-improv: Non-blocking: `link_grouped` is only on `EntityLinker`, so a custom `Linker` subclass with `single_best_match=True` would raise an `AttributeError` here. Should `link_grouped` go on the `Linker` base class, or should `__init__` check the linker type when the flag is set?

byokg-rag/src/graphrag_toolkit/byokg_rag/byokg_query_engine.py
```diff
@@ -217,7 +222,13 @@ def query(self, query: str, iterations: int = 2, cypher_iterations: int = 2, use
...
+                    linked_entities = [candidates[0] for candidates in grouped if candidates]
```
noel-improv, line 243: With the fuzzy linker, tied candidates come back in set order (`FuzzyStringIndex.add` builds `vocab` from a `set`), so `candidates[0]` can change between runs. For the mention "Amazon", "Amazon Prime" and "Amazon Fresh" both score 90, and the pick flipped between `PYTHONHASHSEED=0` and `1`. Could ties be broken deterministically, say by node id?

byokg-rag/src/graphrag_toolkit/byokg_rag/byokg_query_engine.py
```diff
@@ -22,7 +22,8 @@ def __init__(self,
                  llm_generator=None,
                  kg_linker=None,
                  cypher_kg_linker=None,
-                 direct_query_linking=False):
+                 direct_query_linking=False,
+                 single_best_match=False):
```
noel-improv, line 26: Could `single_best_match` get a row in the constructor table in `docs-site/src/content/docs/byokg-rag/configuration.mdx`? The PR ticks "Documentation updated", but the docs-site doesn't mention the flag or `link_grouped` yet.

byokg-rag/src/graphrag_toolkit/byokg_rag/graph_retrievers/entity_linker.py
```diff
@@ -111,4 +110,40 @@ def link(self, query_extracted_entities, retriever=None, topk=None, id_selector=
             parsed_results = []
             for res in results:
                 parsed_results.append(res['document_id'])
-            return parsed_results
\ No newline at end of file
+            return parsed_results
+
+    def link_grouped(self, query_extracted_entities, retriever=None, topk=None):
```
noel-improv: Would `self.entity_linker.link(mentions, topk=1, return_dict=False)` give the same seeds? Each index's `match()` already returns top-k per input, so that would skip the new public method, keep the single batched embedding call, and work with any `Linker`.

byokg-rag/src/graphrag_toolkit/byokg_rag/graph_retrievers/entity_linker.py
```diff
@@ -111,4 +111,40 @@ def link(self, query_extracted_entities, retriever=None, topk=None, id_selector=
...
+            raise ValueError("Error: Either 'retriever' or 'self.retriever' must be provided")
+        if topk is None:
+            topk = self.topk
+
+        # Match each mention on its own through the same matcher link() uses, so
+        # per-mention candidates agree with the union path regardless of index type.
+        grouped = []
+        for mention in query_extracted_entities:
+            hits = retriever.retrieve(queries=[mention], topk=topk)["hits"]
+            grouped.append([hit["document_id"] for hit in hits])
```
noel-improv: potentially blocking

`hits[0]` assumes best-first, but nothing sorts — `NeptuneAnalyticsGraphStoreIndex.match` yields in whatever order the topK query returns. Should this be `max(hits, key=lambda h: h['match_score'])`? With `single_best_match` there's no second candidate to fall back on.

byokg-rag/tests/unit/graph_retrievers/test_entity_linker.py
```diff
@@ -139,6 +156,49 @@ def test_link_multiple_queries(self, mock_retriever):
         assert result[1] == ['e2']
 
 
+class TestEntityLinkerLinkGrouped:
+    """Tests for EntityLinker.link_grouped (per-mention candidate grouping)."""
+
+    def _matcher(self):
```
noel-improv: potentially blocking

Both of these mock out the thing under test, so replacing the per-mention loop with a batched retrieve keeps them green. Could one test run `link_grouped` against a real `FuzzyStringIndex` and assert the flattened result matches `link()`?

byokg-rag/tests/unit/graph_retrievers/test_entity_linker.py
```diff
@@ -120,6 +120,23 @@ def test_link_no_retriever_error(self):
         with pytest.raises(ValueError, match="Either 'retriever' or 'self.retriever' must be provided"):
             linker.link(query_entities)
     
+    def test_link_accepts_id_selector_kwarg(self, mock_retriever):
```
noel-improv: potentially blocking

`link()` accepts `id_selector` and never forwards it to the retriever, so an allowlist is silently ignored while `graph-retrievers.mdx:55` advertises it. This test locks that in. Drop it, or fix the forwarding separately?

byokg-rag/tests/unit/graph_retrievers/test_entity_linker.py
```diff
@@ -120,6 +120,23 @@ def test_link_no_retriever_error(self):
...
+        """id_selector stays in the signature (unused) so existing callers don't break."""
+        linker = EntityLinker(retriever=mock_retriever, topk=3)
+
+        result = linker.link([['Amazon']], id_selector=['entity1'], return_dict=True)
+
+        assert isinstance(result, dict)
+
+    def test_link_positional_argument_order(self, mock_retriever):
```
noel-improv: potentially blocking

`document_id` is a scalar in all three shipped indexes, and a list would make `explored_entities.update()` raise. Should the fixture match the real shape?

byokg-rag/tests/unit/graph_retrievers/test_entity_linker.py
noel-improv, line 163: Routing through `retrieve` (so `match`, not `query`) answers Mykola's point about the fuzzy length filter, but the equivalence is only asserted in the docstring. These tests mock the retriever, and the engine tests mock `link_grouped`, so nothing runs it against a real index — swapping the per-mention loop for a batched retrieve keeps them all green.

Mykola suggested the check: on a `FuzzyStringIndex`, compare `link()` and `link_grouped` per mention. Could that go in as a test, say flattened `link_grouped(ms)` equals `link(ms, return_dict=False)`?

noel-improv (commented):
Nice work, just a few callouts; `link_grouped` duplicates `link()`'s body, and the two have drifted: `id_selector` only appears on one, and they raise at different points. Adding a `group_by_mention` flag to `link()` would drop the `hasattr` check in the engine. Declared on the `Linker` ABC with a default that loops over `link([m])`, it would leave third-party linkers unchanged. The guard doesn't hold anyway, since `entity_linker` is public and reassigning it after construction still triggers the mid-query `AttributeError` it's meant to prevent.

Calling `retrieve()` per mention keeps `inputs` at length 1, so a dense index runs N embedding round trips instead of one batched call. `topk` isn't forwarded either, so each search runs at the linker's default width of 3 and keeps one hit. Since `parse_response` doesn't dedupe and LLMs repeat entities, a repeated mention now costs a second round trip instead of collapsing into one batch.

`single_best_match` also prunes `explored_entities`, which feeds the path retriever, so path coverage narrows too. The docstring and `configuration.mdx` mention only the triplet seeds.

Two smaller nits. The global-sort justification in the docstring applies only to `FuzzyStringIndex`. Both dense indexes already return hits grouped per input, per the same wording at `graph-retrievers.mdx:63`. `single_best_match` is also missing from the full-initialization example in `query-engine.mdx`, so that list disagrees with `configuration.mdx:30`.

byokg-rag/src/graphrag_toolkit/byokg_rag/graph_retrievers/entity_linker.py
```diff
@@ -18,13 +18,16 @@ def __init__(self, *args, **kwargs):
...
+    def link(self, queries: List[str], return_dict=True, group_by_mention=False, **kwargs):
         """
         Process to link the given queries to graph (nodes/edges).
 
         Args:
             queries: List of input query texts to perform graph linking on
             return_dict: Whether to return a dictionary of linking results or linked entities only
+            group_by_mention: Return one candidate list per query instead of a flat
+                union, so the caller can tell which candidate came from which query.
+                Takes precedence over return_dict.
```
acarbonetto: Overrides instead of "takes precedence". 
IF that's true, we should consider an enum instead of a true/false list so that the user cannot be confused.

byokg-rag/src/graphrag_toolkit/byokg_rag/graph_retrievers/entity_linker.py
acarbonetto: update ordering: 
```
            If return_dict is True:
               ...
            If group_by_mention is True:
               ...
            else
               ...
```

byokg-rag/src/graphrag_toolkit/byokg_rag/graph_retrievers/entity_linker.py
```diff
@@ -34,23 +37,37 @@ def link(self, queries: List[str], return_dict=True, **kwargs):
...
+            # Empty-match placeholder. The list-valued fields predate the scalar
+            # contract documented above and the List[Dict] wrapper diverges from
+            # EntityLinker's single Dict; both left as-is to avoid changing the
+            # dict path's shape in a seed-pruning change.
             return [{'hits': [{'document_id': [],
                             'document': [],
                             'match_score': []}
                             ]
                     } for _ in queries]
         else:
```
acarbonetto: remove `else`

byokg-rag/src/graphrag_toolkit/byokg_rag/graph_retrievers/entity_linker.py
```diff
@@ -73,23 +90,35 @@ def __init__(self, retriever=None, topk=3, **kwargs):
...
+            If group_by_mention is True:
+                List[List[str]]: One best-first candidate list per mention, in the
+                    same order as query_extracted_entities
             If return_dict is True:
-                List[Dict]: A list of dictionaries containing linking results for each query
+                Dict: The retriever's result, {'hits': [...]} with one hit per match
+                    and scalar fields. Note this is a single dict, not the List[Dict]
+                    the Linker ABC's own default returns — a pre-existing divergence
+                    between the base default and this implementation.
             If return_dict is False:
```
acarbonetto:
```suggestion
            else:
```

acarbonetto (approved):
Please address documentation comments.  OTherwise, looks good.

noel-improv (approved):
Thanks for addressing

PR #541, opened 2026-09-12
[FEATURE] Keep each graph at the source id width it was written with

noel-improv:
## Description

A source id keeps the first 8 characters of the document's text digest, so a corpus loaded without metadata discriminates on 32 bits and two different documents are given one id after about 30,000 of them. They then share an S3 prefix and merge into one `__Source__` node, with nothing logged or raised. This changes the default to the whole digest, and gives each graph a width it keeps, so an existing graph is not handed a second set of ids for documents it already holds.

A graph records its source id width the first time it is written to, and every run reads that record. A graph written before the record existed has its width read from a sample of stored source ids and then recorded. A document arriving at any other width stops the build with an error naming the document and both widths.

The first two commits are the collision reproducer from #536, which this builds on: the fix removes the `xfail` marker those tests carry.

## Changes

- `IdGenerator` builds source ids from named constants and gains `width_of_source_id`, so the id format is written and read in one place. An id the generator did not write has no width and is left alone.
- New `indexing/source_id_width.py`: resolves the width a run must use, reads and records the width on a graph, and guards a build against documents at another width.
- `LexicalGraphIndex` resolves the width before extraction and runs the guard before the build pipeline.
- `ExtractionPipeline` takes the resolved width instead of reading the configured one.
- `SOURCE_ID_WIDTH` defaults to `FULL`, and a value set explicitly must match the width of the graph being written to.
- `__SYS_Config__` added to the lexical graph labels, so the record is rewritten per tenant like every other node.
- The `xfail` marker is removed from `tests/unit/indexing/test_source_id_collision.py`.

## Problem

Two documents that share a truncated source id are indistinguishable downstream: the S3 prefix is the source id, the graph MERGEs on it, and every chunk, topic and statement id is derived from it. Measured on the current width, 127 pairs collide at 1M documents and 11,762 at 10M.

Related issue (if any): #533

## Testing

Unit tests cover the parser, the resolver, the graph read and record, the guard, and the wiring into `LexicalGraphIndex`. The full unit suite passes at 2,201.

`tests/integration/indexing/test_source_id_width_live.py` runs the same ten checks against a real graph and skips unless `NEO4J_TEST_URI` or `NEPTUNE_GRAPH_TEST_ID` is set. Both stores were run: 10 passed against Neo4j 5 in a container, and 10 passed against a Neptune Analytics graph, which exercises `~id` rather than a property. Each run uses its own tenant and deletes what it wrote.

Multi-tenant behaviour was checked by hand on Neptune: two tenants hold separate records at different widths, and the graph ends up with `__SYS_Config__` and `__SYS_Config__<tenant>__` as separate labels.

## Checklist

The default width changes, which is a behaviour change, documented here and in the configuration docs. An existing graph keeps its own width, so ingesting into one is unaffected. Two cases do change:

- An extract run that cannot read the target graph, which means a dummy graph store or `ExtractionPipeline` used directly, takes the new default. Building that output into a graph written at the old width stops with the mismatch error. Set `SOURCE_ID_WIDTH=LEGACY` to extract for such a graph.
- Deleting every source leaves the width record in place, so starting again at a different width means deleting that record too.

## Still open

- Code calling `ExtractionPipeline` or `BuildPipeline` directly does not go through the guard, which is wired into `LexicalGraphIndex`.
- The width of a graph with no record is read from the first 100 source ids the store returns, so a graph whose widths already diverge past that window reads as whichever width the sample holds. Every graph written from here on records its width, and the record is read in full.

---

### DIFF
--- a/lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/id_generator.py
+++ b/lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/id_generator.py
@@ -4,7 +4,7 @@
 from typing import Optional
 
 from graphrag_toolkit.lexical_graph import TenantId, GraphRAGConfig
-from graphrag_toolkit.lexical_graph.config import DEFAULT_SOURCE_ID_HASH_LENGTH, MAX_SOURCE_ID_HASH_LENGTH
+from graphrag_toolkit.lexical_graph.config import SourceIdWidth
 from graphrag_toolkit.lexical_graph.indexing.utils.hash_utils import get_hash
 from graphrag_toolkit.lexical_graph.utils.arg_utils import coalesce
 
@@ -29,9 +29,9 @@ class IdGenerator(BaseModel):
     tenant_id:TenantId
     include_classification_in_entity_id:bool
     use_chunk_id_delimiter:bool
-    source_id_hash_length:int
+    source_id_width:SourceIdWidth
 
-    def __init__(self, tenant_id:TenantId=None, include_classification_in_entity_id:bool=None, use_chunk_id_delimiter:bool=False, source_id_hash_length:int=None):
+    def __init__(self, tenant_id:TenantId=None, include_classification_in_entity_id:bool=None, use_chunk_id_delimiter:bool=False, source_id_width=None):
         """
         Initialize the IdGenerator.
 
@@ -41,26 +41,20 @@ class IdGenerator(BaseModel):
             use_chunk_id_delimiter: Whether to use delimiter in chunk ID hashing to prevent
                 boundary collisions. Defaults to False for backward compatibility with existing
                 graphs. Set to True for new graphs to enable collision-resistant hashing.
-            source_id_hash_length: Characters of the text digest used in a source ID.
-                Defaults to the configured value. Eight is what existing graphs were
-                written with; widening it changes every source and chunk ID.
+            source_id_width: Characters of the text digest used in a source ID.
+                Defaults to the configured width. LEGACY is what existing graphs were
+                written with; FULL changes every source and chunk ID.
 
         Raises:
-            ValueError: If the resolved width falls outside the digest.
+            ValueError: If the width is not a SourceIdWidth.
         """
-        source_id_hash_length = coalesce(source_id_hash_length, GraphRAGConfig.source_id_hash_length)
-
-        if not 1 <= source_id_hash_length <= MAX_SOURCE_ID_HASH_LENGTH:
-            raise ValueError(
-                f'source_id_hash_length must be between 1 and {MAX_SOURCE_ID_HASH_LENGTH} '
-                f'[source_id_hash_length: {source_id_hash_length}]'
-            )
+        source_id_width = coalesce(SourceIdWidth.parse(source_id_width), GraphRAGConfig.source_id_width)
 
         super().__init__(
             tenant_id=tenant_id or TenantId(),
             include_classification_in_entity_id=coalesce(include_classification_in_entity_id, GraphRAGConfig.include_classification_in_entity_id),
             use_chunk_id_delimiter=use_chunk_id_delimiter,
-            source_id_hash_length=source_id_hash_length
+            source_id_width=source_id_width
         )
 
     def _get_hash(self, s):
@@ -98,7 +92,7 @@ class IdGenerator(BaseModel):
             hashed substrings derived from the input text and metadata.
 
         """
-        return f"aws::{self._get_hash(text)[:self.source_id_hash_length]}:{self._get_hash(metadata_str)[:4]}"
+        return f"aws::{self._get_hash(text)[:self.source_id_width]}:{self._get_hash(metadata_str)[:4]}"
 
     # Delimiter used to separate text and metadata in chunk ID hashing.
     # Using null byte as it cannot appear in valid UTF-8 text strings.
--- a/lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/id_generator.py
+++ b/lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/id_generator.py
@@ -10,6 +10,11 @@ from graphrag_toolkit.lexical_graph.utils.arg_utils import coalesce
 
 from llama_index.core.bridge.pydantic import BaseModel
 
+SOURCE_ID_PREFIX = 'aws'
+METADATA_DIGEST_LENGTH = 4
+
+_HEX_DIGITS = frozenset('0123456789abcdef')
+
 class IdGenerator(BaseModel):
     """
     A class responsible for generating unique and tenant-specific identifiers.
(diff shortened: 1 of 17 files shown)


lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/source_id_width.py
```diff
@@ -0,0 +1,224 @@
...
+            if not item.nodes:
+                return None
+            node = item.nodes[0]
+        else:
+            node = item
+
+        source = node.relationships.get(NodeRelationship.SOURCE)
+        return source.node_id if source else None
+
+    def _check(self, inputs:Iterable[SourceType]):
```
acarbonetto, line 206: `SourceIdWidthGuard._check` records the graph's width while processing the first known-width document, before validating the rest of the batch. 

- build([doc_at_FULL, doc_at_LEGACY]) on an empty graph: the first doc sets graph_width=FULL and writes __SYS_Config__, then the second doc raises SourceIdWidthMismatchError.
- No sources are written, but the graph now permanently records FULL. A later build at LEGACY fails until an operator manually deletes the record.
- test_nothing_is_recorded_when_the_documents_are_refused only covers the configured-contradiction case (refused on the first doc), not this leak.

The consequence. The build failed, so no sources/chunks were written — but the width record was already persisted in step 1. The graph is now permanently stamped FULL despite having no data. A later build([doc_LEGACY]) will keep failing against that phantom FULL record until an operator manually deletes it. The write is a side effect that outlives the transaction it belonged to. 

The fix: te the entire batch first (all documents agree on a width, and it doesn't contradict configured), and only call record_graph_source_id_width once, after the loop confirms the whole batch is acceptable — not mid-loop on the first document.

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/source_id_width.py
```diff
@@ -0,0 +1,224 @@
...
+    the generator wrote.
+
+    The sample is the first SOURCE_ID_SAMPLE_SIZE source ids the store returns, so
+    a collection whose widths already diverge beyond that window reads as whichever
+    width the sample holds. The guard records the width it resolves, so a
+    collection is sampled once and read from its record on every run after that.
+    """
+    results = graph_store.execute_query(
+        f'MATCH (s:`__Source__`) RETURN {graph_store.node_id("s.sourceId")} AS sourceId '
+        f'LIMIT {SOURCE_ID_SAMPLE_SIZE}'
```
acarbonetto: consider adding an ORDER BY clause to ensure consistency between runs

lexical-graph/src/graphrag_toolkit/lexical_graph/lexical_graph_index.py
```diff
@@ -423,6 +424,21 @@ def _configure_extraction_pipeline(self, config: IndexingConfig):
 
         return (pre_processors, components)
 
+    def _source_id_width(self):
```
acarbonetto: Up to four separate width-resolution queries per extract_and_build: _source_id_width() does recorded+sampled (2), and the guard's _check independently re-runs both (2 more). On Neptune these are network round-trips on every ingest. Could resolve once and thread through.

We could potentially cache these calls to increase efficiency or pass down.

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/source_id_width.py
noel-improv: Fixed in 18f6f10a. The guard now records the width only after every input has passed, so a batch refused partway leaves no record.

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/source_id_width.py
noel-improv: Added: ORDER BY sourceId LIMIT 100, with a test. The sort only runs when a collection has no width record yet, which happens once per collection since the resolved width gets recorded.

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/source_id_width.py
```diff
@@ -0,0 +1,243 @@
...
+def recorded_source_id_width(graph_store:GraphStore) -> Optional[SourceIdWidth]:
+    """
+    The width recorded on the collection, or None when it holds no record.
+
+    Every record is read rather than the first one. Nothing constrains a store to
+    a single __SYS_Config__ node, so two first runs at different widths can each
+    create one, and two records that disagree are an error rather than whichever
+    the store happens to return first.
+    """
+    results = graph_store.execute_query(
```
oussamahansal, line 73: `recorded_source_id_width` matches every `__SYS_Config__` with no tenant filter and errors if the widths disagree, but `record_graph_source_id_width` (line 154) writes the node under a tenant-scoped `sysConfigId` via `format_id`. So width is written per-tenant but read collection-globally: two tenants whose first runs resolve different widths (e.g. one with `SOURCE_ID_WIDTH` set, racing before either records) each create their own record, and every later run then hits "records a source id width more than once, and the records disagree" and is blocked until an operator deletes one. Decide whether width is global (then the write key needn't be tenant-scoped) or per-tenant (then the read must filter to this tenant) — right now it's half of each.

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/id_generator.py
```diff
@@ -98,7 +97,33 @@ def create_source_id(self, text:str, metadata_str:str):
...
+    def width_of_source_id(source_id:str) -> Optional[SourceIdWidth]:
+        """
+        The width a source id was written at. A source id is ``aws::<text>:<metadata>``,
+        or ``aws:<tenant>:<text>:<metadata>`` once rewritten for a tenant, so the text
+        digest is the third field in both. Returns None for an id this generator did
+        not write: IdRewriter keeps any id starting ``aws:`` as given, and a build
+        accepts nodes whose source is any id at all.
+
+        Both digests are checked for hex, not just length. A caller-supplied id of
+        the same shape, ``aws:docs:deadbeef:2024``, otherwise reads as a width and
```
oussamahansal: The docstring offers `aws:docs:deadbeef:2024` as the caller-supplied id the hex check guards against, but `deadbeef` and `2024` are both valid hex, so that exact id still returns `LEGACY` — the check only rejects fields with non-hex characters. Use an example that actually has a non-hex field, or drop the "checked for hex, not just length" claim for this case.

noel-improv: Good catch, that id is all hex and reads as LEGACY. Switched the example to `aws:docs:handbook:2024` and said plainly that an id whose fields happen to be hex at a known width can't be told apart.

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/source_id_width.py
noel-improv: The read is per tenant too, through the label rather than the query. I checked it on Neptune: two tenants held separate records at different widths. You're right that nothing shows it, though. Added a test that runs the read through the wrapper for two tenants and a docstring line saying the label does the scoping.

oussamahansal (approved):
lgtm

acarbonetto (approved):

Issue #493, opened 2026-08-20
[Fix] Multi-threaded pipe bottleneck


noel-improv:
<html><head></head><body><h3>Package</h3>
<p>lexical-graph</p>
<h3>Problem statement</h3>
<p><code>extract()</code> fans documents out to N workers, then pulls everything back through a single generator chain in the parent (<code>lexical_graph_index.py:498</code> and <code>:503</code>):</p>
<pre><code class="language-python">nodes | extraction_pipeline | Pipe(handler.accept) | build_pipeline | sink
</code></pre>
<p><code>Pipe</code> is plain generator composition with no threads, so extraction scales with workers but every document afterwards passes through one stage, one at a time. Eight times the workers returns 1.58x, converging near <strong>33 docs/min</strong>.</p>
<p>At 1M documents that works out to roughly 21 days no matter how large the fleet, against a Bedrock quota floor of about 1.5 days.</p>
<h3>Evidence</h3>
<p>Two runs sit behind this. The 33 docs/min ceiling comes from the worker sweep: 2,000 documents, 2/4/8/16 workers at 16 threads each, on-demand, <code>ml.m5.12xlarge</code>. The rows below come from the single-worker runs: 500 documents, one worker, on-demand, <code>ml.m5.4xlarge</code>. Each run splits into a model phase and a drain phase, and the drain costs about 1.4 s per document.</p>

Variable changed | Drain | docs/min
-- | -- | --
16 / 32 / 64 threads | 13.10 / 11.43 / 11.72 min | 38 / 44 / 43
local disk instead of S3 | 14.1 min | 35.5
Nova 2 Lite instead of Sonnet 4.6 | 12.7 min | 39.4

<p>Removing S3 made it slower, which rules out upload cost. And the drain hardly moves between two models whose extraction phases differ by 2.8x.</p>
<p>Four candidates sit in that stage, none of them timed separately: <code>extraction_monitor</code> (<code>:496</code>, present only when a progress monitor is passed), <code>Pipe(handler.accept)</code>, <code>build_pipeline</code> (created with <code>num_workers=1</code> hardcoded at <code>:490</code> and running only a <code>NullBuilder</code>), and <code>sink</code>.</p>
<h3>Proposed solution</h3>
[...]

PR #542, opened 2026-09-12
[FIX] Run a single-worker pipeline in-process instead of spawning a pool per batch

noel-improv:
## Description

`run_pipeline` builds a fresh spawn-context `ProcessPoolExecutor` on every call, even when `num_workers` is 1. The build stage behind `LexicalGraphIndex.extract()` calls it once per batch of four documents to run a `NullBuilder`, so a 500-document extraction starts 125 interpreters, each importing the package, to pass nodes through unchanged. That is the single-threaded drain measured in #493: about 1.4 s per document that did not move with thread count, worker count, model or storage backend. This runs a single-worker pipeline in the calling process instead. Two or more workers still use the spawn pool, so the fork deadlock fix from #419 stays in place.

## Changes

- `pipeline_utils.run_pipeline`: when `num_workers == 1`, apply the transform to each batch in-process and yield, with no executor.
- Tests: the single-worker test now asserts no pool is created and the transform receives the batch; new tests cover cache and kwargs being passed exactly as the pooled path passes them, a real transformation end to end, and the multi-worker path still using a spawn pool.

## Problem

Every drain measurement in #493 fits one process start per build batch. Measured here, one `run_pipeline` call with nothing to do costs 2.2 to 2.3 s regardless of node count or worker count, and one pool round trip is 0.005 s under fork against 2.27 s under spawn. #419 switched the pool to spawn on 2026-07-23; every drain measurement we have is from after that. On the instance, 11.43 min of drain over 125 batches is 5.5 s per batch, and the other four runs land between 5.6 and 6.8 s per batch.

Per-stage timing of the exact `extract()` pipe, 120 documents, chunker in place of the LLM, on a laptop:

| Stage | Before | After |
|---|---|---|
| extraction pipeline | 2.74 s | 2.69 s |
| handler.accept | 0.10 s | 0.07 s |
| build pipeline | 67.66 s | 0.00 s |
| whole run | 70.49 s | 2.76 s |

Handler output is byte-identical between the two runs (sha256 over every written file).

Related issue (if any): #493

## Testing

The three single-worker tests fail on `main` and pass with the change. Full unit suite: 2,138 passed. The per-stage harness above was run on the change itself, not only with the patch simulated.

Re-run on the instance (2026-09-12, stack `drain-1789186485`): WikiHow, all 5,000 documents, 1 worker, 32 threads, on-demand `us.amazon.nova-2-lite-v1:0`, `ml.m5.4xlarge` + `db.r8g.2xlarge`, S3 doc store, this branch packaged by `build-tests.sh`. Extraction phase 04:25:21 to 04:56:10 (30.8 min). Everything after it, the S3 upload of 5,000 documents and 1,250 build batches, ran from 04:56:10 to 04:56:33: 23 seconds, about 0.005 s per document, against 1.4 s per document before. On the previous code the same 1,250 batches would have been roughly 115 minutes of spawns. Whole run 31.7 min for 5,000 documents on one worker, 158 docs/min, against the 33 docs/min ceiling #493 measured across multiple workers. The 5,000 was unintended: `BENCHMARK_EXTRACT_DOC_LIMIT` is read by the `batch_extract.py` tests, not by `benchmark_extract.py`.

## Checklist

Behaviour differences on the single-worker path: exceptions from a transformation now raise where the pipeline runs rather than after the pool shuts down, and with `in_place=True` the transform mutates the caller's node objects instead of pickled copies. Neither caller reads the input nodes after `run_pipeline` returns.

## Not in this PR

`build()` and `extract_and_build()` with real builders still spawn `build_num_workers` processes per batch of four documents. That wants a pool reused across batches and is a separate change.

---

### DIFF
--- a/lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/utils/pipeline_utils.py
+++ b/lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/utils/pipeline_utils.py
@@ -53,6 +53,16 @@ def run_pipeline(
         **kwargs
     )
 
+    # One worker has nothing to run in parallel, and a spawned worker costs a
+    # full interpreter start and package import on every call. The build stage
+    # behind extract() calls this once per batch of four documents, so on a
+    # 500-document run that was 125 spawns and about 1.4 s per document.
+    if num_workers == 1:
+        for node_batch in node_batches:
+            for processed_node in transform(node_batch):
+                yield processed_node
+        return
+
     # Use "spawn": a forked worker can inherit a held lock (e.g. a logging
     # thread's) and deadlock. Spawn starts workers from a clean interpreter,
     # which also drops the GraphRAGConfig singleton's programmatically-set
--- a/lexical-graph/tests/unit/indexing/utils/test_pipeline_utils.py
+++ b/lexical-graph/tests/unit/indexing/utils/test_pipeline_utils.py
@@ -161,31 +161,89 @@ class TestRunPipeline:
                 assert results[3].text == "Node 4"
     
     def test_run_pipeline_with_single_worker(self):
-        """Verify run_pipeline works with single worker."""
+        """One worker runs in the calling process: no pool, so no spawn."""
         mock_pipeline = Mock(spec=IngestionPipeline)
         mock_pipeline.transformations = []
         mock_pipeline.cache = None
         mock_pipeline.disable_cache = True
-        
+
         batch = [TextNode(text="Node 1", id_="1")]
         node_batches = [batch]
-        
+
         with patch('graphrag_toolkit.lexical_graph.indexing.utils.pipeline_utils.run_transformations') as mock_transform:
             mock_transform.return_value = batch
-            
+
+            with patch('graphrag_toolkit.lexical_graph.indexing.utils.pipeline_utils.ProcessPoolExecutor') as mock_executor:
+                results = list(run_pipeline(mock_pipeline, node_batches, num_workers=1))
+
+                assert len(results) == 1
+                mock_executor.assert_not_called()
+                mock_transform.assert_called_once()
+                assert mock_transform.call_args.args[0] == batch
+
+    def test_single_worker_passes_cache_and_kwargs_like_the_pool(self):
+        """The in-process path must hand run_transformations the same arguments
+        the pooled transform gets, or one worker would behave differently from two."""
+        mock_cache = Mock()
+        mock_pipeline = Mock(spec=IngestionPipeline)
+        mock_pipeline.transformations = ['t']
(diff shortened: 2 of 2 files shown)


lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/utils/pipeline_utils.py
```diff
@@ -53,6 +53,16 @@ def run_pipeline(
...
     )
 
+    # One worker has nothing to run in parallel, and a spawned worker costs a
+    # full interpreter start and package import on every call. The build stage
+    # behind extract() calls this once per batch of four documents, so on a
+    # 500-document run that was 125 spawns and about 1.4 s per document.
+    if num_workers == 1:
+        for node_batch in node_batches:
+            for processed_node in transform(node_batch):
+                yield processed_node
```
acarbonetto: The extraction path computes `num_workers = min(self.num_workers, len(round_buckets))`, so a round with a single bucket now hits the in-process path even when the user configured multiple workers. The LLM extractors use a module-level singleton thread pool (`utils/llm_concurrency.py`) that was only ever reclaimed by spawned-worker process exit. Its `shutdown()` has zero callers in the codebase, and its own docstring says the in-process caller must own that shutdown — but the new path doesn't call it. The pool's threads (each holding a bedrock-runtime connection) persist for the life of the parent process and accumulate across runs.

Please add a test case to ensure shutdown() is called.

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/utils/pipeline_utils.py
```diff
@@ -53,6 +54,19 @@ def run_pipeline(
...
     )
 
+    # One worker has nothing to run in parallel, and a spawned worker costs a
+    # full interpreter start and package import on every call. The build stage
+    # behind extract() calls this once per batch of four documents, so on a
+    # 500-document run that was 125 spawns and about 1.4 s per document.
+    if num_workers == 1:
+        try:
+            for node_batch in node_batches:
+                for processed_node in transform(node_batch):
```
oussamahansal, line 64: This path runs `transform` in the caller's process with the default` in_place=True`, so it mutates the input nodes; the spawn path pickles them to a child and leaves the caller's originals untouched. Moving `num_workers == 1` from spawn to in-process thus changes whether inputs are mutated. No current caller trips it — `extraction_pipeline` and `build_pipeline` build fresh `node_batches` per batch and only consume the yielded output — but it's a real divergence between the two paths if a future caller reuses its inputs. Worth a comment, or pass `in_place=False` here for parity.

oussamahansal (approved):

acarbonetto (approved):

PR #543, opened 2026-09-13
fix(s3-based-docs): stop S3DocUploader from losing failed uploads or hanging

### LINKED ISSUE #519: [BUG] S3BasedDocs staging hangs, or reports a failed upload as written

mittalpk:
Fixes #519.

`S3DocUploader._upload_doc` caught its own S3 exception, logged it, and returned `None`. The callback then queued that `None` as a successful result, so a document that failed to write was counted and reported as staged — a retry has no way to know it needs to redo it.

Separately, `_doc_publisher` only put its `target_count` on the queue after its submission loop completed without raising. Any exception partway through that loop (or the producer thread dying outright) meant `target_count` never arrived, so `_upload_batch`'s polling loop waited forever.

Changes, following the invariant from the issue ("every submitted document puts exactly one item on the queue"):
- `_upload_doc` re-raises instead of swallowing.
- `_get_callback_fn` puts a failure sentinel on the queue when the future raised, instead of putting nothing.
- `_submit_proxy` releases its semaphore in the failure path too, and reports whether it actually registered a callback, so `_doc_publisher` only counts documents that are guaranteed to produce a queue item.
- `_doc_publisher` reports `target_count` in a `finally`, so a mid-loop exception still lets `_upload_batch` finish rather than hang.
- `_upload_batch` also detects a dead producer thread directly (checks `is_alive()` on a `queue.Empty` timeout) and raises, as a backstop for the case where the method can't report anything at all.
- Failed uploads are excluded from `S3BasedDocs.accept()`'s yielded/staged count for free, since they're never yielded by `_upload_batch` in the first place.

Added 3 regression tests covering: a failed upload isn't yielded as written, a real mid-loop exception still lets the batch complete instead of hanging, and a fully dead producer thread raises instead of polling forever (this one reproduces the exact `patch.object(S3DocUploader, '_doc_publisher', side_effect=...)` case from the issue). Confirmed fail-before (hangs/times out) against unmodified `main` via `git stash`, pass-after with the fix. Full `tests/unit/indexing/load/` suite passes (254 tests, no regressions).


oussamahansal:
@mittalpk Thanks for your contribution. Clean PR I have just some few comments.

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/load/s3_based_docs.py
```diff
@@ -212,42 +217,51 @@ def _task_complete_callback(future):
...
+            return True
         except Exception as e:
             logger.exception(f'Error in submit proxy: {str(e)}')
-    
+            self._semaphore.release()
+            return False
+
     def _doc_publisher(self, queue:queue.Queue, source_documents:List[SourceDocument]=[]):
 
         s3_client = GraphRAGConfig.s3
```
oussamahansal: `s3_client = GraphRAGConfig.s3` sits above the `try`, so if the S3 client can't be built, `_doc_publisher` dies before the `finally` that enqueues `target_count`. The consumer then falls through to the 60s-timeout dead-thread guard instead of finishing immediately with `target_count=0.` Move the assignment inside the `try` so the `finally` reports on this failure too.

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/load/s3_based_docs.py
```diff
@@ -212,42 +217,51 @@ def _task_complete_callback(future):
...
+                        count += 1
 
         except Exception as e:
             logger.exception(f'Error in doc publisher: {str(e)}')
+        finally:
+            # _upload_batch waits for exactly this many items before it stops
+            # polling the queue, so target_count must reach it even when the
+            # loop above raised partway through - otherwise a failure here
+            # hangs the consumer indefinitely instead of surfacing it.
+            self._queue.put(count)
```
oussamahansal: The `finally` reports via `self._queue.put(count)` while items flow through the `queue` parameter; they're the same object here, but mixing the two is easy to misread. Use the `queue` param for consistency with the rest of the method

oussamahansal (commented):
Few nits

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/load/s3_based_docs.py
mittalpk: Fixed — moved s3_client = GraphRAGConfig.s3 inside the try so this hits the finally too, and added a test that reproduces it directly.

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/load/s3_based_docs.py
mittalpk: Fixed, using the queue parameter now instead of self._queue.

noel-improv:
@mittalpk Thanks for picking this up, we actually hit the same bug and I have #522 open against it (also fixes #519). Both rewrite the same producer/consumer block, but the one real difference is failure semantics. #522 collects failures and fails the run. This counts a failed document and skips it, run completes, worth a comparative review at least for overall clarity. Please take a look and raise any questions or concerns.

mittalpk:
Thanks for the heads up, and sorry for the duplicate effort. Looked at #522 — it's the more complete fix: the 1s liveness poll catches a dead producer far faster than my 60s one, and you've got broader coverage (multi-failure reporting, dead-producer path). On the fail-fast vs. skip-and-continue question, you wrote the issue, so I'll defer to your own read of "an upload failure raises." Closing this in favor of #522.

PR #560, opened 2026-09-16
[FEATURE] Read a half-written source document as incomplete

noel-improv:
## Description

#532 writes a completion marker once every chunk for a document has been stored, but nothing reads it. A run killed part way through leaves a source document prefix holding some of its chunks and no marker, and the reader merges whatever it finds into one document. The build sees a short document rather than a missing one, and a restart has no way to tell the two apart.

This makes the reader skip a prefix its markers do not account for, so an interrupted document is re-extracted instead of being built from half its chunks.

The hard part is what to do about everything already staged. A strict reader would call all of it incomplete and re-stage the lot, which means paying for extraction again. Backfilling markers for prefixes that look whole does not work either: a truncated document and a short document are the same bytes, which is why the marker exists. So the collection answers the question. Staging writes a record under the collection before any document, and a reader that finds the record holds every prefix under it to its markers. A collection staged before markers existed carries no record and is read as it stands.

A date cutoff would do the same job with two weaknesses: someone has to configure the date, and a collection staged by older code after that date still reads wrong. The record has neither.

## Changes

- `collection_record_key` / `writes_completion_markers`: `_staging.json` under the collection, written before any document. A missing record answers no; a denied read or a throttle raises rather than quietly dropping the check.
- `is_complete`: a prefix is whole when its markers cover exactly the nodes found under it. Chunks no marker covers mean a truncated document; a marker covering objects that are gone means the prefix lost something.
- `S3ChunkDownloader` and `S3DocDownloader` skip a prefix that fails the check, logging the prefix at warning level.
- `S3DocUploader` writes a marker for each object it stores. Markers started on the per-chunk path, so without this a recorded JSONL collection would read back as incomplete for good.
- `_write_completion_marker` moved onto `EncryptedPut`, which both uploaders already share.

## Problem

Nothing distinguishes a document whose chunks all landed from one whose run was killed half way through. Both read back as a document, one of them short, and the build indexes the short one.

Related issue (if any): #532

## Testing

Unit tests cover a prefix with no marker, a marker covering a chunk that is gone, several markers accounting for one prefix between them, a collection with no record read as it stands, the record written before any document, a failed record stopping staging, and the JSONL uploader marking what it wrote. Full unit suite: 2,299 passed.

`tests/integration/indexing/load/test_s3_staging_recovery_live.py` stages a collection against real S3, damages it the way an interrupted run would, and reads it back. Twelve cases across both storage formats: the marker deleted, an object deleted from under a marker, re-staging repairing the damage, an undamaged collection reading whole, and a collection whose record was removed reading everything. All twelve pass against S3 in us-east-1. Objects are written under a unique prefix per run and deleted afterwards.

The completeness check was then forced to always pass, and the four bad-state cases failed while the other eight held.

## Checklist

Two behaviour notes for the release:

- An incomplete prefix is skipped, not raised on. The document is absent from the build and is re-extracted on the next run. Raising instead would stop a build over one interrupted document.
- A collection staged before this change is read exactly as it is today. The check starts applying to collections staged from here on.

## Not in this PR

- Reporting which documents were skipped back to the caller. The warning log names each prefix; a restart that acts on the list needs more than that.
- A failed marker write is still logged rather than raised, as it was in #532. It costs a re-stage, where raising would lose a stream the chunks themselves survived.

---


lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/load/s3_based_docs.py
```diff
@@ -109,6 +197,54 @@ def _put(self, key:str, body:str, content_type:str, s3_client):
...
+        Only an empty collection is recorded, so staging into one written before
+        markers existed leaves its documents reading as they do today rather
+        than turning incomplete. Written before the first document, so any
+        collection holding an object holds its record too. A failure raises:
+        without the record every later read of this collection trusts what it
+        finds.
+        """
+        collection_path = join(key_prefix, collection_id, '')
+
+        existing = s3_client.list_objects_v2(
```
acarbonetto, line 230: ensure that `s3:ListBucket` is included in all IAM roles and add this to the documentation - or else we will get an access denied here.

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/load/s3_based_docs.py
```diff
@@ -766,22 +897,27 @@ def accept(self, source_documents: List[SourceDocument], **kwargs: Any) -> Gener
...
 
             if self.for_jsonl:
-                self._uploader = S3DocUploader(
-                    bucket_name=self.bucket_name, 
+                uploader = S3DocUploader(
+                    bucket_name=self.bucket_name,
                     collection_prefix=collection_prefix,
                     s3_encryption_key_id=self.s3_encryption_key_id,
                     num_threads=self.num_threads,
                     deterministic_document_key=self.deterministic_document_key
```
acarbonetto, line 924: the document key continues to truncate to 5 characters.  This _could_ be a problem in some circumstances.  We should consider removing this truncation: _doc_suffix (line 383–385)

acarbonetto (approved):

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/load/s3_based_docs.py
```diff
@@ -79,10 +85,109 @@ def is_completion_marker(key:str) -> bool:
...
+    """Split a source document prefix's keys into content and markers."""
+    markers = [key for key in keys if is_completion_marker(key)]
+    content = [key for key in keys if not is_completion_marker(key)]
+    return content, markers
+
+
+def chunk_id_from_key(chunk_key:str, source_doc_prefix:str) -> str:
+    """
+    The node id a chunk object is keyed on. Taken by stripping what the writer
+    added, because a node id can hold both a '/' and a '.'.
```
oussamahansal, line 170: The stated reason no longer holds: source_document_prefix now runs every node id through validate_id_segment, which rejects /. Only the '.' half is still load-bearing (a node id ending in .json keys an object ending .json.json).

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/load/s3_based_docs.py
```diff
@@ -254,15 +410,20 @@ def _upload_doc(self, root_path:str, doc:SourceDocument, s3_client):
...
+            nodes = written_nodes(doc)
+
             s = '\n'.join([
                 json.dumps(n.to_dict())
-                for n in written_nodes(doc)
-            ]) 
+                for n in nodes
+            ])
 
             self._put(doc_output_path, s, 'text/plain', s3_client)
```
oussamahansal, line 420: The marker write is guarded on if nodes but the object put isn't, so a document whose nodes all carry INDEX_KEY passes the source_document.nodes guard and writes an empty .jsonl with no marker — I staged one and got p/c/src1/src1-35df1.jsonl with a zero-byte body. The prefix then satisfies is_complete only by vacuous equality (set() == set()) and reads back as a SourceDocument with no nodes. Move the put under the same if nodes check.

oussamahansal (approved):

lexical-graph/src/graphrag_toolkit/lexical_graph/utils/id_validation.py
```diff
@@ -0,0 +1,27 @@
...
+
+import re
+
+# An allowlist (vs blocking specific separators) closes URL-encoding, Unicode,
+# and double-encoding bypasses in one check. IdGenerator ids
+# (aws::<hex>:<hex>:<hex>, optional tenant segment) fit.
+_ID_SEGMENT_PATTERN = re.compile(r'[A-Za-z0-9:._-]+')
+
+
+def validate_id_segment(value:str, name:str) -> None:
```
acarbonetto, line 12: Can you ensure that validation still passes for existing user-controlled ids?  

Please check: spaces (annual report 2024), non-ASCII (rapport-été).

noel-improv: Both pass. IdRewriter replaces any id that doesn't already start with aws:, so 'annual report 2024' reaches validation as aws::5eb63bbb...:298c. Same for 'rapport-été' and Cyrillic. The one shape that survives unchanged is an id already claiming the prefix, like aws:annual report 2024, which gets rejected. Added tests for both directions.

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/load/s3_based_docs.py
noel-improv: Right, only the '.' half is load-bearing now. Reworded: a node id can hold a '.', so an id ending in .json keys an object ending .json.json.

acarbonetto (approved):

PR #561, opened 2026-09-17
[FEATURE] End a source document when its last part lands

noel-improv:
## Description

Follows #560, which skips a source document prefix its markers do not account for. A marker covers one `SourceDocument`, and extraction emits a source as several of those when its chunks span rounds. Each part was marked as it landed, so the markers under the prefix accounted for exactly the chunks present and a source that lost its later parts read back as a whole document.

The part that ends a source now declares every chunk id stored for that source, and the reader compares the objects present against that declaration.

Reproduced against S3 before the change: a source staged as two rounds read back whole at 4 chunks, and with round two removed it read back as a valid 2-chunk document rather than being skipped.

## Changes

- `SourceDocument.final_part` says whether a document holds the last of its source's nodes. Defaults to true, so every producer that emits a source as one document is unchanged.
- `ExtractionPipeline` names the sources a round leaves with nothing left to extract, and marks only the last document emitted for one of those. Chunks are counted on the way in: a checkpointed run extracts fewer nodes than it was handed, so counting output would leave a resumed source open for good.
- Both uploaders accumulate the chunk ids they stored per source. The marker on the part that ends a source carries the whole set, and `is_complete` compares the prefix against that.
- Running out of stream ends any source still open. A round can finish a source without emitting a document for it, so the closing part never arrives; a run that dies instead never reaches this point and its prefix stays incomplete.
- A source whose chunk upload failed is dropped rather than closed, so it gets no marker either way, as in #532.

## Problem

Three things make a bare per-document marker insufficient:

- A source can split **within** one round, not only across rounds. The batch extractor sorts its output by node id and documents are cut on contiguous runs of source id.
- The JSONL upload pool gives no ordering guarantee between parts, so a closing marker can be written before an earlier part's object.
- Chunk and marker write failures are logged rather than raised, so a part can be marked over a source whose earlier part partly failed.

A declared set survives all three. A flag on its own does not.

Related issue (if any): #560

## Testing

Unit tests cover a source split across rounds, a source split within one round by sorted output, a round that drops chunks, each source ending on its own last document, the declared set on both uploaders, a source left open by the end of the stream, and a source whose chunk failed getting no marker. Full unit suite: 2,320 passed.

`tests/integration/indexing/load/test_s3_staging_recovery_live.py` runs against real S3 and now covers a split source losing its opening part, in both storage formats. 18 cases pass in us-east-1.

Each half was reverted in turn: forcing every part to be final failed the two split tests, and dropping the declared set from the marker failed the three declaration tests.

End to end against Bedrock, S3 and Neo4j: 3 documents extracted with 3 markers and the collection record, an intact build giving 3 `__Source__` nodes, and a rebuild after removing one marker giving 2.

## Checklist

One behaviour worth naming for review: a run that completes normally closes every source still open, declaring what was stored. That is what makes a source whose closing part never arrived readable at all, and it means the end of a run is treated as evidence. A run that dies is not, and leaves the prefix incomplete.

## Not in this PR

- `FileBasedDocs` writes no markers and is unchanged. The `final_part` default keeps it behaving as it does today.
- Reporting skipped documents back to the caller. The warning names each prefix; a restart that acts on the list needs more than that.

---


lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/load/s3_based_docs.py
```diff
@@ -309,10 +566,19 @@ def _doc_publisher(self, queue:queue.Queue, source_documents:List[SourceDocument
...
+
+                    source_id = source_document.source_id()
+                    nodes = written_nodes(source_document)
+                    root_path = source_document_prefix(
+                        self.collection_prefix, source_document, nodes
+                    )
+
+                    self._open_source(source_id, root_path, [n.node_id for n in nodes])
+                    source_chunk_ids = (
+                        self._end_source(source_id) if source_document.final_part else None
```
oussamahansal: The guard at line 567 tests source_document.nodes, but _upload_doc writes the marker only if nodes — i.e. written_nodes(doc). A final part whose nodes all carry INDEX_KEY passes the guard, _end_source pops the source here, then the marker write is suppressed, so nothing closes the source and _close_open_sources has nothing left to close. I ran both uploaders over [part1(a,b, final_part=False), part2(index-only, final_part=True)]: S3ChunkUploader writes a final marker, S3DocUploader writes only the non-final part marker, so the prefix reads incomplete on every later run. S3ChunkUploader gets this right by keying its submit loop on written_nodes (line 865); do the same here, or move the _end_source call into _upload_doc alongside the if nodes check.

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/load/s3_based_docs.py
```diff
@@ -79,10 +85,120 @@ def is_completion_marker(key:str) -> bool:
...
+    A truncated document leaves nodes no marker covers, and one whose objects
+    were partly removed leaves a marker covering nodes that are gone.
+
+    A source stored in several parts is accounted for by the part that ends it,
+    which declares every chunk id stored for the source. A marker written before
+    that existed declares only its own chunk ids.
+    """
+    markers = read_markers(marker_keys, bucket_name, s3_client)
+
+    final_markers = [marker for marker in markers if marker.get('final', True)]
```
oussamahansal, line 188: declared is built only from final markers, but it's compared for exact equality against every node id found under the prefix, and _open_sources only knows what this run staged. Reusing an explicit collection_id after a run died mid-source leaves a final: False part marker for, say, {a,b}; the resumed run's checkpoint drops those chunks (the case test_a_round_that_drops_chunks_still_ends_the_source covers), so it declares only {c,d} — is_complete({a,b,c,d}, ...) returns False even though every chunk is present, and the deterministic marker key means a re-run reproduces the same declaration, so the source is dropped from the build permanently. Either fold non-final markers' chunk_ids into declared, or make the comparison declared <= set(node_ids).

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/load/s3_based_docs.py
```diff
@@ -109,6 +225,130 @@ def _put(self, key:str, body:str, content_type:str, s3_client):
...
+    its last chunk, which declares every chunk id stored for the source. A
+    source that lost a chunk is poisoned and never ends, so its prefix reads as
+    incomplete and the document is staged again.
+
+    Writing goes through EncryptedPut, which the host class also inherits.
+    """
+
+    # Supplied by the host class.
+    bucket_name:str
+    _open_sources:Dict[str, Tuple[str, List[str]]]
```
oussamahansal, line 252: _open_sources and _poisoned_sources are annotated here as "Supplied by the host class" but pydantic won't inherit PrivateAttr from a plain mixin, so both hosts redeclare them (lines 482-483, 802-803). A third host that inherits CompletionMarkers and forgets gets an AttributeError on the first _open_source rather than a construction failure. Worth a line in the docstring saying the host must declare them as PrivateAttr.

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/extract/extraction_pipeline.py
```diff
@@ -631,6 +616,29 @@ def get_source_metadata(node):
...
 
+        # Chunks fed in for each source and not yet submitted. Counted on the way
+        # in rather than on the way out, because extraction can emit fewer nodes
+        # than it was given: a checkpointed run drops chunks already extracted.
+        outstanding = defaultdict(int)
+
+        def source_id_of(node):
+            return node.relationships[NodeRelationship.SOURCE].node_id
+
+        def sources_finished_by(round_jobs):
```
oussamahansal, line 627: sources_finished_by reads as a query ("The sources this round leaves with nothing left to extract") but decrements and deletes entries in outstanding, so calling it twice on the same round gives different answers. lets add it in the docstring

oussamahansal (changes requested):

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/load/s3_based_docs.py
noel-improv: Confirmed and fixed, S3ChunkUploader closes the source, S3DocUploader writes only the non-final marker, as you said.

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/load/s3_based_docs.py
noel-improv: confirmed this, although reproducing it turned up a second direction. With declared built from final markers alone, a prefix whose earlier part's objects are gone read as complete; {c,d} present, final marker declaring {c,d}, and the missing {a,b} nowhere in declared. Folding fixes both; both are in `TestARunResumedAfterAnEarlierRunDied`.

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/load/s3_based_docs.py
noel-improv: Added to the CompletionMarkers docstring: the host declares `_open_sources` and `_poisoned_sources` as PrivateAttr, because pydantic does not collect them from a plain mixin, and a host that leaves them out fails on first use rather than at construction.

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/extract/extraction_pipeline.py
noel-improv: Added. The docstring now says it consumes from outstanding as it counts, so it answers for each round once, and a second call on the same round reports nothing.

oussamahansal (approved):

acarbonetto (approved):

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/load/s3_based_docs.py
```diff
@@ -590,6 +708,10 @@ def upload(self, source_documents: List[SourceDocument]):
             source_docs_batch = []
             logger.debug(f'Total uploaded: {total}')
 
+        # Outside the pool: a closing marker supersedes the one its last part
+        # wrote, so it cannot be written while that part is in flight.
+        self._close_open_sources(GraphRAGConfig.s3)
```
oussamahansal, line 713: _close_open_sources treats the end of upload() as the end of the run, but S3BasedDocs caches self._uploader across calls (if not self._uploader at :1063, whose comment says "the next call"), so a second accept() on the same instance closes whatever the first left open with a marker declaring only that call's chunks. Driving accept twice with part1(a,b, final_part=False) then dying writes final=True, source_chunk_ids=['a','b'] and is_complete returns True with c,d never stored. Unreachable today since Pipe(handler.accept) calls accept once per run, but the two comments assume opposite lifetimes, maybe pick one or reset the mixin state per upload().

oussamahansal (approved):

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/load/s3_based_docs.py
noel-improv: Good catch, and the half I could reach was worse than the scenario you described: `_close_open_sources` cleared `_open_sources` but never `_poisoned_sources`, so a source that lost a chunk stayed unmarked for every later stream and the chunks the next one stored for it were never accounted for. Fixed with a test that fails without it. Your second point stands, the close hook's docstring and the cached uploader still assume opposite lifetimes. I'd rather settle that deliberately than fold it into this PR; raising it separately as a follow-up

PR #563, opened 2026-09-18
[FIX] Strip Llama special tokens from batch prompt content

oussamahansal:
## Description

The Meta Llama batch path renders chat messages into Llama's raw instruct template, where turn boundaries are text rather than structure. Message content was interpolated verbatim, so untrusted chunk text could close its turn and open a forged one. This strips Llama special tokens from content before formatting.

## Changes

- Add `_strip_llama_special_tokens`
- `_format_llama_prompt` hoists system messages to the front, so a later turn can't displace the instruction block.

## Problem

Related issue (if any): #

## Testing

## Checklist

---


noel-improv (approved):
Nice work with this, can we confirm this non-blocker quadratic strip loop finding; attacker-reachable input path, measured 1.7s at n=20000 and the loop has no cap. Could it worth capping the passes?

mykola-pereyma (approved):
LGTM. This closes the Llama batch-path turn-forgery injection cleanly.

The `<\|[^<>|\n]*\|>` regex is scoped right — excluding `<`, `>`, `|`, and newline from the body stops a stray `<|` from swallowing text up to a distant `|>` — and the fixed-point loop correctly defeats splice reconstruction (`<|eot_i<|image|>d|>` → empty). Hoisting system messages to the front is good defense-in-depth so a forged later turn can't displace the instruction block. Token coverage is complete for the Llama 3.x/4 `<|...|>` families this template targets. The 5 tests exercise it through the public `get_request_body` and assert both the security invariants (no chunk-contributed headers/eot) and content fidelity.

One non-blocking note: hoisting reorders all system messages to the front, so a caller that intentionally placed a system message mid-conversation would see its order change — fine for the extraction path here.

PR #569, opened 2026-09-21
[FIX]: build S3 Vectors metadata filters as dicts, not JSON strings

oussamahansal:
## Description

S3 Vectors metadata filters were hand-assembled as a JSON string and round-tripped through `json.loads`. Filter keys and values were interpolated with no escaping, so a crafted value could close the application's clause list and open a second `$and` and Python's JSON parser keeps the last duplicate key, replacing the whole clause set. This builds the filter as native Python dicts and hands them to boto3.
## Changes

- `parse_metadata_filters_recursive` builds native dicts end to end
- `formatter_for_type` returns the native Python value instead of a pre-quoted
  JSON fragment

## Problem

Related issue (if any): #

## Testing

## Checklist

---


lexical-graph/src/graphrag_toolkit/lexical_graph/storage/vector/s3_vector_indexes.py
```diff
@@ -59,65 +59,79 @@ def to_s3_operator(operator: FilterOperator) -> tuple[str, Callable[[Any], str]]
...
+        return lambda x: format_datetime(x)
+    elif type_name == 'int':
+        return lambda x: int(x)
+    elif type_name == 'float':
+        return lambda x: float(x)
     else:
         raise ValueError(f'Unsupported type name: {type_name}')
 
 
 def parse_metadata_filters_recursive(metadata_filters:MetadataFilters) -> Dict[str, Any]:
```
noel-improv: **[non-blocking]** `main` still has the string builder. It and `v3.20.x` each sit two commits past their 2026-09-22 merge base, and `origin/main`'s copy of this function is still `json.loads(f'{{"$and": [{",".join(filter_strs)}]}}')`, so the injection is live there.

lexical-graph/src/graphrag_toolkit/lexical_graph/storage/vector/s3_vector_indexes.py
```diff
@@ -59,65 +59,79 @@ def to_s3_operator(operator: FilterOperator) -> tuple[str, Callable[[Any], str]]
...
+                filter_clauses.append(nested_filters)
         else:
             raise ValueError(f'Invalid metadata filter type: {type(metadata_filter)}')
 
-    if not filter_strs and (
+    if not filter_clauses and (
         metadata_filters.condition == FilterCondition.AND
         or metadata_filters.condition == FilterCondition.OR
     ):
         return {}
```
noel-improv: **[non-blocking]** This predates the PR, but it is the same threat model, so worth asking while the function is open. A filter tree whose clauses all reduce to nothing returns `{}` here, and `search_vectors` gates on `if metadata_filters:`, so an empty dict drops the `filter` key and the query runs unscoped:

```python
inner = MetadataFilters(filters=[], condition=FilterCondition.AND)
outer = MetadataFilters(filters=[inner], condition=FilterCondition.AND)
parse_metadata_filters_recursive(outer)   # {}  ->  query_vectors called with no filter
```

I checked that against a stub client and the `filter` key is absent. So a caller that thinks it is scoping a query gets everything back instead of an error. Would it be worth failing closed, or at least logging when a non-empty `MetadataFilters` collapses to `{}`? This can also go in a separate ticket if you would rather keep this PR to the one concern.

lexical-graph/src/graphrag_toolkit/lexical_graph/storage/vector/s3_vector_indexes.py
```diff
@@ -59,65 +59,79 @@ def to_s3_operator(operator: FilterOperator) -> tuple[str, Callable[[Any], str]]
     
     return operator_map[operator]
 
-def formatter_for_type(type_name:str) -> Callable[[Any], str]:
-    
+def formatter_for_type(type_name:str) -> Callable[[Any], Any]:
```
noel-improv: **[nit]** Now that the quoting is gone, this is the same function as `metadata.formatter_for_type` (metadata.py:101) apart from `text` returning `str(x)` rather than `x`. This module already imports `type_name_for_key_value` and `format_datetime` from there, so the local copy could go and `formatter_for_type` join that import. That would take the count from four copies to three, with `graph_utils` and `pg_vector_indexes` keeping theirs for their own reasons.

lexical-graph/src/graphrag_toolkit/lexical_graph/storage/vector/s3_vector_indexes.py
```diff
@@ -59,65 +59,79 @@ def to_s3_operator(operator: FilterOperator) -> tuple[str, Callable[[Any], str]]
...
+    invalid filter type or condition.
+    """
     def to_key(key: str) -> str:
         if key == VALID_FROM:
-            return f'{SOURCE_VERSIONING_PREFIX}valid_from'      
+            return f'{SOURCE_VERSIONING_PREFIX}valid_from'
         elif key == VALID_TO:
             return f'{SOURCE_VERSIONING_PREFIX}valid_to'
         else:
             return f'{SOURCE_METADATA_PREFIX}{key}'
```
noel-improv: **[nit]** The write path checks key length against `MAX_KEY_LENGTH` and raises `ValueError` (line 226), but filter keys are not checked, so an over-long key now reaches boto3 as a literal and comes back as a `ClientError` instead.

noel-improv (approved):
Looks good to me, just a few callouts nothing blocking.

mykola-pereyma (approved):
LGTM. Moving the filter from a hand-assembled JSON string to native dicts is the right fix — it separates query structure from data the same way parameterized queries do, so a value or key full of JSON punctuation and a second `$and` now stays a literal and can't close the clause list or exploit `json.loads`'s last-duplicate-key behavior to drop the app's scoping clause. Keys stay prefixed with `source.metadata.`, IS_EMPTY uses a real `False`, and nested groups nest as dicts. The red-state tests are a highlight — one payload that crashes pre-fix and a balanced one that silently replaced the clause set pre-fix, plus the scope-clause-not-dropped and injected-key-never-a-clause assertions and benign coverage.

Two non-blocking follow-ups (not for this PR): int/float now coerce explicitly (`int(x)`/`float(x)`), so a value like `'5.0'` mis-classified as int would raise where the old identity + `json.loads` path didn't — low risk, worth a glance at `type_name_for_key_value`; and it'd be worth checking the sibling vector-store backends for the same JSON-string filter-assembly pattern.

PR #571, opened 2026-09-22
[FIX] Dedupe beam search results within a single expansion round

### LINKED ISSUE #570: BeamSearch.beam_search returns duplicate node ids when a node is reachable from two beam members in the same round
graphrag-lexical-graph 3.19.1 at HEAD `8f884d8`, Python 3.12, lexical-graph package.

`BeamSearch.beam_search()` (`lexical-graph/src/graphrag_toolkit/lexical_graph/retrieval/retrievers/beam_search_base.py`) can return the same node id twice in one call, when that node is reachable from two different beam members in the same expansion round.

The batch-drain loop reads up to `batch_size` (10) queue entries, filtering on `current_id not in visited`. `visited` itself is only updated in the loop that runs after the whole batch is drained. If a node was enqueued twice in the previous round (two parents both link to it via `get_neighbors_batch`), both entries can pass the drain filter before either is marked visited, land in the same batch, and both get appended to `results` with no check in that second loop.

`TopicBeamSearch.do_graph_search` builds `winning_topic_ids` directly from these results, so one topic can occupy two of the configured `beam_width` slots, returning fewer distinct topics than `beam_width` promises.

Repro (no repo deps beyond `numpy`, only exercises `BeamSearch` directly):

```python
import numpy as np
from graphrag_toolkit.lexical_graph.retrieval.retrievers.beam_search_base import BeamSearch

class FakeBeamSearch(BeamSearch):
    def __init__(self, graph, embeddings, **kwargs):
        super().__init__(**kwargs)
        self.graph, self.embeddings = graph, embeddings
    def get_neighbors(self, node_id):
        return self.graph.get(node_id, [])
    def get_neighbors_batch(self, node_ids):
        return {n: self.graph.get(n, []) for n in node_ids}
    def _get_embeddings(self, ids):
        return {i: self.embeddings[i] for i in ids if i in self.embeddings}
    def _get_top_k(self, query_embedding, embeddings, top_k):
        return self._score_neighbors(query_embedding, None, embeddings, top_k)

# A and B both seed the search and both link to the same neighbor C
graph = {"A": ["C"], "B": ["C"], "C": []}
embeddings = {"A": [1.0, 0.0], "B": [0.9, 0.1], "C": [0.5, 0.5]}

bs = FakeBeamSearch(graph, embeddings, beam_width=10, max_depth=3, scoring_mode='cosine')
results = bs.beam_search(np.array([1.0, 0.0]), ["A", "B"])
print(results)
# [('A', ['A']), ('B', ['B']), ('C', ['A', 'C']), ('C', ['B', 'C'])]
# C appears twice.
```

Reproduces the same way under `scoring_mode='path_weighted'` (the actual default `TopicBeamSearch` uses). Not a crash, just a silent recall/diversity reduction: fewer distinct results than `beam_width` intends.

AmirF194:
`beam_search()` drains up to 10 queue entries per round, filtering on `current_id not in visited`, but `visited` is only updated in a second loop that runs after the whole batch is assembled. If a node was enqueued twice in the previous round (two different parents both linking to it), both entries can pass the drain filter before either is marked visited, land in the same batch, and the second loop appends both to `results` with no check.

Added the missing check at the top of that second loop, same pattern the drain loop already uses one level up.

`TopicBeamSearch.do_graph_search` builds `winning_topic_ids` straight from these results with no dedup, so this was silently handing back fewer distinct topics than `beam_width` promises whenever two beam members shared a neighbor in the same round.

Two new tests reproduce it directly against `BeamSearch` (a seed pair both linking to one shared neighbor), one per scoring mode since `cosine` and `path_weighted` score neighbors on different code paths. Without the fix the shared neighbor comes back twice; with it, once. Full `lexical-graph` suite: 2377 passed, 53 skipped on Python 3.10 and 3.12, coverage 67.78% (repo's own `--cov-fail-under=56`). `bandit -r src/ -ll` clean, license-header check clean.

Fixes #570


oussamahansal (approved):
Thanks for your contribution. LGTM

Issue #269, opened 2026-05-14
[BUG] Windows: cannot install via git+https or zip due to NTFS-illegal filenames (aws:: paths in example data)


russor45:
### Summary

Installing any subdirectory package from this repo on Windows fails because the repository contains example data files with `::` in their paths. Windows NTFS forbids `:` in filenames and directory names, causing both `git clone` and zip extraction to fail with a fatal error.

### Affected packages
- `graphrag-lexical-graph` (subdirectory: `lexical-graph`)
- `graphrag-toolkit-lexical-graph-falkordb` (subdirectory: `lexical-graph-contrib/falkordb`)

### Steps to reproduce

On Windows, run either:

```
uv pip install "graphrag-lexical-graph @ git+https://github.com/awslabs/graphrag-toolkit.git@b9b1c1b4ccddb7b8bb73f71e7a7b250e0350ccb3#subdirectory=lexical-graph"
```

or the zip form:

```
uv pip install "graphrag-lexical-graph @ https://github.com/awslabs/graphrag-toolkit/archive/b9b1c1b4ccddb7b8bb73f71e7a7b250e0350ccb3.zip#subdirectory=lexical-graph"
```

### Git clone error
```
error: invalid path 'examples/lexical-graph/workshop/source-data/amazon/aws::00a69cae:36fd/aws::00a69cae:36fd:058d6f34.json'
fatal: Could not reset index file to revision 'b9b1c1b4ccddb7b8bb73f71e7a7b250e0350ccb3'.
```

### Zip extraction error
```
Failed to extract archive
I/O operation failed during extraction
failed to create directory `.../graphrag-toolkit-b9b1c1b4.../examples/lexical-graph/workshop/source-data/amazon/aws::00a69cae:36fd/`:
The filename, directory name, or volume label syntax is incorrect. (os error 123)
```

### Root cause

The directory `examples/lexical-graph/workshop/source-data/amazon/` contains files and folders whose names use `::` as part of AWS resource identifiers (e.g. `aws::00a69cae:36fd`). This is valid on Linux/macOS but illegal on Windows NTFS, where `:` is a reserved character used for NTFS alternate data streams.

### Impact

- `uv sync`, `pip install`, and any tooling that installs these packages on Windows is completely broken, even when targeting subdirectories that contain no such filenames themselves. The entire repo archive must be extracted before the subdirectory can be used.
- Affects all Windows developers and any Windows-based CI runners.

### Fix options (easiest first)

[...]

PR #574, opened 2026-09-22
[FIX]: apply the id path-traversal guard to all three file sinks

oussamahansal:
## Description

`FileSystemTap._validate_id` guarded only one of the three sinks that join an id onto a prepared output directory. `IdRewriter` returns any id already starting `aws:` unchanged, so an id such as `aws:../../etc/x` reached `FileBasedDocs.accept` and `CheckpointWriter.accept` untouched.
## Changes

- Move the validator to `indexing/utils/path_utils.validate_id`
- validate `source_id` before `os.makedirs` and `node_id` before `open`
- validate `node_id` before the join and `touch`

## Problem

Related issue (if any): #

## Testing

## Checklist

---


mykola-pereyma (approved):
LGTM. This closes the id path-traversal across the sinks cleanly. The allowlist validator (`[A-Za-z0-9:._-]+` fullmatch + explicit `.`/`..` reject) is the right call — separators can't appear, so relative, `aws:`-prefixed, and absolute ids are all rejected, and URL-encoded/Unicode separators fail the ASCII allowlist too. Placement is correct at every sink: it validates the exact value that gets joined — post-rewrite in `CheckpointFilter`, and in `FileBasedDocs` both the id and the sanitizer's output (nice catch, since the sanitizer defaults to a no-op and a custom one can introduce a separator). The red-state tests that snapshot the tree and assert nothing was created outside it are a great touch.

A couple of non-blocking nits:

1. The PR description is stale — it says "Move the validator to `indexing/utils/path_utils.validate_id`", but the module is `utils/id_validation.py` and the function is `validate_id_segment`, and it already existed on base (this PR adds the call sites rather than moving anything). Worth fixing the writeup so the history reads correctly.

2. Colon is allowlisted because the real ids need it (`aws::<hex>:<hex>`), which is fine on POSIX. On Windows a single-letter-drive + colon (e.g. `c:foo`) joined via `os.path.join` can resolve drive-relative; `aws:`-style prefixes are safe and `FileBasedDocs` also validates the sanitized output (and `windows_safe_filename` rewrites colons), so it's low risk — just worth a comment near the pattern.

3. Completeness rests on the sink enumeration being exhaustive. Might be worth a short note (or a follow-up) that any future id→path sink must call `validate_id_segment`, so this doesn't regress.

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/load/file_based_docs.py
```diff
@@ -238,17 +240,28 @@ def accept(self, source_documents: List[SourceDocument], **kwargs: Any) -> Gener
...
+
+        Raises:
+            ValueError: If a source id or node id would write outside the collection directory.
         """
         sanitize = self.filename_sanitizer if self.filename_sanitizer else lambda x: x
 
+        def safe_name(value:str, name:str) -> str:
+            """Validate the id, then the sanitizer's output, which is what gets joined.
+            A custom sanitizer can introduce a separator as easily as remove one."""
+            validate_id_segment(value, name)
```
noel-improv: Validating the raw id before the sanitizer runs breaks the case the sanitizer was added for (#269): a BYO id like `aws:annual report 2024` now raises before it can be fixed. Would validating only the sanitized output be enough? It's the string that gets joined, so it still closes the traversal.

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/extract/batch_extractor_base.py
```diff
@@ -209,6 +210,7 @@ def _post_process_node(self,
         return node
     
     def _save_node_in_temp_dir(self, node:TextNode, temp_dir:str):
+        validate_id_segment(node.node_id, 'node_id')
```
noel-improv: Non-blocking: the read side at line 274 joins `node_id` from the batch output's `recordId` onto `temp_dir` without the guard. Lower stakes since it's a read, but should it get the same call?

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/build/checkpoint.py
```diff
@@ -62,8 +63,14 @@ def checkpoint_does_not_exist(self, node_id):
             bool: Returns False if the checkpoint exists, indicating the node should
                 be ignored. Returns True if the checkpoint does not exist,
                 indicating the node should be included.
+
+        Raises:
+            ValueError: If the id would resolve outside the checkpoint directory. The
+                writer rejects the same id, so probing it here would only decide the
```
noel-improv: Non-blocking: with a non-default tenant the filter validates the rewritten id and the writer the raw one, so "the writer rejects the same id" isn't quite true. Maybe soften it?

lexical-graph/tests/unit/indexing/test_id_path_validation.py
```diff
@@ -0,0 +1,299 @@
...
+
+
+def _nested_dir(root, leaf):
+    """A sink directory deep enough that a climbing id leaves it but stays in `root`."""
+    path = os.path.join(root, 'a', 'b', 'c', 'd', leaf)
+    os.makedirs(path)
+    return path
+
+
+class TestFileSystemTap:
```
noel-improv: Non-blocking: `test_file_system_tap.py::TestFileSystemTapIdValidation` already pins this. Worth dropping the duplicate so the tap's contract lives in one place?

lexical-graph/tests/unit/indexing/test_id_path_validation.py
```diff
@@ -0,0 +1,299 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Red-state tests for path traversal in every id-to-path sink.
```
noel-improv: Non-blocking: could the module docstring drop the "Red-state" / before-and-after narration? The per-class docstrings already state what each sink must hold.

noel-improv (changes requested):
Guard placement looks right, I confirmed the new tests go red with the three source changes reverted. One regression in FileBasedDocs to fix; the rest are optional.

PR #577, opened 2026-09-23
[FEATURE] Record how an extraction run divided its input, and follow it on restart

noel-improv:
An extraction run that dies partway through has to start over, because nothing records how it divided its input. This adds that record: a run writes a plan before it submits any work, and a restart of the same run id obeys the plan instead of dividing the collection again.

## What a plan holds

The ordered document ids, the sizing settings, the worker count, and the configuration the run started under. It stores the inputs to the division, never the resulting assignments, because document and chunk ids are content hashes and the dividing steps are deterministic given the same inputs.

The plan lands at `<key_prefix>/<collection_id>/_runs/<run_id>/plan.json`, beside the collection it divides. Nothing new is provisioned. The directory is reserved the way `_markers` already is, so a delimited listing does not return it as a source document.

## Why the worker count is pinned

Extraction clamps the worker count to the host's core count, so the same collection divides differently on a smaller machine. A restart adopts the recorded count and runs the pieces across however many processes the host has. That is the one setting a restart may find different without failing, since the operator did not choose it.

A changed batch size or a changed set of documents fails with an error naming what differs. Everything else is recorded for diagnosis and logged when it changes, because a different model or region does not move a document into a different job.

## Partition count and process count

`num_workers` decided both how the input was divided and how many processes ran the pieces. Those are now separate: `num_workers` divides, `num_processes` runs, capped at the core count. Without this a restart on a smaller host cannot divide as the original run did. It is a standalone fix and reverts on its own.

## Turning it on

Pass `run_id` and `run_plan_store` to `ExtractionPipeline`. With neither, extraction behaves as it does today. A `run_id` with nowhere to keep its plan is refused rather than ignored, so a run cannot look restartable without being so.

The auto-tuned path is not covered yet and raises an error naming the limitation. Resuming from a plan is the next piece; this one only records and follows.

## Tests

Unit tests cover the store and the wiring. Live tests against real S3 cover what a mocked client cannot: a conditional write refusing the second writer of one run id, a missing object reading as a run that has not started, and the run directory staying out of the collection listing. Removing the conditional write makes the second write succeed silently against S3, and the live test catches it.

Unit suite 2393 passing. Live suite 34 passing against a real bucket, which includes the existing staging-recovery tests, since this changes how the collection record decides a collection is empty.

The plan needs the ordered document ids, so the collection is read and its ids derived before the plan is resolved. Only the id rewriter runs at that point; the pre-processors stay in the batch loop, so a restart that is about to be refused has not paid for them. The documents are held for the length of the run, which the streaming path previously avoided. Worth measuring on a large collection before the flag comes off.

This carries two concerns, the partition and process split and the plan itself, and is larger than one PR should be. The split is the natural cut if you would rather review them apart.


lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/extract/extraction_pipeline.py
```diff
@@ -36,7 +37,47 @@
...
+    order.
+    """
+    if not document_ids:
+        return source_documents
+
+    by_document_id = defaultdict(deque)
+    for source_document in source_documents:
+        by_document_id[_document_id(source_document)].append(source_document)
+
+    return [by_document_id[document_id].popleft() for document_id in document_ids if by_document_id[document_id]]
```
oussamahansal: The `if by_document_id[document_id]` guard is dead: `_check_against` has already proven the recorded and current multisets are equal before `resolve` returns a recorded plan, and on the fresh-write path `document_ids` is derived from this same list. If it ever did fire it would silently drop documents from extraction while the run reported success, which is the opposite of what the plan is for. I think you should drop the guard here or raise naming the missing id

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/extract/extraction_pipeline.py
```diff
@@ -405,12 +463,61 @@ def extract(self, inputs: Iterable[SourceType]):
...
+        run of the same id recorded.
+
+        Documents come back in the plan's order, so a restart on a host with
+        fewer cores divides them as the original run did.
+
+        Only the id rewriter runs here, because ids are all the plan needs. The
+        pre-processors stay in the batch loop, where they see a batch at a time
+        and a restart that is about to be refused has not paid for them.
+        """
+        source_documents = list(self.id_rewriter.handle_source_docs(
```
oussamahansal: Running the id rewriter over the whole input here means ids are now fixed against un-pre-processed content, and because `_new_id` short-circuits on `id_.startswith('aws:')` the `handle_source_docs` call at line 559 becomes a no-op on this path. The docstring explains why the pre-processors stay in the loop but not that their effect on ids is now gone. a `SourceDocParser` implementation that edits text or metadata would produce different ids under `run_id` than without it. `InferClassifications` only sets `self.classifications` so nothing in-tree is affected but `SourceDocParser` is a public extension point. I think you should either move the rewrite after the pre-processors or document that pre-processors must not alter node content.

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/extract/extraction_pipeline.py
```diff
@@ -245,11 +286,24 @@ def __init__(self,
         include_classification_in_entity_id = coalesce(include_classification_in_entity_id, GraphRAGConfig.include_classification_in_entity_id)
         extract_timestamp = kwargs.pop('extract_timestamp', None)
         source_id_width = kwargs.pop('source_id_width', None)
-
-        if num_workers > multiprocessing.cpu_count():
+        partition_workers = kwargs.pop('partition_workers', None)
```
oussamahansal: `partition_workers` has no caller outside the tests the plan path pins the count via `plan.num_workers` in `_extract_fixed_batch`, so this is a second mechanism for the same job reachable only from test code. Is it the seam for pinning on the auto-tuned path later (where `run_id` raises), or should the plan path use it?

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/extract/extraction_pipeline.py
```diff
@@ -405,12 +463,61 @@ def extract(self, inputs: Iterable[SourceType]):
...
+        source_documents = list(self.id_rewriter.handle_source_docs(
+            list(source_documents_from_source_types(inputs))
+        ))
+
+        plan = self.run_plan_store.resolve(RunPlan(
+            run_id=self.run_id,
+            document_ids=[_document_id(source_document) for source_document in source_documents],
+            num_workers=self.num_workers,
+            batch_size=self.batch_size,
+            config=GraphRAGConfig.get_config_snapshot()
```
oussamahansal: `get_config_snapshot()` filters fields by `pickle.dumps`, but `RunPlan.to_json` serialises with `json.dumps` Every field that survives the filter today is a scalar (`OpenSearchServerlessGeneration` and `SourceIdWidth` are `str`/`int` mixins, so both the dump and the post-round-trip comparison work), but a future picklable-but-not-JSON-serialisable field a `set` a `datetime` a plain `Enum` would blow up the plan write at run start before any work is submitted. I think it is worth a `default=str` or a serialisability filter on the config before it goes into the plan

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/extract/extraction_pipeline.py
```diff
@@ -36,7 +37,47 @@
...
+
+    The id rewriter derives it from the document's content, so it is the same
+    id on every run. Chunks carry it on their source relationship; a document
+    that has not been chunked yet is its own single node.
+    """
+    if source_document.refNode is not None:
+        return source_document.refNode.node_id
+
+    if not source_document.nodes:
+        return None
```
oussamahansal: A `SourceDocument` with no `refNode` and no nodes records `null` in `document_ids`, which `RunPlan` types as` List[str]`. It round-trips, but if that prefix has content on the restart the multiset check reports it as one id gone and another added and refuses the run. how about you raise here instead of naming a document `null`

lexical-graph/tests/unit/indexing/extract/test_extraction_pipeline_auto_tune.py
```diff
@@ -571,3 +571,55 @@ def test_a_source_arriving_once_still_ends(self):
         emitted = list(self._pipeline()._emit_extracted(nodes))
 
         assert [sd.final_part for sd in emitted] == [True]
+class TestAutoTunedRoundsKeepThePinnedPartitionCount:
```
oussamahansal: add some two blank lines above

lexical-graph/tests/integration/indexing/extract/test_run_plan_live.py
```diff
@@ -0,0 +1,205 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
oussamahansal: Should this be included with the integration test suite for Restart mechanism?

oussamahansal (changes requested):

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/extract/extraction_pipeline.py
noel-improv: Now raises `RunPlanMismatch` naming the id. It's a guard behind `resolve()`, and the comment says so. 19cff8e9

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/extract/extraction_pipeline.py
noel-improv: Documented, not moved: ids are fixed from content as supplied, and a `SourceDocParser` used with run plans must not alter node content. Moving the rewrite would make plan identity depend on pre-processor output. 19cff8e9

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/extract/extraction_pipeline.py
noel-improv: It's the seam for the auto-tuned path, where `run_id` raises. The plan path ignores it and uses `plan.num_workers`. Kept for that; can drop it if auto-tune never gets restart.

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/extract/extraction_pipeline.py
noel-improv: Filtered rather than `default=str`, which would make the recorded value compare unequal to the live one on every restart. Non-JSON settings are left out at DEBUG. 19cff8e9

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/extract/extraction_pipeline.py
noel-improv: Raises now, at plan-write time, rather than recording `null` and refusing the restart later. 19cff8e9

lexical-graph/tests/unit/indexing/extract/test_extraction_pipeline_auto_tune.py
noel-improv: Done, in 71e4a760 — that class belongs to #581.

lexical-graph/tests/integration/indexing/extract/test_run_plan_live.py
noel-improv: It's a pytest live test against a bucket, not a stack test — it pins S3's conditional-write and missing-key behaviour. Stack coverage is the two `run_restart` tests in `lexical.short`, plus a `lexical.restart` suite in #581.

lexical-graph/tests/integration/indexing/extract/test_run_plan_live.py
```diff
@@ -0,0 +1,205 @@
...
+    keys = [{'Key': o['Key']} for page in pages for o in page.get('Contents', [])]
+    if keys:
+        s3_client.delete_objects(Bucket=S3_TEST_BUCKET, Delete={'Objects': keys})
+
+
+def _store(key_prefix):
+    return RunPlanStore(
+        bucket_name=S3_TEST_BUCKET,
+        key_prefix=key_prefix,
+        collection_id=COLLECTION_ID,
```
mykola-pereyma: The gitleaks alert on this file is a false positive — it matches the token `key_prefix=key_prefix` (a `key`-ish name assigned a 10+char value), not a real secret. Renaming just this helper's parameter (the public `RunPlanStore(key_prefix=...)` kwarg stays) makes the value `prefix` too short to match, so the scanner won't re-raise it — no suppression comment needed. Callers pass positionally, so they're unaffected.

```suggestion
def _store(prefix):
    return RunPlanStore(
        bucket_name=S3_TEST_BUCKET,
        key_prefix=prefix,
        collection_id=COLLECTION_ID,
```

After this lands, please also dismiss the two existing code-scanning alerts as false positives — changing the line stops future re-raises but won't auto-close the current ones.

mykola-pereyma (commented):
One suggestion to clear the gitleaks false positive surgically (no rename of the real `key_prefix` API). Otherwise no security blockers from my side — see the review comment.

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/extract/run_plan.py
```diff
@@ -0,0 +1,158 @@
...
+        later build still restarts rather than failing on every attempt.
+        """
+        recorded = json.loads(body)
+        known = {f.name for f in fields(cls)}
+
+        unknown = sorted(set(recorded) - known)
+        if unknown:
+            logger.warning(f'Ignoring run plan fields this version does not know {unknown}')
+
+        return cls(**{name: value for name, value in recorded.items() if name in known})
```
mykola-pereyma: `cls(**{...})` here raises a bare `TypeError` if the stored JSON is missing a required field (`run_id`/`document_ids`/`num_workers`/`batch_size`) — e.g. a truncated or corrupt plan object. Consider checking for missing required fields before constructing and raising `RunPlanMismatch` (or `ValueError`) naming them, so a corrupt plan fails clearly rather than through a constructor error. The existing unknown-field handling above is a nice model for the symmetric missing-field case.

mykola-pereyma (changes requested):
Requesting two functional items be addressed before this goes in (both correctness, not security — the security side is clean):

**1. Plan identity can silently diverge when a `SourceDocParser` mutates content (the open thread on `extraction_pipeline.py:540`).** Running `IdRewriter` over the whole input fixes ids against un-pre-processed content and makes the later `handle_source_docs` call a no-op on this path, so a `SourceDocParser` that edits text or metadata would produce *different* ids under `run_id` than without it — and the recorded plan's `document_ids` would no longer line up on restart. Right now this is held only by the docstring convention ("a `SourceDocParser` used with run plans must not alter content"). Because the failure is silent (or surfaces later as a confusing `RunPlanMismatch`/misassignment), please make it enforced rather than documented: e.g. assert/guard that the ids after pre-processing equal the ids the plan was built from (and raise a clear error naming the offending doc if a parser changed them), or otherwise fail fast when a content-mutating parser is combined with a `run_id`. A loud runtime error beats a convention here.

**2. `RunPlan.from_json` raises a bare `TypeError` on a truncated/corrupt plan.** `cls(**{...known...})` will `TypeError` if a *required* field (`run_id`, `document_ids`, `num_workers`, `batch_size`) is missing from the stored JSON, rather than a clean, typed failure. Normal plans written by `to_json` always have every field, so this is an edge — but a partial/corrupt object in S3 currently blows up unhelpfully. Please validate the required fields are present and raise `RunPlanMismatch` (or a `ValueError`) naming what's missing, so a bad plan reports clearly instead of surfacing a constructor `TypeError`. See the inline note.

Everything else looks good — happy to re-review once these two are in.

PR #578, opened 2026-09-23
[FIX] Let a secret-scan finding be fixed in the next commit

noel-improv:
The scan reads the commits a change adds, not the tree it produces, so a finding stays red until the commit that introduced it leaves the range. Fixing the file in a later commit does not clear the check. The only ways out are rewriting the branch or suppressing the finding, and on a branch someone has already built on, both are expensive.

This came up on #577: `generic-api-key` matched `bucket_name=..., key_prefix=key_prefix, collection_id=COLLECTION_ID` in a new test. No secret, just three keyword arguments on one line. Splitting them fixed the file, the check stayed red, and the branch had to be squashed.

## Change

Add a working tree scan and let it decide whether the job passes. The history scan still runs and still uploads its SARIF, now as a warning rather than a failure.

Both scans stay because they answer different questions. A secret that reached a commit needs rotating whether or not a later commit removed it, and only the history scan sees that. Whether the code as it stands carries a secret is what should block a merge, and only the tree scan sees that.

## Verified

Against the #577 branch before it was squashed, where the line was added in one commit and fixed in the next:

```
history scan  -> leaks found: 1     (warning, uploaded to the Security tab)
tree scan     -> no leaks found     (job passes)
```

`gitleaks dir` exists in 8.30.1, the version this workflow pins.

## Tradeoff

A secret introduced and removed inside one pull request no longer blocks the merge. It is still reported, and rotation is the real remedy there — the commit is in the branch either way. If the preference is to keep blocking that case, the alternative is narrowing `generic-api-key` so it stops matching identifier assignments, which is more tuning and catches fewer shapes.


mykola-pereyma (approved):
LGTM. Two clean changes.

The test workflows (benchmarks/byokg/lexical) just add `v*.*.x` to the push + pull_request branch filters so the suites run on release-branch PRs — straightforward and correct.

The secret-scan refactor is well reasoned and doesn't weaken detection, it broadens it: the new `gitleaks dir .` working-tree scan still hard-fails on any secret currently in the checkout (anywhere, not just this PR's commit range), SARIF still uploads to the Security tab, and the weekly full-history scan still hard-fails as the rotation backstop. The only relaxation — a secret that was in an earlier commit but is no longer in the tree becomes a `::warning::` ("needs rotating") rather than a hard block — is the right call: rotation, not branch-rewriting, is the remedy once a secret hits a commit. Nice touch installing gitleaks under `runner.temp` so the tarball isn't caught by the tree scan.

One heads-up for whoever merges the stack: because the working-tree scan now runs on release-branch PRs, the `key_prefix=key_prefix` false positive on #577 will hard-fail there until the one-line rename (already suggested on that PR) or an allowlist lands — worth clearing before/with #577.

PR #579, opened 2026-09-23
[FEATURE] Record the job each extraction partition ran under, and reuse it

noel-improv:
## Description

**Stacked on #577.** That PR's commits are the first two here and are not part of this review; the change under review is `feat(lexical-graph): record the job each partition ran under, and reuse it`.

A restart could divide its input the way the interrupted run did, but not tell which of those partitions had already run. Every one was submitted again, at the price of the jobs the first run had already paid for.

## Changes

- Each partition records what a run did with it: the job's name and ARN, the attempt, where its output landed, and whether that output was read. The records live beside the plan, under `<key_prefix>/<collection_id>/_runs/<run_id>/`.
- A restart reads the record before submitting. A completed job is downloaded and parsed rather than invoked again. A job in any other state, one that cannot be read, or one whose output parses to nothing counts as having produced nothing, and the partition is resubmitted as its next attempt. That costs a job rather than the chunks.
- A partition names itself by its stage and a digest of the node ids it holds. The plan fixes how the collection divides, so a restart forms the same partitions and asks after them by the same names. Nothing has to be threaded down to the worker processes, which are handed nodes and nothing else.
- Job names carry the run, the partition, the attempt and the submission, inside Bedrock's 63 characters, so a job in the console says where it came from.
- Records roll up into one object per run once the workers are done, and the rollup answers a read for the partitions folded into it. A run that dies leaves its records where a restart still finds them.

## Problem

Related issue (if any): #325

## Why the stage is part of a partition's name

A pipeline runs both the proposition extractor and the topic extractor over the same nodes, and they divide identically. Named by a digest of the node ids alone, the topic stage found the proposition stage's record, saw its job completed, and served the proposition output as topics. On a first run, not only on a restart. Review caught it; there is a test for it.

## Why the record goes down before the wait

`create_and_run_batch_job` now calls back with the ARN as soon as the job exists, and the record is written there rather than after the wait. The wait is the part that can outlive the process, and a job nobody recorded is a job a restart pays for twice.

The name carries the submission's own suffix for the same reason in reverse: a job created whose record never landed leaves a restart computing the same attempt, and Bedrock refuses a name already in flight, which would wedge that partition on every restart after.

## Testing

- [x] Existing tests pass (`pytest`) - 2,436 in `tests/unit` with this branch alone

Live tests against a real bucket cover what a mocked client cannot: the rollup merge writing one object and then removing the records it folded in, and the run directory staying out of the collection listing. Removing the marker requirement from the resume read fails ten of them.

One pre-existing failure is unrelated and reproduces on a clean tree: `test_integ_dependency_compatibility.py` errors on pip resolution.

## Checklist

A run without a `run_id` keeps no records and behaves as it does today, including its job names.


lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/extract/batch_extractor_base.py
```diff
@@ -94,10 +106,97 @@ def _run_non_batch_extractor(self, nodes):
...
+            results = list(process_batch_output_sync(output_dir, record.input_filename, self.llm))
+        except Exception as e:
+            logger.warning(
+                f'[{self.description} batch] Could not read the output of a completed job, '
+                f'resubmitting [partition: {record.partition_id}, job_name: {record.job_name}, '
+                f'error: {e!s}]'
+            )
+            return None
+
+        if not results:
```
oussamahansal: The guard is emptiness, not coverage, so a completed job whose output parses to one of the partition's two hundred nodes is accepted and the partition sealed `COMPLETE`. The other 199 fall through to step 4 of `_process_nodes`, which sets their extraction to `None` behind a `debug` log, and the record means no restart will ever retry them. The test helper `_run` mocks `process_batch_output_sync `to return one result for a three-node batch and asserts `COMPLETE`, so the suite pins the behaviour its own docstring says the feature exists to prevent  compare `len(results)` against `len(node_batch) `here.

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/extract/batch_extractor_base.py
```diff
@@ -94,10 +106,97 @@ def _run_non_batch_extractor(self, nodes):
...
+        try:
+            status = bedrock_client.get_model_invocation_job(jobIdentifier=record.job_arn)['status']
+        except Exception as e:
+            logger.warning(
+                f'[{self.description} batch] Could not read the job an earlier run submitted, '
+                f'resubmitting [job_arn: {record.job_arn}, error: {e!s}]'
+            )
+            return None
+
+        if status != 'Completed':
```
oussamahansal: Treating `InProgress` as having produced nothing resubmits a job that is still running and that is the state most likely to be found right after a crash, the original job runs to completion and bills anyway, the new one bills again, and both count against the concurrent-job quota. The parametrized test says this is deliberate but `InProgress`/`Submitted`/`Validating` are the states where waiting on `record.job_arn` is strictly cheaper than a resubmission. I think it is worth splitting them from the terminal states

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/extract/extraction_pipeline.py
```diff
@@ -405,12 +473,71 @@ def extract(self, inputs: Iterable[SourceType]):
...
+            if self._auto_tune:
+                raise ValueError(
+                    'A run plan does not yet cover the auto-tuned path. Run without run_id, '
+                    'or without an auto-tuning batch extractor.'
+                )
+            source_documents, plan = self._follow_run_plan(inputs)
+            yield from self._extract_fixed_batch(source_documents, plan=plan)
+            # A run that dies never reaches here, and leaves its records where
+            # a restart still finds them.
+            self.manifest_store.merge_rollup(GraphRAGConfig.s3)
```
oussamahansal: This is the only call that makes a restart read one object instead of one per partition,and it has no test through the pipeline every `merge_rollup` test in the suite calls the store directly. Can you add a test that a plan run rolls up at the end and that a run abandoning the generator does not?

lexical-graph/tests/unit/indexing/extract/test_run_manifest.py
```diff
@@ -0,0 +1,276 @@
...
+    def test_a_partition_still_working_keeps_its_record(self):
+        store, s3_client = _store(), _s3_holding({})
+        store.write(_record('working', state=SUBMITTED), s3_client)
+
+        store.merge_rollup(s3_client)
+
+        assert store.partition_key('working') not in s3_client.deleted
+        assert store.read('working', s3_client).state == SUBMITTED
+
+    def test_a_rolled_up_record_is_removed_only_after_the_rollup_is_written(self):
```
oussamahansal: The test name promises the record is removed only after the rollup is written, but it asserts that both happened not the order `written` and `deleted` are separate logs, so a `merge_rollup` that deleted first would still pass. I think you should have the mock append both into one event list and assert the rollup key precedes the deletes. also make sure it passes in the integrations tests

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/extract/batch_extractor_base.py
```diff
@@ -94,10 +106,97 @@ def _run_non_batch_extractor(self, nodes):
...
+                f'[partition: {record.partition_id}, job_name: {record.job_name}]'
+            )
+            return None
+
+        logger.info(
+            f'[{self.description} batch] Downloading the output of a job an earlier run '
+            f'completed [partition: {record.partition_id}, job_name: {record.job_name}]'
+        )
+
+        output_dir = os.path.join(self.batch_inference_dir, 'recovered', record.partition_id)
```
oussamahansal: `recovered/<partition_id>` is never removed even under `delete_on_success` while the submit path deletes its `root_dir` at the end `_prepare_directory` only creates, so a second recovery of the same partition downloads on top of the first one's files. I would say maybe clean the directory on the way in and remove it alongside `root_dir`

lexical-graph/tests/integration/indexing/extract/test_run_plan_live.py
```diff
@@ -0,0 +1,190 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
oussamahansal: Also make sure this is also part of the integration test suite

oussamahansal (changes requested):

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/extract/batch_extractor_base.py
noel-improv: Right, and the test helper was pinning the wrong behaviour. The output now has to describe every node in the partition, and the same rule applies to a job this run submits: a short output leaves the record as `submitted` so a restart redoes it, while what the job did describe is still used. One rule, `_describes_every_node`, both paths. Mutated to always-yes it fails five tests. 07aca7f8

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/extract/batch_extractor_base.py
noel-improv: Split as you suggested. `Submitted`/`Validating`/`Scheduled`/`InProgress` are waited on via `wait_for_job_completion`; a wait that ends badly resubmits. `Stopping` stays with the terminal states since it can only end `Stopped`. 07aca7f8

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/extract/extraction_pipeline.py
noel-improv: Added: a plan run rolls up once at its end, a generator abandoned mid-run does not, and a run without a `run_id` has nothing to roll up. 07aca7f8

lexical-graph/tests/unit/indexing/extract/test_run_manifest.py
noel-improv: The mock now keeps one event list for puts and deletes, and the test asserts the rollup's put precedes the record's delete. 07aca7f8

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/extract/batch_extractor_base.py
noel-improv: Cleared on the way in and removed under `delete_on_success`, with a test that a second recovery does not read the first one's files. 07aca7f8

lexical-graph/tests/integration/indexing/extract/test_run_plan_live.py
noel-improv: It's a pytest live test against a bucket, not a stack test — it pins S3's conditional-write and missing-key behaviour. Stack coverage is the two `run_restart` tests in `lexical.short`, plus a `lexical.restart` suite in #581.

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/extract/run_manifest.py
```diff
@@ -0,0 +1,215 @@
...
+
+        # Checked here as well as on every path built from it, so a run id that
+        # climbs out of the collection is refused before any work starts.
+        self.run_path(run_id)
+
+        self.run_id = run_id
+        self._rollup = None
+
+    def partition_key(self, partition:str) -> str:
+        return join(self.run_path(self.run_id), PARTITION_DIR, f'{partition}.json')
```
mykola-pereyma: Please guard `partition` here with the #574 validator before it's joined:

```suggestion
    def partition_key(self, partition:str) -> str:
        validate_id_segment(partition, 'partition_id')
        return join(self.run_path(self.run_id), PARTITION_DIR, f'{partition}.json')
```
(add `from graphrag_toolkit.lexical_graph.utils.id_validation import validate_id_segment` at the top.)

Why it matters: `partition_id()` is a safe hash on the write path, but this value also arrives via `PartitionRecord.from_json` (see the read-back path), and it flows unvalidated into `os.path.join(...)` + `shutil.rmtree(...)` in `batch_extractor_base._recover_partition` — a `../../..` (or absolute) id would resolve outside the batch dir. It's the same id→path sink class #574 guarded on `run_id`. Note this is heavily gated (needs write access to the run's manifest object in the bucket AND a Completed Bedrock job ARN in the record before the rmtree is reached), so it's defense-in-depth, not a reachable external traversal — but guarding it here (one line) closes it cleanly for every key and every downstream join.

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/extract/batch_extractor_base.py
```diff
@@ -94,10 +111,122 @@ def _run_non_batch_extractor(self, nodes):
...
+            return self._resubmitting(record, f'An earlier run left this partition in {status}')
+
+        logger.info(
+            f'[{self.description} batch] Downloading the output of a job an earlier run '
+            f'completed [partition: {record.partition_id}, job_name: {record.job_name}]'
+        )
+
+        # A second recovery of the same partition must not read the first one's
+        # files alongside its own.
+        output_dir = os.path.join(self.batch_inference_dir, 'recovered', record.partition_id)
```
mykola-pereyma: This is the destructive sink that motivates guarding `partition_id`: `record.partition_id` comes from `manifest_store.read(...).from_json(...)`, so its value is whatever the stored JSON holds — not necessarily the safe key hash. `os.path.join(self.batch_inference_dir, 'recovered', '../../..')` (or an absolute id) resolves outside the batch dir, and `rmtree` (here and again at line 207 under `delete_on_success`) would delete the resolved target. Reaching it needs a Completed `job_arn` in the record too, so it's gated/defense-in-depth — but validating `partition_id` at `partition_key()`/`from_json` (see the run_manifest.py comment) removes the possibility entirely rather than relying on the upstream hash.

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/extract/run_manifest.py
```diff
@@ -0,0 +1,215 @@
...
+    def from_json(cls, body:str) -> 'PartitionRecord':
+        """A record read back, without the fields a later build added."""
+        recorded = json.loads(body)
+        known = {f.name for f in fields(cls)}
+
+        unknown = sorted(set(recorded) - known)
+        if unknown:
+            logger.warning(f'Ignoring partition record fields this version does not know {unknown}')
+
+        return cls(**{name: value for name, value in recorded.items() if name in known})
```
mykola-pereyma: `cls(**{...})` raises a bare `TypeError` if the stored JSON is missing a required field (`partition_id`/`attempt`/`state`) — e.g. a truncated or partially-written record — rather than a clean typed failure. Consider checking the required fields are present and raising a typed error naming what's missing (the unknown-field handling just above is a good model for the symmetric case). This is also the natural place to `validate_id_segment(recorded.get('partition_id', ''), 'partition_id')` so a round-tripped record can't carry a path-bearing id downstream.

mykola-pereyma (changes requested):
The functional items from the earlier round (the 1-of-N `_describes_every_node` data-loss fix, the in-progress-job resubmission, the recovered-dir reuse, the rollup ordering) look well handled at head — thanks. Requesting two smaller things before it lands, both about `partition_id` handling. Neither is a live external vuln (see the inline note on the reachability gating), so treat this as hardening + robustness rather than a critical block.

1. `partition_id` reaches path/key sinks — including a destructive `shutil.rmtree` — without the `validate_id_segment` guard #574 standardised. It's a safe hash on the write path, but a `PartitionRecord` round-tripped through `from_json` carries whatever `partition_id` the stored JSON holds. Adding one `validate_id_segment(partition, 'partition_id')` in `partition_key()` (and/or in `from_json`) closes it consistently with `run_path`'s guard on `run_id`.

2. `PartitionRecord.from_json` raises a bare `TypeError` on a truncated/corrupt record missing a required field — same nit as on #577's `RunPlan.from_json`. A clear typed error would read better.

Inline details below. Happy to re-review once these are in.

PR #580, opened 2026-09-23
[FEATURE] Read what an interrupted extraction left before spending again

noel-improv:
## Description

**Stacked on #579, which is itself stacked on #577.** Only the last two commits are part of this review: `feat(lexical-graph): read what an earlier attempt left before spending again` and `feat(lexical-graph): build the staging handler a restart needs in one call`.

With the plan recording how a run divides and the partition records saying which of those divisions already ran, a restart can read both before it spends anything. It could not before: it discovered finished work one partition at a time, inside the workers, after the run had started, and it stored every document again including the ones already sitting whole in the collection.

## Changes

- `plan_resume` reads the partition records and takes one listing of the collection, then logs what the run is about to redo and which jobs an earlier attempt left unfinished. An operator sees this before a multi-day run starts.
- The sources that listing finds are handed to the staging handler, which leaves them where they are.
- `resuming_handler(run_plan_store, run_id)` builds that handler in one call, deriving the bucket, prefix and collection from the plan store.

## Problem

Related issue (if any): #325

## What the skip does not change

The division. Every document still flows through the pipeline and partitions are formed exactly as the plan says, so the partition records still line up. Only the writes are skipped, and a skipped document stays in the stream the handler yields.

Dropping documents before the run divides its input would have been simpler and wrong: fewer documents divide differently, so the partitions change, their ids change, and the records written against the old ids no longer match.

## What counts as already stored

A source counts as stored when its markers account for the chunks beside them, which is what the reader checks at build time. Anything less is unsound: one source can be stored as several documents under one prefix, each with its own marker, so a prefix holding a marker can still be missing the part that never landed. There is a test for exactly that.

The chunk format keys every chunk by its node id, so one listing supplies what is present and only the markers are read. The JSONL format keeps many nodes to an object, where a listing cannot reach their ids, so it is answered with nothing and such a run stores its documents again. `S3BasedDocs` ignores a skip set for that format rather than trusting a caller to know.

## Where the factory lives

Beside the driver rather than on `S3BasedDocs`, because the load package imports nothing from extract and a method there would need a lazy import to keep it that way. The pipeline cannot build the handler itself: it holds the records, and the handler is composed separately downstream of it.

## Testing

- [x] Existing tests pass (`pytest`) - 2,454 in `tests/unit`

Live tests against a real bucket cover the resume read in both storage formats, a source missing a chunk its marker names reading as not stored, and the operator surface end to end: stage a source, build a handler for the same run, stage both, and find the first untouched and the second written.

## Checklist

A handler built without `skip_source_ids` stores everything, as it does today.

The modules are not exported from `indexing/extract/__init__.py`, so an operator imports them by module path, as #577 does for the plan store. Exporting one of the three makes the set inconsistent, and exporting all three is a decision about the public surface rather than a convenience; it is written up as a follow-up rather than settled here.

The staging skip is the half of this that can be dropped if you would rather it waited. The read, the report and the factory stand without it.


lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/load/s3_based_docs.py
```diff
@@ -1090,9 +1121,33 @@ def accept(self, source_documents: List[SourceDocument], **kwargs: Any) -> Gener
...
-        for doc in self._uploader.upload(source_documents):
+        skipped = []
+
+        for doc in self._uploader.upload(self._not_already_staged(source_documents, skipped)):
             doc_count += 1
             yield doc
 
+        # Still part of this run's output, for a caller reading the stream
+        # rather than the collection.
+        for doc in skipped:
```
oussamahansal: Every skipped document is held in `skipped` and yielded only after the upload stream is exhausted, which breaks the ordering `S3ChunkUploader.upload `documents ("Documents are yielded in order" the nodeless branch goes out of its way to keep a document's place for that reason) `accept` sits mid-pipe (`... | Pipe(handler.accept) | build_pipeline | sink`) so on the restart this feature targets most of the collection already staged  it holds that entire set of `SourceDocuments` in memory and gives the build pipeline nothing for them until extraction has finished. Yield each skipped document in place instead of collecting it.

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/extract/resume.py
```diff
@@ -0,0 +1,128 @@
...
+def plan_resume(manifest_store, s3_client, for_jsonl:bool=False) -> ResumeReport:
+    """
+    What an earlier run of this id left behind, read before any work is
+    submitted so an operator sees what a long run is about to redo.
+    """
+    partitions:Dict[str, PartitionRecord] = manifest_store.read_partitions(s3_client)
+
+    outstanding = [record for record in partitions.values() if record.state != COMPLETE]
+
+    staged = staged_source_ids(
```
oussamahansal: `staged_source_ids` reads the collection, not this run, so `sources_staged` counts sources any earlier run staged. A brand-new `run_id` pointed at a populated collection therefore logs `Resuming a run `with `is_restart` true and no records at all, and `staging_handler's` promise that "a run with nothing behind it skips nothing" only holds for an empty collection. I think here you either scope the count to this run's records or say in both docstrings that the skip is collection-wide

lexical-graph/tests/unit/indexing/load/test_completion_marker.py
```diff
@@ -639,3 +639,49 @@ def test_a_prefix_missing_an_earlier_part_is_still_incomplete(self):
...
+
+    def _handler(self, skip_source_ids=None, for_jsonl=False):
+        from graphrag_toolkit.lexical_graph.indexing.load.s3_based_docs import S3BasedDocs
+
+        return S3BasedDocs(
+            region='us-east-1', bucket_name='b', key_prefix='p', collection_id='c',
+            for_jsonl=for_jsonl, skip_source_ids=skip_source_ids,
+        )
+
+    def test_a_source_already_staged_is_not_written_again(self):
```
oussamahansal: All four new tests call `_not_already_staged` directly, so the part `accept` adds buffering the skipped documents and yielding them after the upload stream has no unit coverage. Can you add a test that drives `accept` with a mocked uploader and asserts what comes out and in what order

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/extract/resume.py
```diff
@@ -0,0 +1,128 @@
...
+    The JSONL format keeps its node ids inside the objects, where a listing
+    cannot reach them, so it is answered with nothing.
+    """
+    if for_jsonl:
+        logger.debug('The JSONL format keeps its node ids where a listing cannot reach them')
+        return set()
+
+    collection_path = join(key_prefix, collection_id, '')
+
+    content = {}
```
oussamahansal: The module docstring sells this as the answer to "one call per document, which at a million documents is its own scale problem", but it accumulates every object key in the collection into `content` before comparing any of them. `list_objects_v2` returns keys in lexicographic order, so a source's keys arrive contiguously. In this case how about if we  compare and discard one source at a time rather than buffering the whole collection? let me know what you think

lexical-graph/tests/unit/indexing/load/test_completion_marker.py
```diff
@@ -639,3 +639,49 @@ def test_a_prefix_missing_an_earlier_part_is_still_incomplete(self):
         )
 
         assert not is_complete(['c', 'd'], list(markers), 'b', s3_client)
+class TestARunResumingItsOwnWork:
```
oussamahansal: 2 blank lines above

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/load/s3_based_docs.py
```diff
@@ -1090,9 +1121,33 @@ def accept(self, source_documents: List[SourceDocument], **kwargs: Any) -> Gener
...
+            yield doc
+
         end = time.time()
         logger.debug(f'Finished writing {doc_count} source documents to S3 [bucket: {self.bucket_name}, prefix: {collection_prefix}] ({end - start} seconds)')
```
oussamahansal: `doc_count` at counts only uploaded documents so the closing log undercounts what accept yielded

oussamahansal (changes requested):

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/load/s3_based_docs.py
noel-improv: Yielded in place now, merged by position against the uploader's in-order stream, and the input is read once — which matters because in the pipe `accept()` is handed a generator. An uploader that returns too few, too many, or never reads its input raises rather than losing documents quietly. ffc3c6ef

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/extract/resume.py
noel-improv: Said in both docstrings, and the report follows it: `is_restart` now means this run has partition records of its own, so a new `run_id` over a populated collection reads "Starting a run over a collection with sources already staged". Scoping the skip to the run isn't possible — staging keeps no run id, and a source stored whole isn't worth storing again under any run. ffc3c6ef

lexical-graph/tests/unit/indexing/load/test_completion_marker.py
noel-improv: The resume tests now drive `accept()` through a mocked uploader and assert what comes out and in what order, including a generator input. ffc3c6ef

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/extract/resume.py
noel-improv: Took your suggestion. S3 lists a general purpose bucket in order, so each source is compared as the listing moves past it and only the source in hand is held. ffc3c6ef

lexical-graph/tests/unit/indexing/load/test_completion_marker.py
noel-improv: Done. ffc3c6ef

lexical-graph/src/graphrag_toolkit/lexical_graph/indexing/load/s3_based_docs.py
noel-improv: Counts skipped documents too, and the log now says written vs already staged. ffc3c6ef

mykola-pereyma (commented):
Reviewed the increment over #579 (resume.py, s3_based_docs.py, run_plan.py). From the security side it's clean — no change request from me:

- `resume.py` is read-only planning: `staged_source_ids` does one ordered S3 listing and a completeness comparison; a `source_id` derived from bucket keys only builds an S3 prefix string and joins a set — it never reaches a local-path join/open/rmtree. No pickle, no new id→path sink.
- `s3_based_docs._staged_in_order` uses `doc.source_id()` only as a set-membership test (skip re-staging a source stored whole), preserves the in-order contract, and fails loud (RuntimeError) rather than dropping/reordering documents. No new local-path construction, and it doesn't remove the `validate_id_segment` guards from #574.
- `run_plan.RunPlanStore.staging_handler` just wires `plan_resume(...)` → `S3BasedDocs(skip_source_ids=...)`; the `TYPE_CHECKING` import avoids a circular import. No sink.

The functional points from the earlier round (in-order staging merge, `is_restart` semantics, and streaming the staged-source comparison one source at a time) look addressed at head — thanks.

One housekeeping item — the gitleaks alert here is the same `key_prefix=key_prefix` false positive as #577, but in a **different file**, so #577's fix won't cover it. It appears at four lines in `tests/integration/indexing/extract/test_run_manifest_live.py`: **62, 88, 242, 258**. Please clear them the same way (rename the local so the value isn't a 10+char `key`-named token, e.g. `prefix=…` / `key_prefix=prefix`), or add a repo-wide `.gitleaks.toml` allowlist for the pattern (one-shot across #577 + here). Heads-up: once #578 lands, its new working-tree scan (`gitleaks dir .`) will hard-fail this PR on these lines until they're cleared — this isn't covered by fixing #577's `test_run_plan_live.py`.

PR #581, opened 2026-09-23
[FEATURE] Cover extraction restart on the integration stack

noel-improv:
## Description

**Stacked on #580, which is stacked on #579 and #577.** Only the last commit is part of this review: `test(lexical-graph): cover restart on the integration stack`.

The restart work carries unit tests and live S3 tests. Nothing exercised it where the release is cut: real stores, the real suite runner, real Bedrock batch jobs.

## Changes

Three test classes, split by what they cost.

- `run_restart.ExtractWithRunPlan` and `run_restart.ResumeInterruptedExtraction` in `lexical.short`. Synthetic documents on the synchronous path, so neither submits a Bedrock job.
- `batch_extract.BatchExtractWithRunPlanToS3` in `lexical.long`, the only one that costs anything. It extracts under a run plan, then restarts the same run id.
- A guard for the auto-tuned path: four unit tests beside the other auto-tune behaviour, and one assertion through the public API.

## Problem

Related issue (if any): #325

## How an interruption is staged

A run cannot be killed part way: the suite runs its tests in one process, in sequence. The interrupted state is built instead, the way the staging-recovery tests added with #560 build one.

Two sources are damaged in different ways, because two different checks refuse them. One loses its completion marker, so nothing closes it. Another keeps its marker and loses a chunk, which is the comparison the marker exists for. Damaging both the same way left the second check untested, which is what the first version of this did.

## Two assertions that were worth rewriting

The worker-count assertion compared two reads of the plan. A recorded plan is never overwritten, so both reads answered from the same object and the test held by construction: it passed even when the restart ignored the recorded count. It now watches what the run divided its input into. Removing `num_workers = plan.num_workers if plan else self.num_workers` fails it with `[1, 1] != [2, 2]`.

The guard asserted that `_extract_auto_tuned` was not called, which the run-plan branch never reaches whatever the guard does. It now asserts the plan is never resolved, and moving the refusal below `_follow_run_plan` fails it.

## What these cannot reach

Worth saying so nobody goes looking. A process dying mid-run, so a partially written record or a job whose record write was lost. A Bedrock job in `Failed`, `Stopped` or `Expired`, which cannot be produced on demand. Worker-count adoption across differently sized hosts. The conditional-write race in `RunPlanStore.resolve`. All four stay unit-tested.

`test_a_restart_submits_no_new_jobs` is weaker than it reads: `list_model_invocation_jobs` is eventually consistent, so a job created seconds earlier may not be listed yet. The attempt-count assertion beside it is the load-bearing one.

## Testing

- [x] Existing tests pass (`pytest`) - 2,460 in `tests/unit`

Both cheap classes were run through the real `IntegrationTestHandler`, the object `test_suite.py` uses, and both report PASS: seven assertions and nine assertions. Each load-bearing assertion was mutation-checked, and each fails when the behaviour it names is removed.

The paid class has not run; it needs the stack. Its central claim was proven separately against a real completed Bedrock job in our account: 735 records recovered from S3, matching the job's `successRecordCount`, with submission wired to raise and never firing. Removing the job from the record makes the same path submit.

One pre-existing failure is unrelated and reproduces on a clean tree: `test_integ_dependency_compatibility.py` errors on pip resolution.

## Checklist

`delete_prefix` went into `integration_test_base.py` rather than into one of the test modules. Both new modules need it, the package has no shared helper module, and nothing else in it imports test module from test module.

Both cheap classes restore `extraction_batch_size` and `extraction_num_workers` in a `finally`. They are process globals, the suite shares one process, and `falkordb`, `delete_sources` and `lexical_graph_cypher_safety` run afterwards without setting their own.


integration-tests/test-scripts/graphrag_toolkit_tests/batch_extract.py
```diff
@@ -247,4 +251,226 @@ def test_extracted_one_doc_for_each_url(self):
...
+                    def test_every_partition_is_recorded_complete(self):
+                        """Each partition the run divided its input into is recorded complete"""
+
+                        self.assertGreater(self._num_partitions, 0)
+                        self.assertEqual(self._partition_states, ['complete'])
+
+                    def test_a_job_name_says_which_run_it_belongs_to(self):
+                        """Every Bedrock job this run submitted carries the run id and an attempt"""
+
+                        self.assertGreater(len(self._job_names), 0)
```
oussamahansal: apply_extraction_doc_limit at line 352 is driven by BENCHMARK_EXTRACT_DOC_LIMIT (build-tests.sh --benchmark-extract-doc-limit), and with it set low a partition holds under 100 chunks, so batch_extractor_base.py:380 logs a warning and runs the non-batch extractor — no Bedrock job, no submitted record. assertGreater(len(self._job_names), 0) and assertGreater(self._num_partitions, 0) then fail and read like a restart bug rather than a configuration one. The pre-existing batch classes tolerate that flag because they only assert the doc count; this is the first one that can't, so handler.skip() when the limit is set would be a FAIL.

integration-tests/test-scripts/graphrag_toolkit_tests/run_restart.py
```diff
@@ -0,0 +1,500 @@
...
+            try:
+                extract(graph_index, docs + [
+                    Document(id_='test-4', text='Boggles can live for up to 14 years')
+                ])
+            except RunPlanMismatch as e:
+                added_document = str(e)
+
+            changed_batch_size = None
+            try:
+                GraphRAGConfig.extraction_batch_size = 3
```
oussamahansal: extraction_batch_size is left at 3 for the auto-tune call at line 229. It passes today only because the refusal at extraction_pipeline.py:478 runs before _follow_run_plan; if that order ever moved, the call raises RunPlanMismatch, the except ValueError at 235 doesn't catch it, and the exception escapes _run_test — the exact failure mode the comment at 191 exists to prevent. Reset it to 2 before building auto_tuned_index.

integration-tests/test-scripts/graphrag_toolkit_tests/integration_test_base.py
```diff
@@ -23,4 +26,15 @@ def init_test_details(self, handler:IntegrationTestHandler, params:Dict[str, Any
     
     def run_test(self, handler:IntegrationTestHandler, params:Dict[str, Any]):
         handler.start_test()
-        self._run_test(handler, params)
\ No newline at end of file
+        self._run_test(handler, params)
+
+
+def delete_prefix(bucket_name:str, prefix:str):
```
oussamahansal: delete_prefix doesn't guard the prefix, so an empty or mistyped one lists and deletes every object in the bucket  and --bucket is caller-supplied, not created by the suite. It also ignores the Errors key in the delete_objects response, so a partial delete reports success. Reject a falsy prefix and raise on Errors.

lexical-graph/tests/unit/indexing/extract/test_extraction_pipeline_auto_tune.py
```diff
@@ -571,3 +571,103 @@ def test_a_source_arriving_once_still_ends(self):
         emitted = list(self._pipeline()._emit_extracted(nodes))
 
         assert [sd.final_part for sd in emitted] == [True]
+class TestARunPlanIsRefusedOnTheAutoTunedPath:
```
oussamahansal: 2 blank lines before the class

integration-tests/test-scripts/graphrag_toolkit_tests/run_restart.py
```diff
@@ -0,0 +1,500 @@
...
+            after_first_run = objects_under(run.bucket_name, collection_path)
+
+            staged_sources = staged_source_ids()
+
+            # Two shapes of interruption, and they are not the same check.
+            # A source with no marker is refused because nothing closes it.
+            # A source that keeps its marker but lost a chunk is refused by
+            # the comparison the marker exists for, and only this second
+            # case reaches it.
+            unmarked_source = staged_sources[-1]
```
oussamahansal: staged_sources[-1] and [-2] assume all six sources staged. If fewer do, intact_sources at line 374 silently shrinks, the damage lands on different sources than the comment describes, and the mismatch only surfaces later at line 485 as a confusing count failure. An assertEqual(len(staged_sources), NUM_DOCS)-style check before the slicing makes the precondition the failure.

integration-tests/test-scripts/graphrag_toolkit_tests/run_restart.py
```diff
@@ -0,0 +1,500 @@
...
+                    """Collection holds every input document once the restart finishes"""
+
+                    self.assertEqual(len(self._sources_after), NUM_DOCS)
+
+                def test_the_jsonl_format_skips_nothing(self):
+                    """A JSONL handler is given nothing to skip, because a listing cannot prove it"""
+
+                    self.assertEqual(self._jsonl_skipped, [])
+
+                def test_the_report_says_what_the_restart_will_reuse(self):
```
oussamahansal: This class builds LexicalGraphIndex with no batch config, so no partition records exist and is_restart is decided entirely by sources_staged (resume.py:88) — 'Resuming a run' would read identically for a brand-new run id against this collection. The sources already staged: 4 half does pin the marker and chunk damage, so it's the "says what the restart will reuse" framing that overstates what's checked; the partition counts in the report go unasserted because nothing on this path writes them.

integration-tests/test-scripts/graphrag_toolkit_tests/batch_extract.py
```diff
@@ -247,4 +251,226 @@ def test_extracted_one_doc_for_each_url(self):
...
+
+                        self.assertEqual(self._rewritten_keys, [])
+
+                handler.run_assertions(BatchRunPlanAssertions)
+
+        finally:
+            # A failure is when the long suite leaves the most behind, so this
+            # runs on every path. Nothing downstream reads either prefix.
+            delete_prefix(s3_results_bucket, f'{extracted_prefix}/{collection_id}/')
+            delete_prefix(s3_results_bucket, f'{batch_inference_prefix}/')
```
oussamahansal: Deleting batch_inference_prefix in finally removes the Bedrock job input and output on the failure path, which is when they're the only thing left to diagnose from. The comment justifies running on every path because a failure leaves the most behind — that's also when you most want it kept. Consider deleting this prefix only after the assertions pass.

integration-tests/test-scripts/graphrag_toolkit_tests/batch_extract.py
```diff
@@ -247,4 +251,226 @@ def test_extracted_one_doc_for_each_url(self):
...
+                    def test_extracted_one_doc_for_each_source(self):
+                        """Extracted collection holds one source doc per input document"""
+
+                        self.assertEqual(self._num_extracted_docs, self._expected_num_docs)
+
+                    def test_every_partition_is_recorded_complete(self):
+                        """Each partition the run divided its input into is recorded complete"""
+
+                        self.assertGreater(self._num_partitions, 0)
+                        self.assertEqual(self._partition_states, ['complete'])
```
oussamahansal: Hardcodes 'complete' while the file already imports MAX_JOB_NAME from run_manifest; import COMPLETE from the same module so a state rename breaks the test instead of passing it.

integration-tests/test-scripts/graphrag_toolkit_tests/batch_extract.py
noel-improv: Good catch — it fails for a configuration the test never set. It now skips when `BENCHMARK_EXTRACT_DOC_LIMIT` is set, the way `query_gpu` skips off a GPU, before any config is touched. c3928b7f

integration-tests/test-scripts/graphrag_toolkit_tests/run_restart.py
noel-improv: Reset to 2 before the index is built, with the ordering hazard named in the comment. c3928b7f

integration-tests/test-scripts/graphrag_toolkit_tests/integration_test_base.py
noel-improv: Both done. A falsy or whitespace-only prefix raises before anything is listed, in c3928b7f; the `Errors` check went in earlier with 71e4a760.

lexical-graph/tests/unit/indexing/extract/test_extraction_pipeline_auto_tune.py
noel-improv: Already in, 71e4a760 — you were looking at the previous head.

integration-tests/test-scripts/graphrag_toolkit_tests/run_restart.py
noel-improv: Raises with the count and the source list where the assumption is made, so a short listing fails there rather than as a count later. c3928b7f

integration-tests/test-scripts/graphrag_toolkit_tests/run_restart.py
noel-improv: Right on both. `is_restart` now means this run has partition records of its own (ffc3c6ef), so on this path the report reads "Starting a run over a collection with sources already staged" rather than "Resuming a run", and the test no longer asserts the latter. Renamed to `test_the_report_counts_the_sources_it_will_leave_alone`, which is what it checks. c3928b7f

integration-tests/test-scripts/graphrag_toolkit_tests/batch_extract.py
noel-improv: Agreed — the failure path was removing the only thing left to diagnose from. The batch prefix is deleted inside the try once the assertions pass; the staged collection still goes on every path. c3928b7f

integration-tests/test-scripts/graphrag_toolkit_tests/batch_extract.py
noel-improv: Done, imports `COMPLETE` from `run_manifest`. c3928b7f

mykola-pereyma (commented):
LGTM on the merits — holding the formal approval until the stack below lands.

The increment is all tests + integration infra + one CFN line, no production `src/` changes. Reviewed and it's clean:
- CFN: the new `DeleteTestArtifacts` statement grants only `s3:DeleteObject` on `${S3Bucket}/*` (the test bucket) to the test role — appropriately scoped for the restart tests that damage a collection.
- `delete_prefix` is now safe: it raises on a falsy/whitespace prefix before listing anything (so an empty prefix can't wipe a caller-supplied bucket) and raises on any `delete_objects` Errors. The destructive test path is bounded.
- No hardcoded secrets/creds/ARNs in the new scripts, and the review thread (skip on doc-limit, batch-size reset, staged-count assertion, deleting the batch prefix only after assertions, importing `COMPLETE`) is resolved.

Pending approval, not because of anything in this PR, but because it's the top of the stack: it must merge after #577 → #579 → #580 (bottom-up), and its branch still carries the changes I requested on #577 (two functional items) and #579 (the `partition_id` guard + `from_json` robustness). Once those are resolved and merged and this is rebased onto the updated `v3.20.x`, I'll approve — my own review found nothing here to change.

